PR #15, opened 2015-12-13
Port egs_view to Qt4

mstoeckl:
I noticed that `egsgui` and `egsinprz` had already been ported to Qt4, so I began work on the same for `egs_view`. I have ported the code files and UI files away from the Qt3Support library as well, so that porting to Qt5 in the future should be easy. I tried to stay within the existing code style (is there a preferred auto-format style for the C++ code?).

The code builds and runs like the original (as far as I can tell) on my computer, which uses Qt4.8. I've compiled it and used it on another system running Qt 4.2, so everything in between should be fine.

In the process of looking through the code, I optimized the section that drew the EGS_Vector of float tuples on screen. Instead of taking ~80% of the cpu time per frame, this process now takes about 8%. In practice, the CPU time per frame is cut roughly in half.

Because porting the UI meant renaming/moving some of the files, you should probably look at the commits in sequence for a better picture.

I have not updated the static build makefiles, since it would take too long for me to get the toolchain set up.

### DIFF
--- a/HEN_HOUSE/egs++/view/clippingplanes.cpp
+++ b/HEN_HOUSE/egs++/view/clippingplanes.cpp
@@ -23,22 +23,12 @@
 #
 #  Author:          Iwan Kawrakow, 2005
 #
-#  Contributors:
-#
-###############################################################################
-#
-#  ui.h extension file, included from the uic-generated form implementation.
-#
-#  If you want to add, delete, or rename functions or slots, use Qt Designer
-#  to update this file, preserving your code.
-#
-#  You should not define a constructor or destructor in this file. Instead,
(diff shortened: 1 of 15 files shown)


ftessier:
Thank you @mstoeckl! This is a fantastic contribution; porting egs_view to Qt4 was high on our wish list (send me your full name so we can add it to the list of contributors in the file headers). I am working to develop a coding style with `astyle`, and I will filter the egs_view files through it, so no worries about format for now. Thank you for optimizing the painting through QImage, that was in dire need of optimizing!!

mainegra:
Merge mstoeckl pull request into our develop branch. This was top on our priority TODO list! I will test it on our end prior to releasing it to master.

PR #16, opened 2015-12-20
Threaded egs_view

mstoeckl:
Because the single threaded nature of `egs_view` made it somewhat annoying to work with, I moved the rendering code onto a separate thread to make the UI more responsive.

The rendering code was initially located in the `GeometryViewControl` class, but to keep the redirection limited I moving a large amount of it to the `ImageWindow` and the new `RenderWorker` classes.

As part of the change, I made the image saving routine take advantage of the render thread, so that when `saveImage` is called, the main UI/forms still respond, although the `ImageWindow` doesn't update in the meantime. 

The viewer's net CPU use has decreased, in part because I let transformations preempt the full-detail scene rendering that took place 500ms after every last interaction. 

Additionally, I modified the code segment that governed the scaling of the low-res renders to be more consistent, and to ignore for its scaling the portion of the frame time corresponding to the track rendering.

The same copyright waiving actions as for PR #15 apply. I do not plan to make any further changes to this pull request (although I can if some parts need to be modified,  such as the variable naming conventions and choices, both of which I didn't pay much attention to  :-) ) I have deliberately avoided modern C++, as some people are stuck with old compilers.


mainegra:
Thank you @mstoeckl for making egs_view so much more efficient! It was already more responsive in PR #15 ! Looking forward to trying this one!

mstoeckl:
I also found a bug from this PR today, but a small one. When you have a very simple scene/very fast CPU so that frames can be rendered at full resolution, the scaling is set incorrectly, leading to a frame solely of the color that the top-left pixel should have. [To reproduce, resize the image window to about 64x64 and spin the `mushroom.geom` model around.] The fix is to insert three lines into `image-window.cpp`.

    @@ -176,6 +176,9 @@ void ImageWindow::rerender(EGS_BaseGeometry* geo) {
                             nx,ny,nnx,nny,nxr,nyr);
     #endif
                     nx = nnx; ny = nny;
    +            } else {
    +                nxr = 1;
    +                nyr = 1;
                 }
         } else {
             // First-time scale values.

I can submit a followup PR if you prefer that style of history.

mainegra:
I think it is better to maintain a consistent way of applying the changes to better reflect the work of our contributors. It is great how much effort you have put into this!
So, yes, a PR would be the preferable way!

Do you have any idea about what is causing the segfault from the PR #16? Which Qt version did you use?

I was able to see that the direction becomes a nan because the length of the xs vector is 0. But I have no idea what is causing this!

Thank you!

Ernesto

mstoeckl:
My main computer has Qt 4.2.3, 4.7.1, and 4.8.7. I've built `egs_view` with all three and can not reproduce your segfault. Can you indicate which geometry file you loaded, whether or not you reloaded, and whether or not `g` (the geometry) is a null pointer in your crash? Your stack trace indicates that `howfar(ireg,xs,u,t,&imed,&n)` never gets called.

mstoeckl:
A quick look through the code indicates that the `ImageWindow` *can* crash if certain events/interactions occur during `qApp->processEvents()` in `setGeometry`. The following changes protect from them and so might fix the segfault. Do they?

    --- a/HEN_HOUSE/egs++/view/image_window.cpp
    +++ b/HEN_HOUSE/egs++/view/image_window.cpp
    @@ -271,17 +271,18 @@ void ImageWindow::startTransformation() {
     void ImageWindow::endTransformation() {
         navigationTimer->stop();
         navigating = false;
    -    render(lastRequestGeo, false);
    +    if (lastRequestGeo) {
    +        render(lastRequestGeo, false);
    +    }
     }
 
     void ImageWindow::resizeEvent(QResizeEvent *e) {
     #ifdef VIEW_DEBUG
     egsWarning("In resizeEvent(): size is %d %d old size is: %d %d" " shown:     %d\n",width(),height(),e->oldSize().width(), e->oldSize().height(),isVisible());
     #endif
    -    // TODO make resizing count as a transformation-event
         QWidget::resizeEvent(e);
     
    -    if (e->size() != e->oldSize()) {
    +    if (e->size() != e->oldSize() && lastRequestGeo) {
             // treat this as a transformation, since more resizes tend to follow
             startTransformation();
             render(lastRequestGeo, false);

mainegra:
Yeap, that did the trick!
Thanks!

PR #17, opened 2015-12-22
Fix egs_view thread porting regression where simple scenes black out

mstoeckl:
See PR #16.

This change fixes a porting regression where on interacting with the image window 
it would black out for the duration of the interaction. It also protects against a possible
bug in which the scene could render before a geometry is set.


mainegra:
I had done it manually from your message. But as agreed, I will pull it from your PR's tomorrow. Thanks again!

PR #21, opened 2015-12-30
Fix sampling of point within cylinder shape

mchamberland:
Take into account midpoint of cylinder when sampling within the shape.


ftessier:
Thank you @mchamberland. Is there a reason why you use `this->xo` instead of simply `xo`? Otherwise, please modify your pull request to just `return EGS_Vector(x,y,z) + xo;`

mchamberland:
Only reason for "this" was that I couldn't remember if it was required or not in C++ and opted for the safe option! I'll modify the request.

ftessier:
It is required in other places (in this same .h file) when a local variable `xo` shadows the object's own `xo`, in which case `this->xo` and `xo` are then different variables. Thank you!

ftessier:
@mchamberland, can you squash your two commits together to keep the history as lean as possible? https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits

mchamberland:
Let me know if that works.

PR #25, opened 2016-01-07
Use EGS_SimpleAliasTable instead of EGS_AliasTable where appropriate

mchamberland:
EGS_ShapeCollection and EGS_SourceCollection can both use the simpler and more efficient EGS_SimpleAliasTable instead of EGS_AliasTable.


ftessier:
@mchamberland can you create a new pull request for these changes, which are now in conflict with the reformatted code. Thank you!

PR #26, opened 2016-01-11
Use EGS_SimpleAliasTable instead of EGS_AliasTable where appropriate

mchamberland:
EGS_ShapeCollection and EGS_SourceCollection both use EGS_AliasTable
when they could be using the more efficient EGS_SimpleAliasTable.

This commit also fixes a typo in the header guard of egs_alias_table.h.


ftessier:
Fixed conflicts, merged and rebased in 2328fa6.

PR #27, opened 2016-01-12
Fix bug in cshrc additions script

mchamberland:
Variable expansion with a colon after it treats the letters after the colon as modifiers. Wrap the variable in curly braces to avoid a bad modifier error.


ftessier:
Indeed, this is true csh (although in bash the syntax is ok). Thank you for pointing this out. By the way you really ought to stop using csh: http://www.grymoire.com/Unix/CshTop10.txt :-)

mchamberland:
Ha! I only use csh because it's the default when I log in at Carleton. I guess I **could** change to bash, with minimal effort.

PR #38, opened 2016-02-13
Track view optimizations in egs_view

mstoeckl:
To make `egs_view` more responsive with large (100MB+) track files, I rewrote the track view class, simplifying the memory layout of the vertices and optimizing the projection code slightly. As a side effect, memory usage and loading time appear to have decreased. Once things were faster, I made them slower by enabling clipping planes for all the tracks.

(The current implementation disables clipping planes that cannot intersect any tracks; this might cause up to a +100% lag spike when zooming in just the right distance, but even with the lag spike the render with clipping planes is faster than the original. The effect can be mitigated by implementing per-track bounding boxes and using those to filter the applicable clipping planes, but I don't have time for that right now.)

There also a quite a few miscellaneous changes with the rest of the code code: the primary change is moving more of the image painting work into `RenderWorker`, which makes it easier to save to image precisely what is onscreen. Saving images now disables the `Save Image` button until the image is saved. I've also simplified some of the file loading code.

Bugfixes include reimplementing the `Show regions` checkbox effect, and fixing a nigh-infinite loop when you zoom too far in to the axes (it's been replaced by float->int conversion overflow at high enough zoom levels, but that's far beyond the smallest features of any physical geometry.)

The only changes outside the `egs++/view` are to the gitignore (hiding QT build folders) and to `egs_vector.h`, changing an `EGS_Vector&` to a `const EGS_Vector&` as the vector passed in is not modified.

Right now, the code should work with Qt4.2 to Qt4.8 on all commit boundaries. I've only tested things on Linux.

I can disentangle the changes further if need be, and also rebase my changes onto `develop` rather than `master`.

Copyright assignment is the same way as before.


ftessier:
Thanks once again, @mstoeckl! Track clipping and faster track rendering proves extremely useful!

PR #59, opened 2016-04-29
Simple format specifier fix for the number of computed histories.

Klunkerball:
I believe it is only a problem with the print out of this number and not with the internal format used in the code.


ftessier:
Thanks Berit, can you pull against the `develop` branch? The `master` branch remains unchanged between releases (except for critical bug fixes), so we never pull into it. Also please remove the period at the end of your commit title, as per the recommended git commit message format https://github.com/nrc-cnrc/EGSnrc/wiki/Git-commit-messages, and change it to "Fix format specifier for the number of histories"

Klunkerball:
Hi Fred,
since the branch to pull against can't be changed once the pull request is issued, I will close this request and create a new one with correct branch and commit message. 
Berit

PR #61, opened 2016-05-02
Fix format specifier for the number of histories

Klunkerball:


ftessier:
@Klunkerball, I cherry-picked your pull request to the new `feature-style` branch. From now on, we will segregate typos and format changes to this branch, which will be squashed and merged with `develop` periodically. This will avoid littering the `develop` branch with a large number of small format-only commits.

brevilo:
Just curious, this is a bug that causes (data-dependent) invalid output, even if it's "format-only". Why would a fix like this be relegated to a `feature-style` branch?

Also, do you really mean "squashing"? If so, please don't. Apart from arguable history-loss, other's might already have cherry-picked commits like this to get things fixed before you integrate/release them. Squashing those commits will likely cause conflicts when they try to merge your releases later on. In short, don't squash any commits on/from public branches. JM2C.

ftessier:
We want to keep `develop` as stable as possible, so I relegated this to the style branch, even though this particular issue may cause a wrong number to be output for large ncase. But I can't put this into `develop` because it needs a fix for Windows (where `%lli` is not the appropriate flag for `long long`).

We reserve the right to squash, force push, or otherwise do whatever we want with feature branches. These are meant as temporary working branches and you should **never** rely on those, including for cherry-picking. All commits on feature branches are temporary. This allows us to work effectively as a team, advertise what we are working on, and allow external collaborators to contribute to features.  I will explain this on the wiki as soon as possible.

brevilo:
> it needs a fix for Windows

Fair enough. Are they still stuck at half C89, half C99 where they do support the type but not the format specifier? Sigh...

> I will explain this on the wiki as soon as possible.

Your project, your rules. Just voiced an opinion.

ftessier:
> Your project, your rules, your community. Just voiced an opinion.

Your opinion is very welcome, especially as we build our workflow model and testing different strategies. Thank you!

Issue #43, opened 2016-02-25
Bug in *.io files of MORTRAN user codes


mainegra:
When EGSnrc reads the `*.io` files looking for pairs `unit extension` per line, it also checks for comments by checking whether the first character of a line starts with `#`. This very simple check will fail if there is a blank line which is the case now in the first line of all `*.io` files.

The affected files in `$EGS_HOME` and `$HEN_HOUSE/user_codes` are:
```
./cavrznrc/cavrznrc.io
./cavsphnrc/cavsphnrc.io
./dosrznrc/dosrznrc.io
./dosxyznrc/dosxyznrc.io
./dosxyznrc_old/dosxyznrc.io
./edknrc/edknrc.io
./examin/examin.io
./flurznrc/flurznrc.io
./g/g.io
./sprrznrc/sprrznrc.io
./tutor4/tutor4.io
./tutor7/tutor7.io
```
and the default io file for BEAM user codes:
```
$HEN_HOUSE/omega/beamnrc/default_beam.io
```
Solution: The blank line should be removed.

ftessier:
Fixed in c89f8ff

Issue #60, opened 2016-04-29
Pegsless mode cannot read density correction files when using material.dat file


mchamberland:
Density correction files fail to be read when using a `material.dat` file in pegsless mode. See the sample material and `egs_chamber` input files below. There is no problem when the material is directly defined in the input file, with or without a density correction file.

[material_short.txt](https://github.com/nrc-cnrc/EGSnrc/files/243150/material_short.txt)
[pegsless_material_file.txt](https://github.com/nrc-cnrc/EGSnrc/files/243151/pegsless_material_file.txt)

ftessier:
Thank you for reporting this problem; @blakewalters can you check into this?

blakewalters:
Hi Marc:

I think I found the source of this problem, but to make sure, could you go into your $HEN_HOUSE/src/get_media_inputs.mortran file and look for the code:

        IF(mindex>0 & eindex>mindex)[
         "remove everything up to ="
         text_string=text_string(eindex+1:);

Change this to:

        IF(mindex>0 & eindex>mindex)[
         "remove everything up to ="
         text_string=text_save(eindex+1:);

and recompile your code.  Does this fix the problem?

Blake

mchamberland:
@blakewalters, yes this fixes the problem. Thanks!

ftessier:
Fixed in c89f8ff

Issue #58, opened 2016-04-28
Segmentation fault in pegsless mode when AE = UE


mchamberland:
In pegsless mode, a segmentation fault occurs if AE = UE.

See the simple input file below, which should be run with egs_chamber in pegsless mode.

[pegsless.txt](https://github.com/nrc-cnrc/EGSnrc/files/241191/pegsless.txt)

I know this is a weird edge case, but I figured there might be a better way to handle this than segfaulting. Maybe an error message and exiting?

ftessier:
This one has your name on it @blakewalters!

ftessier:
Fixed in c89f8ff

PR #62, opened 2016-05-02
Feature pegsless fix plus

### LINKED ISSUE #57: dosxyznrc .egslst files not generated
I just migrated a cluster from Egs circa 2012 to EGSnrc 2016. 

Upon doing so an old egsinp (read unmodified) that had previous to the migration produced an egslst no longer does so. 

I found this in testing but thought it was a configuration issue on my test system. As this is a clean install from git I am now starting to think that it might be a regression with the dosxyznrc code. 

What information do you need from me for troubleshooting this one?

blakewalters:
Fix #57 #58 #60


ftessier:
@blakewalters does this fix #43 as well?

PR #63, opened 2016-05-02
Fix pulse height scoring in tutor7pp

mainegra:
Events depositing the maximum particle's energy are not scored in the histogram. As a consequence, large differences are observed in those cases if one compares the energy deposited in a given region calculated from the fraction of energy deposited to the sum over all histogram bins. Now when Emax is deposited in a region, it is scored in the last bin.


ftessier:
@blakewalters please review and give a thumbs up when you have a chance

Issue #64, opened 2016-05-03
Radiative Compton corrections warning in .egslog, but not in .egslst


ojalaj:
I've enabled radiative compton corrections as instructed in the manuals. When performing a DOSXYZnrc run, a following warning is printed in `.egslog` file:
```
 Bound Compton scattering                                    ON
***************** Warning:
 You are trying to use radiative Compton corrections
 without having included rad_compton.mortran
 Radiative Compton corrections                               ON
 Rayleigh scattering                                         ON
```
but not in `.egslst file`. However, it is `rad_compton1.mortran` that I've included in the `Makefile`, not `rad_compton.mortran`. Which one is correct and is it running correctly? For accelerator models, is `sources.make` the correct file to add the line?

mainegra:
Hi @ojalaj, you need to add `$(EGS_SOURCEDIR)rad_compton1.mortran` just before $(EGS_SOURCEDIR)get_inputs.mortran in your Makefile to void the macro $RADC_WARNING which issues the warning message you are seen and also sets radc_flag to 0. The warning message has a typo since it refers to `rad_compton.mortran` when it should be `rad_compton1.mortran`. I realize that the EGSnrc manual states that `$(EGS_SOURCEDIR)rad_compton1.mortran` should be added just before `$(EGS_SOURCEDIR)egsnrc.mortran`. _This of course should be corrected._ The reason for implementing it this way was as safety measure to make sure that one did not forget to add the source file for radiative Compton corrections while requesting the corrections via the input file.

PR #69, opened 2016-05-04
Fix for radiative Compton correction check

mainegra:
According to issue #64 file `rad_compton1.mortran` has to be placed just before `get_inputs.mortran` to pass the check of whether the file was included or not in the compilation process. However, the EGSnrc manual PIRS-701 states that `rad_compton1.mortran` can be placed just before egsnrc.mortran which is not correct as currently implemented. Several issues are fixed in this pull request:

1. Typo in warning message reported `rad_compton.mortran` was needed rather than `rad_compton1.mortran`.
2. Check removed from `get_inputs.mortran` as it was printing a message that disrupted the output format of the MC transport parameters requested by the user.
3. Now the check is actually done during the HATCH step. As a consequence one can now place `rad_compton1.mortran` anywhere between `egsnrc.macros` and `egsnrc.mortran` in the application's Makefile.


ftessier:
Do you want to fix PIRS-701 accordingly?

mainegra:
We could be more specific and say that `radc_compton1.mortran` can be placed anywhere between egsnrc.macros and egsnrc.mortran in the Makefile. But as it is now it is correct. More important will be to explain how to add this file for different type of applications (MORTRAN, C++).

ftessier:
@mainegra, I tried the fix, and although it seems to turn Radiative corrections OFF if I don't include the `radc_compton1.mortran` file, in the `Electron/Photon transport parameter` output in `.egslog` it still reports that is is ON (see below). Can this be fixed easily?

```

***************** Warning: 
 You are trying to use radiative Compton corrections
 without having included rad_compton1.mortran
Turning radiative Compton corrections OFF ...


 Bound Compton scattering not requested! 

  EGSnrc SUCCESSFULLY 'HATCHED' FOR ONE MEDIUM.

================================================================================

                   Electron/Photon transport parameter

================================================================================

 Photon cross sections                                      xcom
 Compton cross sections                                     default
 Photon transport cutoff(MeV)                                    0.1000E-01
 Pair angular sampling                                       SIM
 Pair cross sections                                         BH 
 Triplet production                                          Off
 Bound Compton scattering                                    OFF           
 Radiative Compton corrections                               ON            
 Rayleigh scattering                                         OFF           
 Atomic relaxations                                          OFF           
 Photoelectron angular sampling                              OFF           

 Electron transport cutoff(MeV)                               0.7000
 Bremsstrahlung cross sections                              BH  
 Bremsstrahlung angular sampling                             SIM
 Spin effects                                                On
 Electron Impact Ionization                                  Off
 Maxium electron step in cm (SMAX)                                5.000    
 Maximum fractional energy loss/step (ESTEPE)                0.2500
 Maximum 1st elastic moment/step (XIMAX)                     0.5000
 Boundary crossing algorithm                                 PRESTA-I  
 Skin-depth for boundary crossing (MFP)                      330.8    
 Electron-step algorithm                                     PRESTA-II 

================================================================================
```

mainegra:
@ftessier this is the expected behavior from my fix. I want to keep the output for the `Electron/Photon transport parameter` faithful to the user's input and then warn that the source file required is not available and hence this option will be turned off. Another problem with updating the output message in the `Electron/Photon transport parameter` is that it will force the use of the macro $RADC_WARNING in `get_inputs.mortran` and force placing `radc_compton1.mortran` just before `get_inputs.mortran` in the Makefile.

ftessier:
Ok, understood, let's leave it like that then, however misleading it may be.

ftessier:
Disruptive to the output format, I agree, but less likely to be missed.

mainegra:
Reverted back to checking in `get_inputs.mortran` since as @ftessier commented above it would be misleading for the user. Now file `radc_compton1.mortran` **MUST** be included before `get_inputs.mortran` in the Makefile as suggested in issue #64. Now the check is done before creating the output strings hence the output block `Electron/Photon transport parameter` correctly shows the radiative Compton corrections to be turned OFF. The EGSnrc manual PIRS-701 needs to reflect this!

ftessier:
Also adjust the comment in the header in `radc_compton1.mortran`:
```                                                                             "
"  - add this file to the list of files needed to build EGSnrc just before    "
"    $(EGS_SOURCEDIR)egsnrc.mortran 
```

mainegra:
It is not misleading anymore with the latest commit!

PR #77, opened 2016-05-07
Add rounded rectangular cylinder geometry primitive

mstoeckl:
This pull request introduces a new rounded rectangular cylinder geometry, with interface modelled after `EGS_EllipticCylindersT`. So far, I've used it to model the roundings of milled slots/rectangular holes more accurately, and because a set of concentric rectangles is often simpler and cheaper than voxels or an `XYZGeometry`. 

It appears to pass the `EGS_GeometryTester`, but there may still be edge cases from floating point effects lurking (for instance, one test gets a single almost-not failure at ~10M trials). I also suspect one or two very minor bugs that most likely lie in the interaction between the rounded rectangles and the various composite geometries.

The included test geometry requires the range selection modification of #76 to remain reasonably simple.

Copyright assignment is as before.

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_roundrect_cylinders/egs_roundrect_cylinders.cpp
+++ b/HEN_HOUSE/egs++/geometry/egs_roundrect_cylinders/egs_roundrect_cylinders.cpp
@@ -1,0 +1,170 @@
+/*
+###############################################################################
+#
+#  EGSnrc egs++ roundrect cylinders geometry
+#  Copyright (C) 2015 National Research Council Canada
+#
+#  This file is part of EGSnrc.
+#
+#  EGSnrc is free software: you can redistribute it and/or modify it under
+#  the terms of the GNU Affero General Public License as published by the
+#  Free Software Foundation, either version 3 of the License, or (at your
+#  option) any later version.
+#
+#  EGSnrc is distributed in the hope that it will be useful, but WITHOUT ANY
+#  WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+#  FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for
+#  more details.
+#
+#  You should have received a copy of the GNU Affero General Public License
+#  along with EGSnrc. If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+#
+#  Author:
+#
+#  Contributors:
+#
+###############################################################################
+#
+#  A set of concentric rounded rectangular cylinders.
+#
+###############################################################################
+*/
+
+
+/*! \file egs_cylinders.cpp
+ *  \brief A set of concentric rounded rectangular cylinders: implementation
+ *  \author Manuel Stoeckl
+ */
+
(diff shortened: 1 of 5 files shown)


mstoeckl:
There currently appears to be a corner-case `howfar` bug; it will be fixed soon.

mstoeckl:
I found and fixed a minor floating point bug in the `howfar` code, which triggers most often when the x- and y- components of the direction vector are equal, the rounded rectangle is a square, and floating point errors
conspire to push the intersection point with the sides of the square further out than would be possible
at infinite precision. I have yet to find any other issues related to the rounded rectangle geometry. [The `EGS_NDGeometry` bug I have a fix for, but it needs more testing.]

HEN_HOUSE/egs++/geometry/egs_roundrect_cylinders/egs_roundrect_cylinders.cpp
```diff
@@ -0,0 +1,170 @@
...
+#
+###############################################################################
+#
+#  A set of concentric rounded rectangular cylinders.
+#
+###############################################################################
+*/
+
+
+/*! \file egs_cylinders.cpp
```
rtownson: File should be egs_roundrect_cylinders.cpp

HEN_HOUSE/egs++/geometry/egs_roundrect_cylinders/egs_roundrect_cylinders.cpp
```diff
@@ -0,0 +1,170 @@
...
+#  WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+#  FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for
+#  more details.
+#
+#  You should have received a copy of the GNU Affero General Public License
+#  along with EGSnrc. If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+#
+#  Author:
```
rtownson: Author should be `Manuel Stoeckl`

HEN_HOUSE/egs++/geometry/egs_roundrect_cylinders/egs_roundrect_cylinders.cpp
```diff
@@ -0,0 +1,170 @@
...
+                egsWarning("createGeometry(round rectangular cylinders): missing/wrong "
+                           "'y-axis' input\n");
+                return 0;
+            }
+            g = new EGS_RoundRectCylindersT<EGS_Projector,EGS_Projector>(x_widths,y_widths,radii,xo,"",
+                    EGS_Projector(EGS_Vector(ax[0],ax[1],ax[2]),"Any"),
+                    EGS_Projector(EGS_Vector(ay[0],ay[1],ay[2]),""));
+        }
+        g->setName(input);
+        g->setMedia(input);
```
rtownson: Add a line here to set region labels:
`g->setLabels(input);`

ftessier:
Fixed up, rebased and merge this pull request in commit b99e3b53cb7b02b7a1dba8c3579153b793e7811d

PR #78, opened 2016-05-07
egs_view no longer stretches objects for nonsquare image window

mstoeckl:
Previously, when the image window was resized, the field of view did not change so that the the visible scene would be distorted. Now elongating the image window increases the size of the field of view accordingly to avoid distortion.

Left: Render at 512x1024, before this change; right: render at 512x1024, after this change.
output_k


rtownson:
I took it for a spin! Works great - couldn't find a way to break it :)

ftessier:
Agreed, tested it as well. Very nice, thank you @mstoeckl!

PR #81, opened 2016-05-12
Uniform zooming in egs_view

mstoeckl:
This changeset adjusts the camera zooming behavior of egs_view. Zooming
is now logarithmic, so that each constant change in zoom input (scrollwheel, etc.)
scales the field of view up or down by the same factor, no matter the current zoom
level. (The exception is an outer bound on the zoom level set to the point where the
FOV is ~60 times wider than the target geometry.) The initial zoom level is still within
a percent or so of the previous initial zoom level.

The zoom rate (i.e., how much a pixel shift with the middle mouse button changes the
FOV width) has been arbitrarily set to `2^(1/48)=1.0145`. This is a middle ground
between the slowest prior zoom rate (`1.0060`) and the fastest (`2.5` [although
occasionally `7.0` happens]).

(Why `2^(1/48)`? Then 24 pixels of change yield an exact doubling/halving in FOV
area. Since 24 is divisible by 1,2,3,4,6, and 8, those with scroll wheel increments
as one of these can easily obtain such an scaling. If `2^(1/48)` is too aggressive,
`2^(1/60)` would be a nice alternative.)

### DIFF
--- a/HEN_HOUSE/egs++/view/viewcontrol.cpp
+++ b/HEN_HOUSE/egs++/view/viewcontrol.cpp
@@ -122,13 +122,11 @@ GeometryViewControl::GeometryViewControl(QWidget *parent, const char *name)
     ambientLight->setValue(ilight);
     ambientLight->blockSignals(false);
     distance = 25;
-    //dCourse->setValue(0);
-//     dFine->setValue(25);
-    dfine = 250;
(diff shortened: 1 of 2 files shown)


ftessier:
Very nice, thank you @mstoeckl!

PR #87, opened 2016-05-18
Improve CD geometry's howfar

mchamberland:
Two cases are mishandled by the howfar method:

1) A particle is approaching the geometry, but due to round off, ends
up sitting exactly on the boundary. At that point, it will be considered
inside the CD geometry which causes a mismatch with its global region
number. The problem is that the check for condition (b) in the code
fails to catch this case. When the particle sits on the boundary, the
howfar call will set 'tb' to the distance to the next boundary, which
can be much larger than 'epsilon' (and it can also return ixnew < 0
if there is no other CD geometry region along the direction of 'u' 
between the particle and the outside of the CD geometry.) Solution:
call howfar, but in the opposite direction, which should set 'tb' to 0 if
sitting on the boundary.

2) A particle has just left the geometry but is still sitting exactly
on the boundary. This corresponds to condition (a) in the code. The
problem arises with concave geometries. If the particle is sitting on
the boundary and pointing away from the CD geometry, the howfar call
will set 'tb' to 0. But for CD geometries made of concave geometries,
the particle may re-enter the CD geometry on its next step, so we should
instead go to checks 1. and 2.

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
+++ b/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
@@ -506,9 +506,25 @@ public:
                     }
                     return ixold;
                 }
-                else if (ixnew < 0 && tb <= epsilon) {                         // (a) is true
+				// Skip for concave geometries: particles can reenter the CD geometry after leaving.
+				// Check 2. below will catch these cases.
+                else if ((ixnew < 0 && tb <= epsilon) && (bg->isConvex() && g[ibase]->isConvex())) {                         // (a) is true
                     return ixnew;
                 }
+				// If a particle approaching the geometry sits on a boundary, we look back to see
+				// if we just entered the geometry (the previous checks fail to catch this case).
+				EGS_Float tb_neg = 1e30;
+				int ixnew_neg = howfar(ixold,x,u*(-1),tb_neg,0,pn);
+				if (ixnew_neg < 0 && tb_neg <= epsilon) {                             // (b) is true
+					t = 0;
+					if (newmed) *newmed = g[ibase] ? g[ibase]->medium(icd) :
+						bg->medium(ibase);
+					if( normal ) {
+						*normal = (*pn)*(-1);
+					}
+					return ixold;
+				}
+
                 //***********************************************************************
 
                 // 1. Check if we exit the base geometry after a sufficiently small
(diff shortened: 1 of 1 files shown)


HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
```diff
@@ -506,9 +506,25 @@ class EGS_CDGEOMETRY_EXPORT EGS_CDGeometry : public EGS_BaseGeometry {
...
+					t = 0;
+					if (newmed) *newmed = g[ibase] ? g[ibase]->medium(icd) :
+						bg->medium(ibase);
+					if( normal ) {
+						*normal = (*pn)*(-1);
+					}
+					return ixold;
+				}
+
                 //***********************************************************************
```
mchamberland: Notice that this is similar to what the code previously used to check for condition (b), but it wasn't enough to catch everything. It turns out the current check for (b) is also not enough to cover all cases. Hopefully, this improves the coverage of edge cases.

HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
```diff
@@ -506,9 +506,25 @@ class EGS_CDGEOMETRY_EXPORT EGS_CDGeometry : public EGS_BaseGeometry {
                     }
                     return ixold;
                 }
-                else if (ixnew < 0 && tb <= epsilon) {                         // (a) is true
+				// Skip for concave geometries: particles can reenter the CD geometry after leaving.
+				// Check 2. below will catch these cases.
+                else if ((ixnew < 0 && tb <= epsilon) && (bg->isConvex() && g[ibase]->isConvex())) {                         // (a) is true
                     return ixnew;
```
mchamberland: I'm not entirely sure if those checks will cover everything... Another solution which seemed to work was to move the check for (a) to below check 2. This allows 2. to properly catch the concave geometry cases and then fall back on the check for condition (a), if needed. Thoughts?

HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
```diff
@@ -506,9 +506,25 @@ class EGS_CDGEOMETRY_EXPORT EGS_CDGeometry : public EGS_BaseGeometry {
                     }
                     return ixold;
                 }
-                else if (ixnew < 0 && tb <= epsilon) {                         // (a) is true
+				// Skip for concave geometries: particles can reenter the CD geometry after leaving.
+				// Check 2. below will catch these cases.
+                else if ((ixnew < 0 && tb <= epsilon) && (bg->isConvex() && g[ibase]->isConvex())) {                         // (a) is true
```
mainegra: In this logical if one must check whether `g[ibase] `exists or `icd = 0` to avoid a **segfault** error in cases where only one geometry is inscribed in one of the regions of the base geometry.

mchamberland:
Thanks, @mainegra! I've made the change.

mainegra:
Squash merge this pull request.

PR #91, opened 2016-05-20
Fix 85: Fix NaN cross-sections when compiled with -ffast-math

blakewalters:
Error traced to variable `VPLASM=SQRT(EDEN*R0*C^2/PI)=Infinity` with
certain inputs when `-ffast-math` compiler option used.  Order of
operations changed to ensure argument inside `SQRT()` is not infinite.
Thus, `VPLASM=SQRT(EDEN/PI*R0*C^2)`.


ftessier:
Thanks, good timing! I just found that on at least one recent version of the gfortran compiler the `-O2` flag alone caused NaNs. I bet this will be fixed as well with your change.

blakewalters:
If not, it's back to EDEN ROC (new CDN Sci Fi series?).

PR #96, opened 2016-06-01
Fix density correction path starting in EGS_HOME

blakewalters:
If the directory `$EGS_HOME/pegs4/density_corrections` is not found then the search begins in `$EGS_HOME/pegs4`.


ftessier:
Thanks @blakewalters. Just change the title to <50 chars (Fix density correction path starting in EGS_HOME) and remove the period at the end.

PR #117, opened 2016-06-30
Fix #8: update transport parameter description

### LINKED ISSUE #8: Bound Compton scattering default
The new default value for the `Bound Compton scattering` transport parameter seems to be `norej` but, according to PIRS 701, the default is documented as being `on`.

ftessier:
@mchamberland pointed out that the pirs701 file `transport_parameters.tex` had not been updated to reflect the change of the Bound Compton scattering default value to `norej`. Update the file in this regard, and update a couple of other bits according the pirs702 version of `transport_parameters.tex`.


rtownson:
Line 206: There's an extra ( bracket.

ftessier:
Thanks, fixed in both pirs701 and pirs702.

PR #118, opened 2016-06-30
Fix egs_polygon_shape number of triangle bugs

### LINKED ISSUE #114: Segfault when destructing polygon shape
The `EGS_PolygonShape` destructor tries to delete `n` triangles, but a polygon shape does not have `n` triangles, it has fewer than `n` triangles:

[Here is where `n` is set.](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.cpp#L68)

[Here is where the triangle array is allocated.](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.cpp#L90)

[Here the comment is wrong.](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.h#L139)

[Here is the destructor.](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.h#L125)

### LINKED ISSUE #115: Segfaults and other crashes due to possible floating point errors in EGS_PolygonShape / EGS_2DPolygon
Polygon shapes are built from the triangles of a polygon. The triangles of an EGS polygon are stored in a pre-allocated array of triangle shapes. For example, a trapeze (4 sides) would only need to store two triangles.

In some cases, one more triangle than is needed ends up being created [in this block of code](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.cpp#L90-L130). I suspect floating point errors in EGS_2DPolygon::isInside(). 

I consider this floating point issue more serious than the CD geometry errors because in this situation, it can lead to segmentation faults (e.g., [here](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.cpp#L130), because the extra triangle is being written to an array that does not have enough memory allocated, or even [here](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/shapes/egs_polygon_shape/egs_polygon_shape.cpp#L133)) and other (seemingly random) crashes.

On my system, the following polygon shape produces segfaults, in a non-consistent manner (but definitely produces 3 triangles instead of 2):

```
:start shape:
   library = egs_polygon_shape
   points = -0.1047 0.01232 0.1047 0.01232 0.0965 -0.01232 -0.0965 -0.01232
:stop shape:
```

But this one does not:

```    
:start shape:
   library = egs_polygon_shape
   points = -0.104701 0.012321 0.104701 0.012321 0.096501 -0.012321 -0.096501 -0.012321
:stop shape:
```

which is why I suspect floating point errors to be the underlying cause here.

This issue may also be related to issue #106.

I've included a tutor7pp input file.

[polygon_shape.txt](https://github.com/nrc-cnrc/EGSnrc/files/339345/polygon_shape.txt)

randlet:
The number of triangles required to span a regular polygon defined by N points is N-2 not N-3 so not enought memory was being allocated.  This has been corrected and resolves #115.

The destructor was also trying to delete N triangles rather than N - 2 which resolves #114

### DIFF
--- a/HEN_HOUSE/egs++/shapes/egs_polygon_shape/egs_polygon_shape.cpp
+++ b/HEN_HOUSE/egs++/shapes/egs_polygon_shape/egs_polygon_shape.cpp
@@ -87,7 +87,7 @@ EGS_PolygonShape::EGS_PolygonShape(const vector<EGS_Float> &points,
     }
     EGS_2DPolygon pol(p1);
     int ntr = 0;
-    triangle = new EGS_TriangleShape* [n-3];
+    triangle = new EGS_TriangleShape* [n-2];
     np = n;
     EGS_Float p_tmp[6];
     while (np > 3) {
--- a/HEN_HOUSE/egs++/shapes/egs_polygon_shape/egs_polygon_shape.h
+++ b/HEN_HOUSE/egs++/shapes/egs_polygon_shape/egs_polygon_shape.h
@@ -122,7 +122,7 @@ public:
                      EGS_ObjectFactory *f=0);
     ~EGS_PolygonShape() {
         if (n > 0) {
-            for (int j=0; j<n; j++) {
+            for (int j=0; j<n-2; j++) {
                 delete triangle[j];
             }
             delete [] triangle;


ftessier:
Thank you! @randlet can you just change your commit title to "Fix #114: egs_polygon_shape number of triangle bugs". I noticed recently github then adds a link to issue #114, which is useful when reviewing pull requests (it even adds the links in this post!)

ftessier:
Oh I see this fixes #115 as well. So use "Fix #114 and #115:..."

randlet:
Done :)

randlet:
Ha. Yes it clearly says "commit title" in your post. Is it Friday yet?

ftessier:
Holiday tomorrow, so it counts as Friday :-)

PR #123, opened 2016-07-19
Fix bug in CD geometry

mainegra:
Contributed by @mchamberland


mainegra:
@mchamberland I had to recreate this pull request because I forgot to squash rather than just merge so as to keep the history on develop as tidy as possible.

PR #137, opened 2016-08-02
Fix #125: Fix sync between IAEA phsp DOSXYZnrc source 20

### LINKED ISSUE #125: Lack of sync between the MU index in DOSXYZnrc source 20
A DOSXYZnrc source 20 simulation with IAEA phsp as source (linac beam through static treatment head components) and shared library beam simulation (through patient/plan-specific components) imposed on a CT-based phantom provides unexpected dose distributions, when compared to identical simulation with the 2013 release (with required modifications to several files in HEN_HOUSE and EGS_home). Somehow it seems that there is something wrong in the sync between the MU index-based shared library simulation and DOSXYZnrc.

blakewalters:
Bug was due to invoking $INIT_PHSP_COUNTERS macro at the
beginning of subroutine srcinit.  This reset the flag
iaea_i_muidx to -99 causing MU Index not to be read from
IAEA format phase space files when, in fact, it was stored.

### DIFF
--- a/HEN_HOUSE/user_codes/dosxyznrc/srcxyznrc.mortran
+++ b/HEN_HOUSE/user_codes/dosxyznrc/srcxyznrc.mortran
@@ -2397,8 +2397,6 @@ $REAL weight,
       cosphi,sinphi;
 $INTEGER i;
 
-$INIT_PHSP_COUNTERS;
-
 OUTPUT61;(' ');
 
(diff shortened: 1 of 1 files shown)


blakewalters:
Realized there's a bit more to this fix.  I'm about to delete this pull request and resubmit with the additional modification.

Issue #139, opened 2016-08-11
Add labels feature to existing applications and documentation


jw3126:
First of all it is great to have a free and powerful MC tool like EGSnrc. Thanks a lot to everybody involved in creating it! One pain point about EGSnrc is with the leaky abstraction of geometry. What I mean is that one is forced very often to reference pieces of the geometry by integers even if they have string names.

For example assume we want to to score dose deposition in a detector. In order to do this, one would create an input file for egs_chamber. In the input file one builds up the geometry from basic pieces like `egs_box`, `egs_gunion`, ... and each of these items is assigned a hopefully descriptive name by the user. 

But when it comes to specifying areas for variance reductions or scoring, all these descriptive names are worth little and one has to play around with `egs_view` to find out region numbers and put them into the input file (BTW I like `egs_view`, it produces amazing visualization!). This is time consuming and error prone, it would be cool if the computer could do it for us. 

1) Are there any plans is this direction? 

In a perfect world, we would have a function, 

`getRegionByName(name)`

which takes a name and returns its region number. Unfortunatelly pieces in the input file are not 1:1 with region numbers, so this can not quite work. For example a single `EGS_ZCylinders` might have many "anonymous" subregions, while an `EGS_transformed` does not increase the region count at all. However each geometry class could have a method

`getRegionByName(name, i)`

which calculates the region number of the ith piece of the subregion called `name`. So for instance if the simulation geometry is a `gunion` of `boxy` the `EGS_Box` and `cyly` the 5 layered `EGS_ZCylinders`, invoking `getRegionByName(cyly, 3)` on the simulation geometry would give the region number of layer 3 of our cylinder. 

2) Do you think that is would be a reasonable approach?
3) To me it seems one could implement this function roughly like `isWhere`, but passing `name,i` instead of a point around. Do I see this correctly or are there hidden traps?

ftessier:
Thank you for your suggestion. In fact, this functionality has already been implemented via region "labels", however this feature has not yet been documented yet, or been added to the applications distributed with the software. Labels allow you can assign a string to region(s) in a geometry in the input file, using the local region indices, e.g.,
```
set label = mylabel 12 17
```
Once defined, you can obtain a list of global region numbers from the label string at any point in the application, e.g.,
```
 vector<int> regions;
g->getLabelRegions(mylabel, label_regions);
```

jw3126:
Thanks, that sounds great! Can you point me at the C++ code, where these are implemented?

[...]

PR #146, opened 2016-08-18
Add region labels to egs_chamber and egs_dose_scoring

rtownson:
Addresses issue #139.

Now you can set labels in geometries in egs_chamber using `set label` within the geometry block. For example:

```
:start geometry:
        ...
        set label = myLabel 3 4 5
:stop geometry:
```

This assigns the local geometry regions 3 4 5 to the label named myLabel. Cavity regions, enhance regions and subgeom regions can then use myLabel in place of the global region numbers. Multiple labels, and mixed global region numbers with labels are both supported.

```
:start scoring options:
    :start calculation geometry:
        geometry name = myGeom
        cavity regions = detectorLabel
        cavity geometry = myGeom
        enhance regions = enhanceLabel
        enhancement = -128
        cavity mass = ...
    :stop calculation geometry:
:stop scoring options:
```

I also added this to `egs_dose_scoring`, so you can use region labels for the `dose regions`.


ftessier:
Fixed, thanks.

PR #147, opened 2016-08-18
Add multiregion scoring to egs_chamber

MartinMartinov:
Modifications by Martin Martinov, Carleton: added the possibility to output dose of every single scoring region, rather than a total sum.  Usage,

:start scoring options:
    multiregion scoring = 1

    :start calculation geometry:
        geometry name  = Phantom
        cavity regions = 1 2 3 4 5
        cavity mass    = 1 1 1 1 1
    :stop calculation geometry:
:stop scoring options:

Note: This option will throw an egsFatal() if the number of regions and masses are not the same

### DIFF
--- a/HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp
+++ b/HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp
@@ -57,7 +57,6 @@
 ###############################################################################
 */
 
-
 #include <cstdlib>
 // We derive from EGS_AdvancedApplication => need the header file.
 #include "egs_advanced_application.h"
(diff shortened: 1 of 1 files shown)


ftessier:
@MartinMartinov individual region dose scoring is already implemented via the dose scoring ausgab object. I prefer not to further bloat egs_chamber beyond its current state if the intended feature is otherwise available.

MartinMartinov:
@ftessier Unfortunately, the ausgab objects don't work with intermediate phase space, because the region indexing switches for each geometry used (ie, lots of seg faults).  The advantage of multiregion scoring is it does work with the IPSS option, which is what I required.

ftessier:
I see what you mean @MartinMartinov, thanks for pointing this out. I will reopen this pull request to remind ourselves about it. Perhaps we can look into implementing labels in the dose scroring objects to allow scoring across different simulation geometries. Otherwise we could resort to the upgrade you propose to this end. Thank you

rtownson:
Currently the plan along this front is to consider implementing a geometry tag in ausgab objects. This would allow for multiple ausgab objects to be defined, each corresponding to a different geometry.

Careful if you are using this pull request for anything important; I came across some dose discrepancies (that I did not investigate).

MartinMartinov:
Do you mind passing the file that caused the dose discrepancies onto me to look at?  I'm currently using a similar implementation, so a discrepancy is quite disconcerting...

rtownson:
This was my mistake - ausgab did not show me all the digits on the density of my region so I used the wrong mass, resulting in a dose discrepancy between ausgab and your code.

Issue #82, opened 2016-05-13
Unnecessary warning when lowest energy input for density correction is equal to AE


mchamberland:
This is just a minor inconvenience. When the lowest energy input for the density effect is exactly equal to AE, a warning is output saying that it is higher:

 *_**LOWEST ENERGY INPUT FOR DENSITY EFFECT IS 5.120E-01
                   WHICH IS HIGHER THAN THE VALUE OF AE= 5.120E-01 MEV
 *__IT HAS BEEN SET TO AE_**

It would be nicer to only output a warning if the lowest energy input is strictly lower than AE (and silently allow lowest energy input == AE).

rtownson:
I believe this is now fixed in the master branch.

PR #148, opened 2016-08-19
Fix #22: Update the electron rest mass value

### LINKED ISSUE #22: Update electron rest mass
EGSnrc uses 511.0034 keV as the rest mass of the electron. But the accepted value is now 0.510 998 9461(31) MeV according to [CODATA 2014](http://physics.nist.gov/cgi-bin/cuu/Value?mec2mev). This is a change of only 0.0009%, but EGSnrc should nevertheless use the updated value. (Suggested by Dave Rogers).

There are about 60 instances of the hard coded 0.5110034 rest mass throughout the code:
`grep -R "0.511003" * | grep -v EX16MVp.egsgph | wc -l`

The hardcoded value for the rest mass in `examin` in two instances is set to 0.511003 instead of 0.5110034 (see https://github.com/nrc-cnrc/EGSnrc/blob/master/HEN_HOUSE/user_codes/examin/examin.mortran#L555)

rtownson:
Addresses issue #22.

Update the electron rest mass value to 0.510998946131
throughout all codes. Instances of 0.511, 0.5110034,
and 0.511003 are all changed.

**Please help me review these changes carefully!**


mainegra:
To implement this change at a global level you can make use of the variable `RM` in common `USEFUL`. The electron rest mass in MeV is available in `EGSnrc` to all apps through the common `USEFUL` via the variables `prm` (double precision) and `rm` (single precision). In `C++` apps one needs to make use of the C structure `the_useful`, i.e., `the_useful->rm`.

rtownson:
Good call @mainegra! I got a big bogged down in the mortran, but I'll give this another go next week.

rtownson:
Now the codes use the_useful and COMIN/USEFUL common blocks for defining the electron rest mass.

ftessier:
Awesome @rtownson! This update was long overdue! I note in `egsnrc.mortran` you removed a few instances of the `medium` variable; was it unused? Also it might be optimal to change the `/1000` instances to `*0.001` (multiplication is more efficient), although compilers probably handle that automatically nowadays...

ftessier:
We might also want to cite the current CODATA value, for future reference.

rtownson:
The `medium` variable is defined in `COMIN/USEFUL`, so I had to remove the local definition to avoid a double declaration of the variable. I will add the CODATA reference. I think most compilers would make the switch to `*0.001`, but I can do it anyways!

ftessier:
@blakewalters, @dworogers, can you review the changes to the electrons mass? In particular, there are a few instances where a low-precision hard-coded value 0.511 was used in comparison operators, e.g. `IF((E(NP)-0.511*ABS(IQ(NP)))>EKMAXPHSP(ISCORE))`. I would like to confirm that the low-precision 0.511 was not intended to avoid finite precision issues in the conditional.

dworogers:
Hi Frederic,
It was just being lazy if I did it. I should have used PRM.  Same in many
user codes. I actually tracked some problems to the very small difference
in FLURZnrc in the lowest energy bin if it was narrow enough to have the
small effect show up.

Dave

dworogers:
Hi Frederic:

I see no reason to keep those low-precision values around.  Certainly when
"0.511" was used in the code it was, as Dave said, just laziness as opposed
to trying to avoid any precision issue.

Blake

ftessier:
Oops missed this: @rtownson the recommended value is  0.5109989461(31), not  0.5109989461**31**. The last two digits in paranethesis are meant to represent the standard deviation in terms of the last digits, so this notation means  0.5109989461 +/-  0.0000000031. Can you correct to 0.5109989461 everywhere? You might also want to add in the comment that this is the **2014** CODATA value; it gets updated every few years...

rtownson:
I couldn't figure out how to get `COMIN/USEFUL` accessible in `iaea_phsp_macros.mortran` without causing problems, so I used a hard-coded value there.

mainegra:
@rtownson there is one spot more to update the e- rest mass: pegs4_macros.mortran, line 419.

The rest mass is calculated from the constants in that file amounting to a value of  0.51100761 MeV

Incidentally, this is what causes the annoying warning messages in EGSnrc of 0.512 being less than 0.512 reported by @mchamberland in issue #82. In fact the values being compared are 0.51200761 (the above plus 1 keV) against an AE of 0.512.

rtownson:
@blakewalters do you mean `PRM` instead of `RM`? I don't see any `0.510999` instances. I imagine we could use `PRM` in more cases.

ftessier:
I have squashed and reworded some commits in this pull request to clean up; but I have not changed anything in the files themselves.

ftessier:
@rtownson care to add your name to the contributors list in the 30+ files?

PR #163, opened 2016-08-31
Fix typo in comment in PYRAMIDS_cm

jantolak:
Trivial change, really. The number of local regions is 3 times the number of pyramid blocks since the inner and outer air regions are actually the same numbered region. Change is to the comment only, for clarity.


ftessier:
Thank you @jantolak! @rtownson can you check this change to the comment?

rtownson:
@jantolak is correct! This is verified elsewhere in the code, e.g. line 1172 and in `WHERE_AM_I_$PYRAMIDS`.

However, please make the **pull request against the `develop` branch instead of `master`**, thanks!

ftessier:
I changed the base to develop, did not know we could do this from our end: just click on the `Edit` button right of the pull request title!

PR #168, opened 2016-09-02
Fix #111: EGS_PrismT geometry bug when enclosing zero along closed axis

### LINKED ISSUE #111: egs_view reports "too big: inf" when translating a prism in the negative x-direction
Try viewing the following file in egs_view, which is simply a base prism which is transformed by translating it in the negative x direction. egs_view complains about `too big: inf` and does not display anything. This seems to be caused by the negative translation along x. Remove the negative sign and it displays properly.

[translate_prism_egs_view.txt](https://github.com/nrc-cnrc/EGSnrc/files/337431/translate_prism_egs_view.txt)

rtownson:
Fix a bug resulting in a crash when a closed prism contained the zero point of the axis along which it is closed. For example, an `EGS_PrismX` with `closed = -1 1` would crash. Transformations to this condition, where zero was enclosed, had the same crash.

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_prism/egs_prism.h
+++ b/HEN_HOUSE/egs++/geometry/egs_prism/egs_prism.h
@@ -23,7 +23,7 @@
 #
 #  Author:          Iwan Kawrakow, 2005
 #
-#  Contributors:
+#  Contributors:    Reid Townson
 #
 ###############################################################################
 */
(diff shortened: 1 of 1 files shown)


ftessier:
@rtownson can you update this? Checkout branch `fix-prism-geometry-bug`, update the file `egs_prism.h`, then `git add egs_prism.h`, `git commit --amend` and finally `git push --force`. Or I can do it, let me know!

PR #120, opened 2016-07-06
Implement user-defined boundary tolerance in geometry objects

mchamberland:
This is an attempt at resolving issue #86. Feel free to simply close if you'd prefer using a different approach than I did.

My goal in this pull request is to have a user-defined `boundaryTolerance` variable that is used to check for floating point errors in geometries. The `boundaryTolerance` variable is a protected member of `EGS_BaseGeometry` and has a default value of 1e-10. Being non-static, this allows the user to set a different value of `boundaryTolerance` in any geometry. The tolerance can be set with the input key `boundary tolerance`.

`EGS_StackGeometry` already has a `tolerance` variable that serves the same purpose. For backwards compatibility, the `tolerance` key is preserved and simply duplicates the functionality of the `boundary tolerance` key.

PR #177, opened 2016-09-08
Fix #86, #106: Add boundary tolerance and epsilon

### LINKED ISSUE #86: Use a default epsilon to check for geometrical round-off errors
- Use a default epsilon and remove all the hard-coded limits. 
- Allow users to re-define the default value for epsilon depending on their world's dimensions. 
- Set default value to 1e-10 m.

### LINKED ISSUE #106: Geometry errors when pyramid or prism is inscribed in composite geometry
Another CD geometry error issue from me, what a surprise!

Create a prism or a pyramid. Inscribe it in a CD geometry (for example, use a box as the base). Envelope the CD geometry in a box. Look at the geometry in egs_view and you'll notice _tons_ of geometry errors. See the input files for tutor7pp below.

I know where the code is getting confused: when a particle is sitting on one of the sides of the prism or pyramid (type EGS_Polygon) and leaving the CD geometry. The howfar of CD geometry somehow does not catch this. My problem is that I find the EGS_Polygon's howfar method hard to read through and understand what's going on, so if someone else wants to tackle this case, be my guest.

[prism_cdgeom_errors.txt](https://github.com/nrc-cnrc/EGSnrc/files/309500/prism_cdgeom_errors.txt)
[pyramid_cdgeom_errors.txt](https://github.com/nrc-cnrc/EGSnrc/files/309501/pyramid_cdgeom_errors.txt)

rtownson:
Fix #86: Added a new `boundaryTolerance` parameter for floating point comparisons in geometry `howFar` routines. The user can change it in the input file. It defaults to `epsilon`, introduced below (which defaults to `1e-10`). 

Added a new constant `epsilon` in `egs_functions.h` that is used in the place of hard-coded small floats when doing floating point comparisons. `epsilon` can be used anywhere in egs++, whereas the `boundaryTolerance` variable is used only in geometry `howFar` functions.

Fix #106: Fixed a few geometry errors in `EGS_Polygon` and `EGS_Pyramid`.

Fixed a bug in `EGS_Projector:getProjection()`. 

This replaces pull request #120.


mchamberland:
The fixes to the prism/pyramid/polygon/projector work great! Thanks, @rtownson! Also, good idea to add the `epsilon' constant to complement the `boundaryTolerance'.

PR #184, opened 2016-09-21
Add egs++ volume calculation

rtownson:
Add a Monte Carlo volume calculation to `egs++`, originally designed by @ftessier. It can be turned on using `:start volume calculation:` in any input file. The recursive algorithm builds an octree with higher resolution in geometries with more regions. A user specified box is used as the boundary for sampling.

For example:

```
:start volume calculation:
    passes = the number of recursion passes to create the octree
    samples = the number samples to perform per node per pass
    box min = minX minY minZ
    box max = maxX maxY maxZ
    labels = labels you have added to your geometry using `set label`
:stop volume calculation:
```

For the volume calculation, the sampling box is divided into progressively smaller voxels as the geometry is analyzed on each pass. The algorithm follows the creation of an `octree`. Areas with a greater density of geometry  regions end up with a higher density of voxels and thus are sampled more often in the volume calculation.

As the volume calculation progresses, the number of samplings in each octree node is also adjusted. On each pass, regions with a higher uncertainty  on the volume are sampled more often, while regions with lower uncertainty are sampled less often.

_I'm open to suggestions on this implementation!_ Is is bad practice to add the `Volume_Node` and `Volume_Info` classes to `egs_base_geometry.h`? Maybe these should go in a separate file. Is there additional or alternative functionality we would like?

### DIFF
--- a/HEN_HOUSE/doc/src/pirs898-egs++/geometry.doxy
+++ b/HEN_HOUSE/doc/src/pirs898-egs++/geometry.doxy
@@ -317,6 +317,51 @@ developing new geometries and has helped to find numerous bugs
 in the initial implementations of the various geometry classes
 provided with \c egspp.
 
+\section volume_calculation Calculating volumes
+
+The volumes of regions or media in your geometry can be calculated via
+Monte Carlo sampling using the built in volume calculation. To perform the
+volume calculation, you define a box within which test points will be generated.
+Make sure that the box encloses your entire volume(s) of interest.
+
+The volumes of each region and medium are automatically printed as output.
+The user may also specify regions to calculate the volume in using labels.
+After setting labels in geometries of interest, list the labels under
+the \c labels input. For each label, the sum of the region volumes is output.
+
+To turn on <tt>volume calculation</tt>, include a new control block in your input
+file:
+\verbatim
+:start volume calculation:
+    passes = the number of recursion passes to create the octree
+    samples = the number samples to perform per node per pass
+    box min = 0 0 0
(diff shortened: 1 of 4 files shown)


rtownson:
Fred has an update and bug fix pending for this PR.

ftessier:
Indeed, we need to clean this up further in terms of memory management and uncertainty tracking. I will merge this into develop after the release, to be merged in master in 2018.

rtownson:
Redesign suggestion:

Maybe the volume calculation should be something you can select from the `run control`, so that from any application you can request to just do a volume calculation on any egs++ geometry. Then if the run type is a volume calculation there can be an extra volume calc input block. Importantly, with the volume calculation structured as a class instead of an app, other applications could make use of the volume calculation tools for their own purposes.

mainegra:
@ftessier and @rtownson , are we still deciding what the best course of action is? Could we not merge `egs_volume` as it is and then continue its improvement? 
As it currently stands users would benefit tremendously from having such a "little" app at their disposal for geometries whith very complex regions.

ftessier:
I propose to include this via a standalone application, and not via the egs++ geometry. Usage is then clear for the user, and development is freed of the concerns of integration in the geometry library (it could be integrated back as a geometry method later on). I will convert this PR back to a standalone `egs_volume` application, and merge it into develop just after the Release 2021. The idea is that any egs++ geometry input file with a `volume calculation` input block.could be run with `egs_volume` to calculate the region volumes.

mainegra (approved):

PR #195, opened 2016-11-04
Fix beamdp GUI browser bug.

blakewalters:
Browser would only show IAEAphsp files if the GUI was
started in the directory containing the files.  Once
you navigated away and back into the directory, these
files were no longer visible.


blakewalters:
Hmmm.  Looks like I didn't commit this change properly.  I'll undo this pull request and try again.

PR #200, opened 2016-11-10
Fix beamdp GUI browser bug.

blakewalters:
Browser would only show IAEAphsp files if the GUI was
started in the directory containing the files.  Once
you navigated away and back into the directory, these
files were no longer visible.


blakewalters:
Note: This pull request replaces #195 which @crcrewso noted had changed the wrong file.

crcrewso:
It worked on my one WSL install that has a few other pull requests. I will test it against a more traditional install in the near future.

PR #183, opened 2016-09-20
Fix pegsless inputs in beam and dosxyz GUI's

blakewalters:
Inputs now resemble those for egs_inprz and are similar to those
used to create pegs data in egs_gui.  In the process, fixed a bug
in dosxyz gui's get_val procedure.

Issue #100, opened 2016-06-02
EGS_GUI & PEGS4 not working, when generating data for compound with density effects


ojalaj:
Since the issue #67 and the fix #80 are in progress, the PEGS seems not to work correctly on command-line either, so I tried to generate PEGS data through EGS_GUI for water with AE=0.512 using the density effect correction file found under the distribution (copied first the file to `$EGS_HOME/pegs4/density_corrections/`).

After selecting the density correction file the GUI for some reason chose the 'Medium type' as mixture, which I had to change to compound by first unselecting ICRU density correction option, then selecting the correct medium type and then re-selecting ICRU density correction option. However, when running the program, an empty `.pegs4dat` file was generated and error below was output under 'Details':

```
0**\* COMPOSITION IN INPUT DENSITY FILE DOES NOT MATCH  THAT BEING USED BY PEGS
 ****\* QUITTING EARLY***
```

ojalaj:
With a pristine clone of most recent develop branch I confirm that this issue still exist, i.e. when with EGS_GUI selecting HEN_HOUSE/pegs4/density_corrections/compounds/water_liquid.density the medium type is grayed out to Mixture. When setting AE=0.512, AP=0.01, UE=55, UP=55, the program creates PEGS data (with 'warning'  ****LOWEST ENERGY INPUT FOR DENSITY EFFECT IS 5.120E-01 WHICH IS HIGHER THAN THE VALUE OF AE= 5.120E-01 MEV  ***IT HAS BEEN SET TO AE***), but if changing the medium type to Compound (which I believe water is), the data creation ends to error seen in the original comment.

blakewalters:
From the point of view of PEGS4, the only difference between a "Compound" and a "Mixture" is that in the former the composition is specified using the stoichiometric coefficients of the constituent elements, while in the latter composition is specified using the fraction by weight of each element.  This means that the same medium could be defined two different ways.  Since density correction files define composition using fraction by weight, when they are read in, the medium type automatically defaults to "Mixture."  The GUI bug that you've found is that, if a density correction file has been specified, then the medium composition isn't automatically (re)read in and medium type (re)set to "Mixture" every time the "ICRU density correction" option is (re)checked.

PR #206, opened 2016-11-30
Fix beam gui gasp input

blakewalters:
Disable the ability to specify the medium is a gas if a density correction file is used.  GASP is not
used in this case.


rtownson:
Looks like this PR also includes the commit from PR #183, so we can close that PR and just consider this one.

ojalaj:
Is the problem described in #100 related to this?

PR #212, opened 2017-01-17
Fix kerma calculations in g and DOSRZnrc with EII

rtownson:
Fix the kerma calculation in `g` and `DOSRZnrc` and add `ausgab` calls for Auger, fluorescence and electron impact ionization (EII). Dave Rogers performed much of the investigation of this issue and proposed the solution.

Under certain conditions (e.g. 25 MeV photons in aluminum), the kerma was shown to have a dependency on `ecut` when EII was turned on. This fix addresses that dependency, which was a result of the `edep` parameter that accumulates energy depositions below the cut-off threshold. The `edep` parameter collects contributions over a series of relaxations, and may include both Auger and fluorescence. In some codes, ausgab calls with `iarg=4` is used to collect `edep` contributions for kerma calculations. However, for kerma one needs to collect only the energy deposited by charged particles liberated by uncharged radiation. Therefore it is necessary to include only those `edep` contributions which are appropriate.

Which `edep` contributions should be kept? Auger from relaxations should be included in kerma for photo-electric and Compton interactions, but not fluorescence. When relaxations occur as a result of EII, we only want to keep the fluorescence contributions to calculate g and Kcol (not kerma). 

This fix was achieved by including a new `edep_local` parameter that includes only the most recent energy deposition event under the energy cut-off threshold. Combined with new ausgab calls, applications can now sort out the sub-threshold contributions to correctly calculate the kerma, Kcol and g. This has impact in the `g` and `DOSRZnrc` applications, which have been modified accordingly.

Note that the `edep` parameter and `iarg=4` calls were not changed, and remain for backwards compatibility.

Additionally, one of the energy cut-off checks for Auger production in the relaxation routines was changed to use `ecut` instead of the minimum of `ecut` and `pcut`. @mainegra - does this make sense?

@blakewalters is it ok to have AUSGAB calls for Auger/fluorescence when the interaction is sub-threshold and the particle is not added to the stack? Do any codes watch the AUSGAB and then expect the particle on the stack?


ftessier:
This is wonderful @rtownson. Thank you for taking on the challenge of implementing a solution, and to Dave Rogers for the initial investigation.

PR #214, opened 2017-01-19
Fix error in fluence for sources 1 and 7

blakewalters:
Previously, particles that missed the geometry were subtracted
from the total fluence (used to normalize dose) instead of
added.  This bug has been here for years, indicating, hopefully,
how infrequently these sources are used.


blakewalters:
Hope we can get this in before the release!

PR #216, opened 2017-01-20
Add EGS_RadionuclideSource

rtownson:
Add support for isotropic radionuclide sources using ENSDF decay data.

Emissions from this source depend on the selection of an isotope (or mixture of isotopes) that are modeled based on the decay structure provided by ENSDF decay data. Radionuclide decays into daughter products are modeled, including internal transitions from excited states and metastable isotopes. Event-by-event correlations are modeled.

By default, relaxations producing fluorescence and Auger electrons are modeled using EADL relaxations in c++. However, there is an option in `EGS_RadionuclideSpectrum` to instead statistically model the relaxation emissions from the ENSDF files. In that case, the fluorescence and Auger electrons are not correlated with disintegration events.

The shower index and emission time of particles is tracked, to allow for modeling of coincidence counting. These can be accessed from an egs application using `source->getShowerIndex()` and `source->getTime()`.

Currently there are some caveats:
1. Angular correlations between internal transitions are not modelled. This is under development.
2. For atomic relaxation emissions due to decays, the shell in which to create a vacancy is correctly sampled. However, the sub-shell is selected uniformly. For example, after choosing a vacancy in the L shell, whether to create L1, L2, or L3 vacancy is sampled uniformly.
3. Alpha particle energy is deposited in the source region. If the source regions are very small and low density, this may not be accurate.
4. The binding energies of atomic relaxations due to decays are based on the EADL data, regardless of the photon cross-sections selected as a transport parameter. This means that if you model the actual radioactive material in the geometry and select XCOM cross sections, there may be inconsistent binding energies between relaxations from decays and relaxations from transport.

For more details, see the doxygen documentation in the code.

This is a major addition: design, feature and documentation critiques are welcome!


ftessier:
Waiting on a reply from LNHB regarding the disctribution of their ENSDF files.

mainegra:
@rtownson I see you are using the 1 byte version of `relax.data`. That means we are duplicating the file just with two different record lengths. The reason I was using the 4 bytes version was to avoid conflicts with the Intel compiler back when this file was created. But perhaps we could now use the 1 byte version throughout EGSnrc.
Also, the binding energies in `relax.data `and `relax_onebyte.data` files are those from EADL and are slightly different from those used in the XCOM library. After pull request #282 is merged, one can get the unique binding energies (EADL or XCOM) from `shell_be` in common `SHELL-DATA`. This will require adding a new C structure in `egs_interface2.h` and linking it to the common block in `egs_interface2.c`. I say `shell_be`, rather than `binding_energies` or `be_array` because it is in the order corresponding to the relaxation algorithm implementation.

rtownson:
I think we can merge this branch now. The caveats are stated in the main PR text. We can consolidate the 1-byte and 4-byte `relax.data` versions later. I will also consider how to handle different binding energies in a later update.

ftessier:
Just rebased on develop. Going through it now, hold off on any updates, or else push using `--force-with-lease`.

ftessier:
Reviewed and adjsuted commit messages etc. Checked for EOL spaces, contributor names, and astyle. Just a couple changes upon a final `astyle` run:
```diff
diff --git a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.h b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.h
index 1ca5093..e4bc176 100644
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.h
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.h
@@ -197,7 +197,7 @@ public:
 
     int processEvent(EGS_Application::AusgabCall iarg, int ir) {
 
-        if(ir == -1) {
+        if (ir == -1) {
             ir = app->top_p.ir;
         }
 
diff --git a/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.cpp b/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.cpp
index 04a40d5..1ff5c81 100644
--- a/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.cpp
+++ b/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.cpp
@@ -295,9 +295,10 @@ EGS_I64 EGS_RadionuclideSource::getNextParticle(EGS_RandomGenerator *rndm, int
     getPositionDirection(rndm,x,u,wt);
     latch = 0;
 
-    if(disintegrationOccurred) {
+    if (disintegrationOccurred) {
         xOfDisintegration = x;
-    } else {
+    }
+    else {
         x = xOfDisintegration;
     }
```

ftessier:
@rtownson would this additional commit 6e52567 take care of moving the ENSDF files to a sub-directory?

ftessier:
Note: I have to add the acknowledgment notice before this pull request is merged.

ftessier:
Just adjusted a couple commit titles.

Issue #154, opened 2016-08-25
Radiative Compton corrections for egs++ codes


ojalaj:
According to PIRS 898 Radiative Compton corrections could be just turned on with egs++ codes, but this generates:
**************\* Warning:
You are trying to use radiative Compton corrections
without having included rad_compton1.mortran
Turning radiative Compton corrections OFF ...

and I could not find instructions where to include line with rad_compton1.mortran with these codes.

mainegra:
@ojalaj this information is sadly still missing in the documentation. We will update the documentation with the required information soon via a pull request. Meanwhile here is the skinny on how to include `rad_compton1.mortran` for all kinds of EGSnrc applications:

To turn ON radiative Compton corrections `rad_compton1.mortran` must be first added to the list of MORTRAN sources to be built:

- For  all C++ applications one must edit the file `$HEN_HOUSE/specs/egspp1.spec` and add
`$(EGS_SOURCEDIR)rad_compton1.mortran` just before `$(EGS_SOURCEDIR)get_inputs.mortran`
in the definition of the C_ADVANCED_SOURCES variable if the application is derived from the EGS_AdvancedApplication class or the C_SIMPLE_SOURCES variable for any application derived from the EGS_SimpleApplication.

- For a specific C++ application one could define the variable CPP_SOURCES in the application's Makefile in the same manner as the C_ADVANCED_SOURCES in the file `$HEN_HOUSE/specs/egspp1.spec` and add `$(EGS_SOURCEDIR)rad_compton1.mortran` just before `$(EGS_SOURCEDIR)get_inputs.mortran`.

- For a MORTRAN application named for instance `app` one must edit the corresponding file `app.make` located in the application's folder `$EGS_HOME/app` and add `$(EGS_SOURCEDIR)rad_compton1.mortran` just before `$(EGS_SOURCEDIR)get_inputs.mortran` in the definition of the SOURCES variable.

- For a BEAMnrc application one must edit the file `sources.make` and add
`$(EGS_SOURCEDIR)rad_compton1.mortran` just before `$(EGS_SOURCEDIR)get_inputs.mortran`
in the definition of the SOURCES variable or the LIB_SOURCES variable if this BEAMnrc application will be used as a particle source for another application.

Hope this helps!

ojalaj:
[...]

PR #221, opened 2017-01-26
Add more detail about enabling radiative Compton corrections

mainegra:
This is different depending on the application type. This addresses issue #154 .


ftessier:
Well put @mainegra! Updated the commit message, but did not otherwise change the file.

PR #224, opened 2017-01-31
Fix electron rest mass access in C++ beam and phsp sources

mainegra:


ftessier:
Just squashed the commits and updated the commit message.

PR #226, opened 2017-01-31
Fix edep_local missing from cpp interfaces

rtownson:
The new `edep_local` variable from recent kerma/g fixes was not declared in the cpp interface files. This resulted in the wrong dose in cpp applications that used `iarg=4` (tutor2pp).


mainegra:
Cool find @rtownson !!!

ftessier:
Adjusted the commit title.

PR #231, opened 2017-02-04
Fix #97: Remove infinite loops and add veryFar parameter

### LINKED ISSUE #97: Infinite loops in CD_Geometry
We have a problem where very rarely (i.e.  ~ 1/10E9 histories) the code is getting stuck in an infinite loop in [this code block in CD_Geometry](https://github.com/nrc-cnrc/EGSnrc/blob/master/HEN_HOUSE/egs%2B%2B/geometry/egs_cd_geometry/egs_cd_geometry.h#L611-L687).    This occurs less frequently after the fixes from @mchamberland's CD geom pull requests are applied but still is causing us some issues.

We don't have a test case where it's fast to replicate the issue yet but I'm assuming this is a floating point /  roundoff issue.  

Ideally we could prevent this from ever happening but given the rarity of the issue I'm wondering if we can just terminate the loop (output a warning?) after a certain number of iterations and return e.g. whatever region in the base geometry we are "in" (ibase_new).  Not sure what the best approach is, but IMO it's bad practice to have loops which may not ever terminate so at the very least I believe there should be an iteration counter (or switch the loop to a for loop) which terminates the simulation if it gets stuck.

rtownson:
Remove infinite loops from egs++, replacing them with for loops. Add a new parameter `loopMax` that dictates how many times near-infinite loops will iterate. Upon hitting the maximum, `egsFatal` will report an error.

Add a `veryLargeFloat` parameter to replace hard-coded instances of `1e30` and such.

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
+++ b/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
@@ -546,7 +546,7 @@ public:
                 // Call to CD geometry howfar with updated region ixold. If it is aimed away from
                 // geometry, it will return ixnew = -1 and assumption a) was correct. If it is
                 // aimed into geometry, it will return ixnew >=0 and assumption b) was correct.
-                tb = 1e30;
+                tb = veryFar;
                 int ixnew = howfar(ixold,x,u,tb,0,pn);
                 // Enters geometry and at a boundary or very close to one.
                 //if( ixnew_pos >= 0 && ixnew_neg < 0 && tb_neg <= boundaryTolerance && tb_pos <= boundaryTolerance) {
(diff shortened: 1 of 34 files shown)


HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
```diff
@@ -681,7 +681,11 @@ class EGS_CDGEOMETRY_EXPORT EGS_CDGeometry : public EGS_BaseGeometry {
 do_checks:
         EGS_Vector tmp(x + u*tb);  // position at which we are inside
         // the base geometry
-        while (1) {
+        for (EGS_I64 i=0; i<1e6; ++i) {
+            if(i == 1e6-1) {
```
ftessier: Run `astyle --options=$HEN_HOUSE/scripts/egs_style_standard egs_cd_geometry.h`

ftessier:
Let's address other instances in the same commit.
```
cd $HEN_HOUSE/egs++
grep -iR "while \(1\)" *   # infinite loops
grep -iR [0-9]e[0-9] *     # large constants: should be a variable or define
```

rtownson:
I'll work on this, adding a couple new constants similar to `epsilon`.

PR #233, opened 2017-02-07
Fix bug setting wrong $EGS_HOME/bin/my_machine folder

mainegra:
BUG: In order to set LD_LIBRARY_PATH for pre-compiled egs_view my_machine
     was changed to linux-static. This affected the location of the
     application in $EGS_HOME.

This commit fixes it.


mainegra:
@ftessier this commit should be merged to develop and master ASPA to avoid users not having access to their applications once compiled!

PR #237, opened 2017-02-09
Remove end-of-line white space

ftessier:
@nrc-cnrc/egsnrc I would like to remove end-of-line (eol) white space in all the EGSnrc files, after which point we can set our editors to remove eol white space upon saving to avoid irrelevant file changes in commits. The corresponding downside of course is that this commit will register a very large number of white space changes; but at least they are all in the same place.

I created a separate commit for each directory, to facilitate reviewing the diffs. The idea would be to squash everything into a single white space removal commit before merging this in.

You can confirm that no change other than eol white space was introduced on this branch with:
```
git diff --ignore-space-at-eol $(git merge-base develop fix-eol-whitespace)
``` 
To view in the terminal the actual spaces removed, run
```
git diff -R $(git merge-base develop fix-eol-whitespace)
```

The eol spaces were removed everywhere in one go with the following command:
```
for f in $(grep -ERl -I "\s+$" HEN_HOUSE/); do 
    echo $f;
    sed -i -e's/[[:space:]]*$//' $f; 
done
```
(Note that `pieces/windows/Makefile` and `pieces/windows/clean_all.bat` both contained `^M` eol markers that were perhaps meaningful on Windows?).

What do you think? We don't **have** to do this. Do you think removing white space, especially in `.data` files, might somehow cause bugs? Can you spot any _required_ eol whitespace? We could also remove white space everywhere except in data files.

(and no, I am not just trying to improve my commit stats...)


rtownson:
I played around with it and couldn't find any bugs so far. This is one I wouldn't push into master for a while though.

ftessier:
Agreed @rtownson; this commit is meant to stay on the develop branch until next year. The master branch is now frozen until the next release, except for fixing critical bugs. @mainegra can you check if we can safely remove `^M` eol characters in `pieces/windows/Makefile` and `clean_all.bat`?

PR #240, opened 2017-02-14
Fixed issues with qmake and git

crcrewso:
This pull request will change the git ignore to ignore Makefiles produced from qmake, though this may require running `git update-index --assume-unchanged . ` in each of the folders with *.pro files. 

I was having issues with git and commits because the makefiles were from someone else's configuration. 

The Wiki building GUI's on Linux should be updated to running qmake and then make on each system. 

With the current setup if there is one deviation from defaults or if there is no linux64.conf make will fail. 

These bugs probably went unnoticed as one would not have to make changes unless one were to update the *.pro's for qt5 (which is what I was doing when this came up).


crcrewso:
I think I did something horribly wrong. I meant to only push one change to .gitignore

ftessier:
@crcrewso I don't understand: there are no Qt generated makefiles under `$HEN_HOUSE/gui` in the repository.

crcrewso:
I just double checked Master and develop, there are makefiles in gui, and in each of the subdirectories. These would have been created by qmake

ftessier:
Ok. So we probably don't want to ignore those files in `.gitignore`: we _want_ git to list them as untracked files.

crcrewso:
I see where you're going with that and I've done some more digging. You're right we want them listed as untracked. I thought git was acting normally when it was tracking all of the Makefiles but I've rebased to nrc/develop and it no longer tracks them. Did I miss a patch maybe? Sorry for confusing untracked and ignored too. 

The modules are using qmake and make in an unusual, nonstandard, and undocumented way (MAKEFILE= was not easy to find, documentation currently recommends qmake then make). Could we update the documentation to match? I'm not entirely sure that this configuration will be able to survive the qt5 transition but I'll endeavor to try.

ftessier:
@crcrewso can you point out where in the documentation we currently recommend `qmake` then `make`? I would like to fix that. Let us know about Qt5. I was not successful with Qt5 on macOS Sierra, but I believe the problem was unrelated to the EGSnrc configuration...

crcrewso:
We don't mention how to use qmake in our documentation links only to make. However that is not the standing documentation for how qmake normally works. I had to reread the last line of the make header to even figure out what was going on with how we use qmake. As such I think our documentation for making the GUI's is intended only for programmers so I would like to see a note that make will call qmake, how and why. 


As for QT5 I'm currently finding time to test the solution proposed from [StackOverflow](
http://stackoverflow.com/questions/42236316/virtual-bool-qabstractscrollareaeventfilterqobject-qevent-is-protected). I'm not that experienced with what's going on with qt's eventFilter message handling so it's taking a bit of time. Let me know what you think of it?

I've gotten egs_configure to build on WSL, Ubuntu, and Fedora. It works in the latter two as well but there is a qt5/x11 issue in the WSL. I have done zero Win32 testing on that branch though. Egs_view is in a very similar state, though it will launch in WSL.   

QT5 (well qtchooser) requires defining both QTDIR and QT_SELECT in bashrc.

PR #248, opened 2017-02-22
fixed error finding shared planes

jantolak:
Fixed an error finding shared planes when the shared edge was the last edge of a subregion. This could lead to an infinite loop trying to find an intersection. Also corrected a spelling error and removed unnecessary check of distance on line 527. Error was brought to my attention by Michael Snyder at Wayne State.


jantolak:
I should have left an example of an input file that would lead to the infinite loop. See attached. This one has a lot of histories because it was after fixing the issue. Without the fix, it only requires a few histories to run into the problem.
[test4a.egsinp.zip](https://github.com/nrc-cnrc/EGSnrc/files/793578/test4a.egsinp.zip)

PR #249, opened 2017-02-22
Implement a simple Travis-CI continuous integration script

ftessier:
Implement a simple Travis-CI continuous integration script which automatically build and configures EGSnrc on each pull request. In time we can add more sophisticated tests, but as a starting point this will at least catch basic configuration failures. @mainegra I added the make -j12 flag as a default in the configuration script; if you want to add it in the guis as well, add a commit to this branch and I will squash it with mine.

### DIFF
--- a/HEN_HOUSE/scripts/configure.expect
+++ b/HEN_HOUSE/scripts/configure.expect
@@ -1,0 +1,50 @@
+#!/usr/bin/expect -f
+
+set timeout -1
+
+# configuration filename
+set confname "expect.conf"
+if {$argc > 0} {
+    set confname [lindex $argv 0]
+}
+
+# clear environment variables
+set env(HEN_HOUSE)  ""
+set env(EGS_CONFIG) ""
+set env(EGS_HOME)   ""
+
+# spawn EGSnrc configuration script
+spawn HEN_HOUSE/scripts/configure
+
+# Information
+expect "Press enter to continue."                       { send "\n" }
+
+# Fortran compiler
+expect "Input fortran compiler:"                        { send "\n" }
+expect "Input standard compilation flags:"              { send "\n" }
+expect "Input optimization flags:"                      { send "\n" }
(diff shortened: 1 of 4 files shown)


ftessier:
Yes, the idea is to make `-j12` the default flag for `make` in the EGSnrc configuration script. I am not sure I understand how this breaks terms of service. Can you provide more details?

crcrewso:
I'll see if I can dig it up but I had thought one of the clusters had a one thread rule on the access server. Obviously this would not carry over to scheduled jobs. 

Making it configurable would be a great all around solution!

HEN_HOUSE/scripts/configure.expect
```diff
@@ -0,0 +1,50 @@
...
+
+# C compiler
+expect "Input the one you would like to use:"           { send "\n" }
+expect "Input C compiler flags to use:"                 { send "\n" }
+
+# Configuration
+expect "Input path to EGSnrc HEN_HOUSE directory:"      { send "\n" }
+expect "Choose a configuration file name:"              { send "$confname\n" }
+set timeout 1
+expect "Do you want to overwrite ?"                     { puts "no\n"; exit }
```
crcrewso: Why is this no? 

I would expect this expect to never be triggered but in the event that it is wouldn't we want it to be yes?

ftessier: I'd rather not overwrite **anything** in an automated script. If there is an overwrite condition, then the user should take care of that manually. The default conf name here is `expect.conf`, which is unlikely to collide with something user selected, but who knows? Since this expect script is in the repository, someone might use it to configure without having to answer all the questions...

crcrewso: I just read through some travis documentation. Are you using their default environments? If so, they're only dual core.

ftessier: Yes I am using their default, as far as I know...

PR #250, opened 2017-02-24
Fix #182: Increase beamnrc sync maxfield and add check

### LINKED ISSUE #182: Suggestion: increase MAXFIELD default value in SYNCVMLC
When trying to run a beam with more than 265 fields in a sequence file I get the following error:

```
Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7F82063D4E08
#1  0x7F82063D3F90
#2  0x7F8205B0649F
#3  0x436BFE in howfar_mlc_
#4  0x4416AC in howfar_
#5  0x4CF69A in photon_
#6  0x4CFF9C in shower_
#7  0x4D05D4 in beam_shower_
#8  0x4F81C7 in beam_shower_loop_
#9  0x501FB3 in MAIN__ at BEAM_06X_linux.F:?
Segmentation fault (core dumped)
```

rtownson:
Increases the `MAXFIELD` parameter for the BEAMnrc `SYNC` component modules to 1024 instead of 256. This will reduce the number of users than need to increase the parameter manually. Also adds a check to make sure that the number of fields read in from a file is less than the maximum.


ftessier:
I merged this by mistake earlier, so I reverted the change on branch develop and I am reopening as another pull request.

PR #253, opened 2017-02-26
Fix #50: Allow use of environment variables at beginning of file names

### LINKED ISSUE #50: Phase-space  and other file locations cannot be specified with a relative path
It would be great if the file locations of phase spaces and other files could be linked relatively or at least through the `$EGS_HOME` and `$HEN_HOUSE` variables. It would make switching between a prototype and production system much easier. This could be flagged as a product improvement request.

blakewalters:
Added subroutine `replace_env(fname)` to `egs_utilities.mortran` which
replaces environment variables (and '~') appearing at the beginning of
file names with their full directory paths. Subroutine is now used in
beamnrc, dosxyznrc and the RZ user codes.


crcrewso:
@blakewalters Looks great!

Have you tested the case where the filename using an environmental variable is less than 80 characters but the full file name is greater? Does it still work?

blakewalters:
@crcrewso good point!  I just passed the file names as arguments to the new replace_env subroutine without looking at their length.  It is, indeed, possible that, once the environment variables have been replaced, the resulting name will exceed the length of the original variable.  Before we merge this change, then, I'll go through the codes and make sure that the variables for all file names passed are at least character*256.

ftessier:
Rebased on develop to test configuration with travis automatically.

ftessier:
Closed by mistake; trying to fix that.

PR #256, opened 2017-03-01
Fix rest mass (prm) with value of zero in DOSXYZnrc (critical fix)

rtownson:
This is a critial fix! The bug affected at least DOSXYZnrc and resulted in an offset of the initial energy of electron beams by the value of the rest mass. The problem was the rest mass variable prm being uninitialized, so it had a value of zero.

Now, prm is initialized in the same location as rm, in egs_utilities.mortran. This fix will also solve the problem if it existed in any other codes.


ftessier:
According to `grep -iR prm HEN_HOUSE/*` this bug would have affected pretty much all applications.

blakewalters:
Just wondering if there's a problem with adding a data statement to the definition of the USEFUL common block in egsnrc.macros:

REPLACE {;COMIN/USEFUL/;} WITH
{;
   COMMON/USEFUL/PZERO,PRM,PRMT2,RM,MEDIUM,MEDOLD;
   $ENERGY PRECISION PZERO,   "precise zero"
                     PRM,     "precise electron mass in MeV"
                     PRMT2;   "2*PRM"
   $REAL             RM;      "electron mass in MeV"
   $INTEGER          MEDIUM,  "medium index of current region"
                     MEDOLD;  "medium index of previous region"
   DATA RM,PRM,PRMT2,PZERO/0.5109989461,0.5109989461,1.0219978922,0.D0/;
}

This works.  The only potential problem I can see is the hard calculation of PRMT2 required.

rtownson:
Blake's solution sounds good to me.

marenaud:
In DOSXYZnrc, does this fix only affect dosxyz-specific particle sources? (ie not shared library or phsp sources) Or will all beam model energies have to be re-tuned?

rtownson:
I believe it only affects mono-energetic sources, because the rest of the sources are initialized after the hatch call.

ftessier:
The following script helps to determine which applications are likely to have been affected by the bug:

```bash
#!/bin/bash

for f in $(find -type f); do
    has_hatch=$(grep -in -m1 hatch $f)
    has_prm=$(grep -in -m1 prm $f)
    if [[ -n $has_prm && -n $has_hatch ]]; then
        echo "======================================================="
        echo "$f:"
        echo
        grep -E --color=auto -in "call *hatch" $f
        echo
        grep -E --color=auto -in prm $f
        read -sn1
        echo
    fi
done
```

blakewalters:
Yes, it only affects monoenergetic sources.  It does not affect sources that use an energy spectrum, nor does it affect phase space or BEAM simulation sources.

ftessier:
According to the test script, beamnrc.mortran is probably affected if bcse is turned on:
```
=======================================================
HEN_HOUSE/omega/beamnrc/beamnrc.mortran:

3863:CALL HATCH; "Input the pegs generated data"

3171:            bcse_factor = 1 + bcse_factor_c*(E(np)-prm)**bcse_power_n;
3207:            bcse_factor = 1 + bcse_factor_c*(E(np)-prm)**bcse_power_n;
```

ftessier:
The following files, where `prm` is used before `call hatch`, are likely tainted by the bug:
```
HEN_HOUSE/omega/beamnrc/beamnrc.mortran
HEN_HOUSE/user_codes/cavrznrc/cavrznrc.mortran
HEN_HOUSE/user_codes/cavsphnrc/cavsphnrc.mortran
HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
HEN_HOUSE/user_codes/sprrznrc/sprrznrc.mortran
```

rtownson:
I have amended the commit to use Blake's suggested fix. It compiles and I tested that it successfully fixes the bug.

PR #259, opened 2017-03-01
Miscellaneous fixes related to CMs

blakewalters:
Fixes required to compile/run a Frankenstein
accelerator consisting of all 25 available CMs.
Includes:
1) Line lengths > 80 char
2) Format specifiers
3) Reordering of variables in CMs common block
4) Use of replace_env subroutine to allow the use of
   env variables when specifying sequence files for
   dynamic CMs.


ftessier:
Forced push to update commit message and remove a few EOL whitespaces in `XTUBE_cm.mortran`. Will merge as soon as the check has passed.

PR #263, opened 2017-03-07
QT5 migration for QT based GUI apps

crcrewso:
Egs_view, egs_configuration, egs_gui, and egs_inprz now compile and run according to QT4 to QT5 migration documentation. I've done some testing, but more is needed by those that actually use egs_inprz and egs_view.


ftessier:
Thanks @crcrewso! Please `git commit --amend` to format your commit message according to https://github.com/nrc-cnrc/EGSnrc/wiki/Git-commit-messages. And now for the more difficult question: I presume Qt5 is not widely deployed yet; so we ought to wait until we distribute this, or else create a separate gui directory for the Qt5 versions?

crcrewso:
I'm away from the desk so I'll answer the more difficult question first.  

Qt5 has been around for a few years. QT4 was declared no longer supported either last year or the year before. We could choose to keep distributing the QT4 binaries for another year but I'm hesitant to hold off too much longer since some distributions will probably be removing their QT4 support from mainline in the not to distant future.

ftessier:
Thanks for the update, we'll look into it and consider moving to Qt5. At any rate I will mark it as Release 2018 material!

crcrewso:
@ftessier 

I updated the commit message, please let me know if it is still not perfect. 

When reading the wiki page you linked, I noticed this line

> This only works for the most recent commit, and only until it is pushed upstream to the origin of your local branch.

I found on [StackExchange](http://stackoverflow.com/questions/10510462/force-git-push-to-overwrite-remote-files) a great hint on how to force the amendment when there are no code additions

rtownson:
I'm closing this PR because the commit is included in PR #417. Thanks for your contribution @crcrewso!

PR #274, opened 2017-03-30
Fix rz compile warnings about radial_bins indexing

rtownson:
The `dosrznrc` and `flurznrc` codes both compiled with warnings about the indexing of `radial_bins`. Increasing the array declaration size in `grids.mortran` to be `$MAXRADII` instead of `$MAXRZ` fixed the warnings. I do not believe this impacted results in any way.

Thanks to Mathew lc for reporting this and suggesting a solution.

### DIFF
--- a/HEN_HOUSE/utils/grids.mortran
+++ b/HEN_HOUSE/utils/grids.mortran
@@ -26,6 +26,7 @@
 "                                                                             "
 "  Contributors:    Blake Walters                                             "
 "                   Iwan Kawrakow                                             "
+"                   Reid Townson                                              "
 "                                                                             "
 "#############################################################################"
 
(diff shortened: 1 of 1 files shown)


mainegra:
dosrznrc and flurznrc have the right dimensions for these variables, i.e.:

RADIAL_BINS($MAXRADII),DEPTH_BINS($MAXZPLANE)

The conflict is introduced in the file `HEN_HOUSE/utils/grids.mortran` where they are defined as:

RADIAL_BINS($MAXRZ),DEPTH_BINS($MAXRZ)

To avoid unnecessary large arrays when not needed, one should fix the mismatch in `HEN_HOUSE/utils/grids.mortran` rather by setting the above variables to the right dimensions as they are defined in the user codes.

rtownson:
Thanks for the better solution @mainegra! I have amended the commit.

ftessier:
Closing pull request because this is fixed in #289.

PR #276, opened 2017-04-10
Fix compilation warnings in egs++

HubertHo:
Resolve the `reorder` and `unused-variable` warnings given when compiling egs++. Unused variables that were used for debugging have been commented out rather than removed. The `cmp` linux command revealed no differences between object files compiled before and after these modifications.


crcrewso:
Why are err3 and err5 not used? Are they never status checked for failure?

ftessier:
Cleaned up and reopened as pull request #301. The comment by @crcrewso about `err3` and `err5` has been addressed by cleaning up error reporting for replicas in the base geometry.

PR #278, opened 2017-04-19
Fix #269: Fix negative usteps with SYNCJAWS.

### LINKED ISSUE #269: Negative usteps in DOSXYZ isource=20 with IAEA phsp source through shared library geometry
Hello.

I'm using current `fix-misc_CM_fixes` branch with changes included in Fix #186 as well.

I'm experiencing negative ustep errors (+1000 always --> stopping the simulations) with DOSXYZnrc isource=20 runs using IAEA phsp source through shared library geometry. I don't know is this related to issues #187 or #90, but for some reason the errors are generated when using similar Varian-generated phsp source that I assume @crcrewso is using in #187 with header info:
```
$RECORD_CONTENTS:
    1     // X is stored ?
    1     // Y is stored ?
    0     // Z is stored ?
    1     // U is stored ?
    1     // V is stored ?
    1     // W is stored ?
    1     // Weight is stored ? <--- Changed from 0 to 1 by addphsp
    0     // Extra floats stored ?
    1     // Extra longs stored ? <--- Changed from 0 to 1 by addphsp
    2     // LATCH EGS variable stored in the extralong array [ 0] <--- Added by addphsp

$RECORD_CONSTANT:
    26.7000     // Constant Z
    1.0         // Constant Weight <-- Manually added by me according to info on Varian webpage

$RECORD_LENGTH:
29   (21 in the original Varian phsp file)
```
One thing I suspect is the number of particles in the large phsp file, which exceeds 2^31-1 (the number is 2367622510). Could this be the reason for negative usteps? However addphsp (version 2016) concatenated this large file without problems and no warnings or errors are generated anywhere else.

What is also weird is that I have run BEAMnrc ISOURC=21 simulations with the phsp file above without any problems. 

One discrepancy I found is that the addphsp for the latest release prints `Warning: IAEA format phsp file does not store LATCH` when concatenating IAEA phsp files, but the manual on p. 107 says "In the case of BEAMnrc, two extra long integers, LATCH and the number of primary histories between this particle and the last particle scored, are always stored."

I have another IAEA phsp file for other Varian linac model with following header information:
```
$RECORD_CONTENTS:
    1     // X is stored ?
    1     // Y is stored ?
    1     // Z is stored ?
    1     // U is stored ?
    1     // V is stored ?
    1     // W is stored ?
    1     // Weight is stored ?
    1     // Extra floats stored ?
    2     // Extra longs stored ?
    3     // ZLAST variable stored in the extrafloat array [ 0]
    1     // Incremental history number stored in the extralong array [ 0]
    2     // LATCH EGS variable stored in the extralong array [ 1]

$RECORD_CONSTANT:

$RECORD_LENGTH:
41
```
with 2080480392 particles. The Z is anyhow constant value for all particles. Simulations with the file run fine.

While reading the current BEAMnrc manual I can't find out which `RECORD_CONTENTS` are really required and which ones are discarded even though they would be included in the header.

blakewalters:
The ability to have time-varying Z-positions of jaws means
that Z_min_CM(ICM_$SYNCJAWS+1) must be redefined for every initial
history.  Previously, Z_min_CM(ICM_$SYNCJAWS+1) was fixed at the
max. Z-position of the jaws.  This meant that particles transported
backwards into SYNCJAWS from the bottom might, on the next step, find
that they actually weren't in the CM at all.  The bug also
affected DYNJAWS.


ftessier:
Added @blakewalters as a contributor in `DYNJAWS_macros.mortran`. Will squash and merge when Travis passes.

PR #284, opened 2017-05-05
Fix #205: bug in egs_chamber radiative splitting and RR

### LINKED ISSUE #205: Bug with bremsstrahlung splitting in egs_chamber
There is a problem with the current implementation of brem splitting in egs_chamber.cpp
The variable for number of split radiative photons is read from the .egsinp file (line 938 of egs_chamber.cpp):

        //
        // ******** radiative event splitting
        //
         int csplit=1;
         if( !vr->getInput("radiative splitting", csplit) && csplit > 1) {
            egsInformation("\n => initScoring: splitting radiative events %d times ...\n", csplit);
           the_egsvr->nbr_split = csplit;
}

However, this variable is overwritten when the radiative event comes from an electron which has survived RR (line 1974):

    // split up fat electrons' radiative events
    // i.e. electrons that survived range-rejection RR
    if( (the_stack->latch[np] != 0) &&
        (iarg == BeforeBrems || iarg == BeforeAnnihFlight || iarg == BeforeAnnihRest) ) {
        the_egsvr->nbr_split = the_stack->latch[np];	// split photons up
        return 0;
}

After RR has been handled, the brem splitting number is set back to 1 (line 1998):
 the_egsvr->nbr_split = 1; 

As I understand it, this means that after the first photon interaction with cross-section enhancement occurs, bremsstrahlung splitting is effectively turned off.

A colleague (Marc-Andre Renaud) and I were able to fix this by declaring a private EGS_ChamberApplication class variable, which is then set to the value of csplit. After the RR, the brem splitting number is set back to this class variable instead of 1.

rtownson:
Fix a bug where radiative splitting was effectively turned off after the first photon interaction with cross-section enchancement. Now nbr_split is set to csplit after RR instead of 1.

### DIFF
--- a/HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp
+++ b/HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp
@@ -659,6 +659,7 @@ private:
 
     EGS_Float        fsplit;    // photon splitting number
     EGS_Float        fspliti;   // inverse photon splitting number
+    int              csplit;    // radiative splitting number
 
     /*! Range rejection flag
       If set to 0, no range rejection is used
@@ -939,8 +940,8 @@ int EGS_ChamberApplication::initScoring() {
         //
         // ******** radiative event splitting
         //
-         int csplit=1;
-         if( !vr->getInput("radiative splitting", csplit) && csplit > 1) {
+        csplit=1;
+        if( !vr->getInput("radiative splitting", csplit) && csplit > 1) {
             egsInformation("\n => initScoring: splitting radiative events %d times ...\n", csplit);
            the_egsvr->nbr_split = csplit;
         }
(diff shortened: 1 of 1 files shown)


ftessier:
@rtownson I am not sure about this one. The comment on lines 2001-2003 indicates that we intend to reset splitting for daughter particles, i.e., only the first radiative interaction should be split. Otherwise, things can get carried away quickly and fill the particle stack. I suggest we leave it as is, or else add an option to split all interactions, or only the first one; with the latter as the default.

rtownson:
You're right, this needs more thought. I'm not sure that radiative splitting is ever turned back on after the first RR, even for new source particles. However, I don't quite understand the interplay of the `latch` and `nbr_split` at the moment.

ftessier:
@rtownson I changed the commit slightly to reset the `nbr_split` radiative splitting number to the input value at the start of each shower. The reason is that if it is not set to one after the first split, then for high energy simulations splitting can get carried away and fill the particle stack quickly!

rtownson:
Looks good to me!

PR #282, opened 2017-05-02
Make EADL atomic relaxations consistent

mainegra:
This pull request addresses issues with the implementation of a more detailed atomic relaxation process and makes it officially the default in EGSnrc. Users wanting to revert back to the previous implementation will have to set the macro `$EADL_RELAX` to `.false.` in file `egsnrc.macros`. 
The mismatch between the set of binding energies used for photo-electric, Compton, and electron impact ionization (EII) interactions and the one used for the actual atomic relaxation creates a small inconsistency violating energy conservation. As reported by @FabioCpp in #277 , a simple simulation of a mono-energetic photon source in an infinite geometry reported deposited energy values that were different from the initial energy by almost 0.5%. Here at NRC a simulation of a 1 MeV photon source in an infinite geometry using `cavity` produced the following results for several elements and media:

|Medium  |  Z |   No EADL  |    EADL |
|-----------|-----|----------------|-------|
| **$EADL_RELAX**| | **.false.** | **.true.** |
|oxygen   |    8 |  1.00000000 |  0.9999992|
|manganese |  25 |  1.00000000 |  0.9999706|
|copper  |    29 |  1.00000000 |  0.9999541|
|tin        | 50  | 1.00000000 |  1.000112|
|terbium  |   65 |  0.99999921 |  _**1.000694**_|
|lead     |   82 |  0.99999831 |  _**1.001839**_|
|einsteinium |99 |  0.99992831 |  **_1.003197_**|
|water |       |    1.00000000  | 0.9999993|
|air     |       |  1.00000000 |  0.9999949|

Another issue with the current implementation is that the x-ray lines will correspond to the EADL radiative transitions regardless of the selection of the photon cross sections. Several changes and corrections have been made to resolve the energy conservation violation and make the x-ray lines correspond to the radiative transitions expected from the photon cross section compilation used. Here is a brief summary of the main changes:

- Make use of a consistent set of binding energies throughout the code. Binding energies are selected depending on the photon cross section libraries requested by the user, i.e., `xcom` or `epdl`. Transition probabilities are still taken from the EADL library. As a consequence, the position of the x-ray lines will match either the EADL radiative transitions (`epdl`) or the radiative transitions from the Table of Isotopes (8th Edition) (`xcom`).
- When `$EADL_RELAX` is true, allow all energetically possible transitions available in the EADL library.
- When `EADL_RELAX` is true, set `shn_array`, the shell type used during Compton interactions, to the local shell number rather than to the shell type as defined in the original implementation.
[...]

PR #285, opened 2017-05-05
Fix sub-threshold relaxation depositions

rtownson:
Previously some of the sub-threshold energy depositions in both relaxation routines contributed only to `edep_local` as though the contribution came from an electron (previously `iarg=$SAUGERTRA`). For photo-electric & compton initiated relaxations, we should have been triggering a photon `edep_local` contribution. Now the relaxation algorithms differentiate correctly between which type of particle initiated the relaxation using `relax_initiator_q`, a new variable in `COMIN/RELAX-DATA`. 

Additionally, the `RELAX-CUTOFF` is now used for the vacancy creation check, rather than `ECUT` or `PCUT`. This fixes a dependency of `g` on `ECUT`. Previously, changes in ecut resulted in a different (shorter) chain of relaxations, meaning that some radiative components were not sorted out correctly to calculate `g`. Unfortunately, this fix comes with a performance hit when using higher ecuts (worst case about 9% slower).

At the end of the compton routine there was an `edep` contribution that was not accounted for using `edep_local`. This has been fixed, and a call to `$SPHOTONA` added.

The sub-threshold Auger and fluorescent iarg calls have been renamed to `$SELECTRONA` and `$SPHOTONA`, respectively, to reflect that the energy depositions may also result from the particle initiating depositions.

Comments in the relaxation routines have been improved.

This also includes a commit that updates `g`. The changes are mostly cosmetic but there is one bug fix for multiple energy calculations.

Thanks to Dave Rogers for his contributions!


rtownson:
Please merge PR #282 first, it will need some updates that I will do in this PR.

rtownson:
This is ready for merging into `develop` now.

PR #288, opened 2017-05-19

seirios:
I noticed a lack of the RANLUX random number generator in egs++, so I implemented a class EGS_Ranlux for it based on [GSL](https://www.gnu.org/software/gsl)'s integer implementation (gsl_rng_ranlux), which is pretty much the same as the original Fortran implementation in EGSnrc. I haven't done extensive testing, but it seems to be working fine.

I think the next step could be to implement it using Luscher's floating-point (SSE optimized) code from his [own implementation](http://luscher.web.cern.ch/luscher/ranlux/ranlux-3.3.tar.gz) or based on [GSL](https://www.gnu.org/software/gsl)'s floating-point implementation (gsl_rng_ranlxs* and gsl_rng_ranlxd*).


mainegra:
@seirios can this pull request be changed to be against develop rather than master?

PR #289, opened 2017-05-19
Fix cavity e fluence

mainegra:
Although [PIRS-898](http://nrc-cnrc.github.io/EGSnrc/doc/pirs898/index.html) states that `cavity` can calculate the e-/e+ fluence, this is not entirely correct. The quantity that is actually calculated seems to correspond more to the product of the fluence times the stopping power. Now the actual charged particle fluence is calculated. This change has been documented in [PIRS-898](http://nrc-cnrc.github.io/EGSnrc/doc/pirs898/cavity.html) in this pull request. A bug in `FLURZnrc `is also fixed which caused an artifact in the first energy bin.

The fluence calculation in cavity can be now done using two methods:

- Algorithm already in place in cavity but adding a division by the stopping power for the lower edge energy L(E_i) of the ith bin. See cavity's documentation in [PIRS-898](http://nrc-cnrc.github.io/EGSnrc/doc/pirs898/cavity.html) for more details.

- Algorithm used in FLURZnrc for comparison
    
The first one is more accurate as it does not make the assumption of stopping power constancy along the CH step, which is made in the `FLURZnrc` algorithm.
    
By default the first algorithm is used. If one wants to switch to a FLURZnrc-like calculation of the fluence, the following input key should be used in the fluence scoring input block:
    
    :start scoring options:
        ...
        :start fluence scoring:
            ...
            method = flurz # stpwr (default) or flurz
            ...
        :stop fluence scoring:
        ...
    :stop scoring options:
    
Range rejection is disabled for a FLURZnrc-like calculation of the fluence and this calculation type relies on the knowledge of the particle's step size. For the calculation type using stopping powers, a warning is issued as proper selection of `ESAVE` is required for accurate results. In general avoid using range rejection when calculating charged particle fluence!

Here is a comparison between `cavity` and `FLURZnrc` for an electron beam of 5 MeV:

eflu_case_1a_vs_original

And a comparison between `cavity` and `FLURZnrc` for a 1.25 MeV photon beam:

eflu_case_2a_vs_original


ftessier:
Just updated the commit messages and fixed a small typo in the doxgen file. I suggest we keep the commit history and merge into `develop`.

mainegra:
@Kawrakow you are right, I will change it to do the division at the end when outputting the results! Thanks for pointing this out!

PR #292, opened 2017-05-20

seirios:
I noticed a lack of the RANLUX random number generator in egs++, so I implemented a class EGS_Ranlux for it based on [GSL](https://www.gnu.org/software/gsl/)'s integer implementation (gsl_rng_ranlux), which is pretty much the same as the original Fortran implementation in EGSnrc. I haven't done extensive testing, but it seems to be working fine.

I think the next step could be to implement it using Lüscher's floating-point (SSE optimized) code from his [own implementation](http://luscher.web.cern.ch/luscher/ranlux/ranlux-3.3.tar.gz) or based on [GSL](https://www.gnu.org/software/gsl/)'s floating-point implementation (gsl_rng_ranlxs* and gsl_rng_ranlxd*).

### DIFF
--- a/HEN_HOUSE/egs++/egs_rndm.cpp
+++ b/HEN_HOUSE/egs++/egs_rndm.cpp
@@ -249,7 +249,7 @@ void EGS_Ranmar::setState(EGS_RandomGenerator *r) {
     copyBaseState(*r);
     EGS_Ranmar *r1 = dynamic_cast<EGS_Ranmar *>(r);
     if (!r1) {
-        egsFatal("EGS_Ranmar::setState: attampt to set my state by a non EGS_Ranmar RNG!\n");
+        egsFatal("EGS_Ranmar::setState: attempt to set my state by a non EGS_Ranmar RNG!\n");
     }
     ix = r1->ix;
     jx = r1->jx;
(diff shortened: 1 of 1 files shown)


seirios:
Sorry for creating a new pull request, but I changed it to be based off my develop branch as well.

ftessier:
Thank you @seirios! Can you comment on the source code from a licence standpoint? Did you write it, say, based on Lüscher's 1994 paper or `egsnrc.mortran`, or did you work off Lüscher's code or the GSL implementation?

seirios:
I based the code on the GSL implementation. I also looked into egsnrc.mortran, but I'm not sure if any bits made it to the code.

mainegra:
@ftessier what ever happened to this PR? Were there issues with Open Source licenses? @seirios did you test this implementation and compared it with the current high-resolution modification of ranmar? I am asking now because I think we should make ranlux or the high-resolution version the _default random number generator_ in EGSnrc. There are situations were ranmar simply gives wrong results, albeit not many. But for robustness it is a better approach to have someone switch to the original ranmar intentionally than having users unaware of this issue obtaining wrong results. **This is espcially important as more simulations move to low energies and small scales.**

crcrewso:
This is a naive question, but Linux RAND and GCC Rand have seen a lot of improvements over the last few years, As such is RANLUX still necessary? (For backwards consistency? for matching reality?) If so, could we include the whole GSL so we're not maintaining our own code in parallel?

mainegra:
@crcrewso I agree with that. The only issue would be if there is a lot of repetition in terms of the tools that EGSnrc uses. So, one would have to comb through the system removing duplicates such as the random number generator. And, it is also a matter of keeping EGSnrc as lean as possible without dependency to other libraries which might be easily available on some OSs but not in other. And might also have version incompatibilities. See for instance the issues with the QT libraries for our GUIs. Although QT has gotten much better since QT4 and QT5.

ftessier:
Agreed that the default generator should be "high-resolution", with decent performance. But ranmar and ranlux are such old generators, no? They don't even make in on any modern list of the optimal rngs these days. If we are going to move to a new default, I suggest what seems to be the current state of the art, the [xoshiro generators](http://prng.di.unimi.it/), specifically the `xoshiro256++` genrator, clocking in at ~0.86 nanoseconds per 64 bits, with a periodicity of 2^256 - 1 and no failures of the BigCrush test suite. There is no issue of copyright since the code is placed in the public domain. The code is only about 60 lines in C.

2256 − 1
--

mainegra:
@ftessier the issue of avoiding a potential bug should be separate from the inclusion of a new, better rng in EGSnrc. Setting the default rng to our best available option is a simple modification that avoids serious errors in some situations. Modernizing our rng's should be a new feature PR.

ftessier:
Agreed that we should make ranlux the default on the mortran backend. But inasmuch as there is currently no implementation of ranlux on the egs++ side, then implementing ranlux or xoshiro equivalent in terms of simplicity etc. I propose we:

- test and merge this ranlux implementation by @seirios 
- make ranlux the default on the mortran and egs++ side
- implement the xoshiro rng in a separate pull request, indeed; and make it the default?

My only concern with switching to ranlux is that simulations will be significantly slower (what is the relative efficiency of ranmar vs ranlux?). That is why I am tempted to jump to xoshiro as the default as soon as possible, to provide both a better and _much faster_ rng.

ftessier:
About the "high resolution", this does not have to do with the quality of the rng, but the number of bits. Why not then stick with ranmar, but set the default to "high resolution"? **Is that an acceptable default?**

mainegra:
The question remains: _Who is going to bell the cat?_ ;-) Meanwhile we can avoid potential errors by simple changing the current defaults. In terms of time penalty, I observed a 4% cpu time increase when using the high resolution `rng` with `egs_app` and `egs_kerma` for a simple calculation. This is not too onerous in myt opinion.

ftessier:
Yes, indeed. We change the default to high resolution ranmar! I thought we thinking about changing the default to ranlux... But of course the high resolution ranmar ought to be the default, thanks for the timing test. Upon reviewing my correspondance, I realize that @MartinMartinov  had pointed it out in 2015: he solved his discrete sampling problem at small scale by turning on the high resolution option of ranmar. Go for that! So defer this discussion to #594 for changing the default to **ranmar high-resolution** everywhere, and continue this thread here solely for the egs++ implementation of ranlux by @seirios .

mainegra:
> About the "high resolution", this does not have to do with the quality of the rng, but the number of bits. Why not then stick with ranmar, but set the default to "high resolution"? **Is that an acceptable default?**

@ftessier I haven't looked inside the Mortran `ranmar` implementation to check whether there is the option to set it to _high resolution_! When I have a chance I will  and then perhaps create a PR setting the _high-resolution_ ranmar as default all across EGSnrc/BEAMnrc.

ftessier:
Ok I will look at the @seirios implementation of ranlux on the egs++ side, to make this the default in EGSnrc on both the mortran and egs++ frontends, pending ranlux efficiency evaluation. I will also implement the "high-resolution" (48 bits) ranmar on the mortran side (although I don't like the way this is generated from two sequential 24-bit numbers from the generator). Finally, I will propose the [xoshiro256++](http://prng.di.unimi.it/) generator via another pull-request.

mainegra:
I like that @ftessier !
Why not xoshiro256+? It's 15% faster with similar properties!

ftessier:
Yes, you are right, we could implement `xoshiro256+` for floating point numbers (53 mantissa bits), and perhaps `xoshiro256++` if we need 64-bit integers (I don't think we do!).

seirios:
Hi there, sorry for the late reply.
@mainegra I tested this RANLUX implementation with a small program that computed dose deposition in a voxelized grid, and it gave very similar results to DOSXYZpp for that use case. I haven't compared it to high-res ranmar, as it's been a while since I worked with EGSnrc.
FYI: The reason for this PR is that I like RANLUX for its mathematically proven properties, although it's slower than other generators, which may be undesirable in practice.
Finally, to add my grain of salt, have you looked at the MIXMAX generator (https://mixmax.hepforge.org)? According to ROOT's documentation it's fast (6 ns/call), and it's the default in CLHEP/Geant4, so it may be worth looking at.
Cheers!

ftessier:
Thank you @seirios for this contribution, and the reference to the mixmax generator, will look at it! In another discussion (https://github.com/pymedphys/pymedphys/issues/399), @SimonBiggs mentioned an article there which reports (transient) initialization issues with ranlux; I was surprised about this, and must admit I have never given much thought to rng initialization. I don't think it matters for EGSnrc (it is only a transient problem), but thought I'd mention it here for the record.

mainegra:
@seirios thanks for the comments. I also am very fond of ranlux! Used it quite a bit back in the 90's when it was considered state-of-the-art!
But the xoshiro256 generation of rngs seem to have much better statistical qualities and are much faster! Xoshiro256+ is quoted at 0.76 ns per rn!
@ftessier I support your suggestion of going ranlux on both ends (Mortran and C++) and adding the note about the faster, riskier options. And we should do a mini timing study of ranlus with luxury level 1 (default).

ftessier:
Interesting and entertaining feud between the xoshiro and pcg generator authors! From there I was led to [a talk by pcg author Melissa O'Neill](https://www.youtube.com/watch?v=45Oet5qjlms) with an great historical perspective and a simple explanation of what pcg does.

ftessier:
On pcg vs xoshiro: after additional reading, I side with the `xoshiro` generator, for its dead simple code, its speed, and its ability to jump ahead to generate independent random sequences for parallel simulations. In terms of randomness _quality_, I defer to the mathematicians; to me it suffices that it passes the TestU01 BigCrush test.

ftessier:
Only thing left as an implementation detail is how to sample doubles uniformly on [0,1). Do we just take the 53 high bits from the xoshiro256+ uint64 result and plug it into the mantissa of a double? Other suggestions welcome.

mainegra:
@ftessier I see a larger project on RNGs in EGSnrc shaping up .... any computer science student interested ?

ftessier:
Another great reference (blog post) on ranlux: https://christoph-conrads.name/faster-ranlux-pseudo-random-number-generators/. I was surprised to learn there that the standard 24-bit ranlux implementation is about twice as slow on X86-64 as a proper 64-bit implementation (see the first table). Really good background information in there too!

ftessier:
@seirios your `develop` branch for this pull request has been deleted, apparently. Hence I will close the PR. Please reopen it if you still have the original branch for your `EGS_Ranlux` class somewhere! As mentioned in the discussion, the implementation was single-precision, but it would remain useful as a starting point for the double-precision version!

seirios:
Hi @ftessier , sorry for that. I moved to GitLab some time ago and removed by EGSnrc fork from here. I've now resubmitted the pull request here https://github.com/nrc-cnrc/EGSnrc/pull/702. Cheers!

PR #295, opened 2017-05-26
Fix a bug in cavity for non fluence calculations from PR #289

rtownson:
Add a check to see if the array Rho has been defined. This avoids a segmentation fault that occurred for non fluence calculations. This bug was introduced in PR #289.


ftessier:
@rtownson I squashed the two commits into one. I also did not add your name to the Contributors, is that ok with you? I will "Squash and merge" as soon as travis is done.

rtownson:
Yes I do not need to be on the contributor list, this was pretty minor.

PR #298, opened 2017-05-30
Fix bug in photoelectric routine

mainegra:
When `$EADL_RELAX = true` and the photon interacts with a shell
below `L3` the photon energy was being deposited locally rather
than emitting a photo-electron with the photon energy. This is
corrected here.

Similarly, when the photon energy falls below the binding energy
`binding_energies($MXSHELL,iZ)`, which for `$EADL_RELAX = true` is the
binding energy of the `M2` shell, one should also emit a photo-electron
rather than deposit the energy on the spot.


ftessier:
Removed end of line white space and formatted commit message. Ready to squash and merge as soon as @blakewalters reviews.

ftessier:
Apologie @blakewalters I forgot to wait for your review! Let me know if you have any comment or issue with this merge.

blakewalters:
Looks good!  Go ahead...Wait, you already did!

PR #301, opened 2017-06-06
Fix compiler warnings for the egs++ library

ftessier:
Adjust the egs++ library source code so that it compiles without warnings under gcc with the following options: `-O3 -g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -Wno-maybe-uninitialized`. Specifically, address instances of the following warnings:
```
-Wparentheses
-Wreorder
-Wunused-value
-Wunused-variable
-Wunused-but-set-variable
-Wreturn-type
```
Also address a multi-line comment warning in `egs_cones.h`, and clean up error reporting for replica input in the base geometry class.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
@@ -25,6 +25,7 @@
 #
 #  Contributors:    Frederic Tessier
 #                   Reid Townson
+#                   Hubert Ho
 #
 ###############################################################################
 #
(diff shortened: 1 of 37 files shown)


randlet:
This is great.  Can I ask why you used `-Wno-sign-comparisons` and `-Wno-maybe-uninitialized`?  The latter in particular can save your bacon and off the top of my head I can't see a good reason to exclude that warning.

randlet:
For example in [this commit](https://github.com/nrc-cnrc/EGSnrc/commit/d7d40a51eb68fa18838d83e3bb048f47b130b431) there was a bug caught where a variable was being used uninitialized.  I caught it because I generally always use -Wall.

In general, I wish -Wall was made a default compiler option in the configure script.

ftessier:
The code compiles cleanly with no optimization, but I want to compile with `-O3` because that may reveal additional warnings. However, it also raises a lot of `-Wmaybe-uninitialized` warnings. So perhaps it would be better to use `-g -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare` as a better baseline.

ftessier:
I agree that `-Wall` should be the default. I will look again at the `-Wmaybe-uninitialized` warnings under optimization to see if I can quell them.

randlet:
Yeah, that would be better. Any uninitialized warning feels to me like either an existing bug, or a bug waiting to happen in future.   Anyways, very happy to see stuff like this being cleaned up :)

ftessier:
Ok, agreed. Turns out the `-Wmaybe-uninitialized` warnings were all due variables that could simply be initialized to 0. Updated the pull request. It now compiles cleanly with `-g -O3 -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare` (and also without optimization).

ftessier:
Minor changes upone a final `astyle` run.

ftessier:
Just rebased on `develop`.

PR #302, opened 2017-06-08
Fix #296: SYNCHDMLC not available in BEAMnrc create_accel

### LINKED ISSUE #296: SYNCHDMLC gui not available in BEAMnrc gui
It appears a gui has been written for the SYNCHDMLC component module, but it does not appear as an option when specifying a new accelerator in the BEAMnrc gui. This is despite being referred to in `beamnrc_gui.tcl` exactly the same way as other modules like as CONESTAK and FLATFILT. Is it possible this is because "synchdmlc" is 9 characters long? (CM Identifier is restricted to eight characters, maybe the same is true for CM Name.)

rtownson:
Fix a bug where SYNCHDMLC was not available as an option when creating a new accelerator in the BEAMnrc gui. This was due to the number of CMs being hard-coded in create_accel.tcl. Now the length of the cm_names array is determined correctly.


ftessier:
Just tweaked the commit message. Title was too long.

PR #316, opened 2017-07-26
Add option to output .3ddose files to EGS_DoseScoring

blakewalters:
The EGS_DoseScoring ausgab object has been modified so that
the user now has the option to output doses to a file in
.3ddose format (i.e. standard format for DOSXYZnrc output)
for any EGS_XYZGeometry(ies) in their simulation geometry.
See details in egs_dose_scoring.h.

This is accompanied by a small change to
EGS_AdvancedApplication::finishSimulation().  At the end
of a parallel run, i_parallel is now set to 0.  This allows
us to use EGS_Application::getIparallel() to query whether
or not this is the last job of a parallel run and, only then,
to output a .3ddose file containing the combined results.


ftessier:
@blakewalters see the second commit 8e74eee for the suggested changes. If you agree then I can squash it onto the first.

PR #320, opened 2017-07-31
Fix #318: Allow extension on input argument

### LINKED ISSUE #318: Allow .egsinp extension when parsing command line arguments
This is a feature request. When running an application from the command line, it would be nice if we could leave in the .egsinp extension of the input file. For example:

`tutor7pp -i test1.egsinp
`
Right now, we get the following result:

```
***************** Error: 
 input file /Users/Marc/clrp/egs_home/tutor7pp/test1.egsinp.egsinp does not exist 

```

I feel like it should be trivial to test for the existence of the .egsinp extension on the argument, but I couldn't quite get it working straight when I took a shot at it (it seems like the command line arguments get parsed in a few different places, maybe?).

rtownson:
Allow the `.egsinp` extension in the `-i` argument instead of requiring no extension.

For example, now you can run both `c++` and `mortran` applications with:
```
myApp -i myInput.egsinp
```
instead of being required to use
```
myApp -i myInput
```
I think this was a common beginner mistake and trivial to address. Fix #318.

### DIFF
--- a/HEN_HOUSE/src/egs_utilities.mortran
+++ b/HEN_HOUSE/src/egs_utilities.mortran
@@ -362,10 +362,11 @@ IF( lnblnk1(input_file) > 0 ) [
     IF( l + l1 + l2 > 1024 ) [
         $egs_fatal(*,'input file name (including path) is too long ',l+l1+l2);
     ]
+    ex = egs_strip_extension(input_file,'.egsinp');
     tmp_string = $cstring(ucode_dir) // $cstring(input_file) // '.egsinp';
     inquire(file=tmp_string,exist=ex);
     IF( ~ex ) [
-      $egs_fatal(*,'input file ',$cstring(tmp_string),' does not exist ');
+        $egs_fatal(*,'Input file ',$cstring(tmp_string),' does not exist.');
     ]
     $AVAILABLE_UNIT(i_input,tmp_string);
     open(i_input,file=tmp_string,status='old',err=:open_input_error:);


mchamberland:
Thanks for the quick turnaround, @rtownson!

mainegra:
This change allows ANY existing file to be read as an input file. There is nothing especial about the `egsinp` extension other than for clarity when organizing the large number of files one has to deal with during mass production runs. But I guess we could leave it up to the user to decide on the naming convention they want to have for their files.
**One caveat though:** if the option -o output_file is not used, then output files such as  *.egslog, *.egslst, etc. will look like input_file.egsinp.egslog, input_file.egsinp.egslst, etc. if using the `egsinp` extension or like full_input_file.egslog, etc.
The code dealing with this is located between lines 632 and 648 in `egs_utilities.mortran`.

ftessier:
Perhaps ideally the code would simply strip the `.egsinp` extension from the file name before line 365? The it is appended as before on line 365. For example, in my own bash scripts I often use `${filename%.egsinp}.egsinp` to this end. This way the "requirement" for a `.egsinp` extension sticks, and the output files names are ok.

rtownson:
Thanks for the input, and good catch Ernesto! I think I have fixed it by doing as Fred suggested. The `.egsinp` extension is enforced now.

mchamberland:
For what it's worth, my use case for this is using the tab key to auto-complete on the command line and not having to delete the `.egsinp` when there is no other file that shares the same name. ;)

HEN_HOUSE/src/egs_utilities.mortran
```diff
@@ -365,7 +365,13 @@ IF( lnblnk1(input_file) > 0 ) [
     tmp_string = $cstring(ucode_dir) // $cstring(input_file) // '.egsinp';
     inquire(file=tmp_string,exist=ex);
     IF( ~ex ) [
-      $egs_fatal(*,'input file ',$cstring(tmp_string),' does not exist ');
+      ex = egs_strip_extension(input_file,'.egsinp');
```
ftessier: @rtownson can't we simply put this strip before line 365 (i.e., before adding `.egsinp` again?). The code would be simpler...

ftessier:
Merely adjusted commit title: lowercase 'a'.

Issue #314, opened 2017-07-20
egs_chamber either produces zeros or exceeds MXSTACK


ojalaj:
I've been testing/learning to use egs_chamber. I've used a simple test file having a cylinder water phantom with a thin disc of air, where a collimated beam from a linac (IAEA phsp file) hits. I've also edited the example files for testing purposes - the one with simple Farmer chamber using either no VRTs at all or then using those ones that are utilized in test files (RR/XCSE/IPSS).

I've run simulations on two Linux clusters - one using the `develop` branch from March and the other one using the latest `develop` branch. In the former cluster I'm able to run the two test files with no VRTs with expected results, but when running the Farmer chamber example with VRTs, I get `In subroutine ANNIH_AT_REST stack size exceeded!` error. I've already previously increased the `MXSTACK` to 900000 to get other test files running, but now when I increase `MXSTACK` to 950000, all three different test files run without errors, but produce zero results with 100% uncertainty.

What is interesting, is that on the other cluster I needed to decrease `MXSTACK` to 10000 in order to get expected results from first two test files with no VRTs, but with the third test case, I get `In subroutine ANNIH_AT_REST stack size exceeded!`. When I increase `MXSTACK` to 100000, all three different test files run without errors, but produce zero results with 100% uncertainty.

So basically there are two problems:

1. Why there is such behaviour, i.e., with `MXSTACK` values below certain level I get expected results, but after applying VRTs, it needs to be increased, but all test files start producing zero results.

2. Why there is so large difference in the `MXSTACK` "threshold" values between two clusters/develop branch versions?

This behaviour has already been reported in G+ forum, but no solutions/explanations were provided.

mainegra:
[...]

PR #323, opened 2017-08-09
Fix two bugs in iaea_phsp_source

blakewalters:
1.  Arrays extrafloat_type, extralong_types, extrainttemp,
    extrafloattemp, could be initialized with zero dimension if
    the IAEA phsp did not store any addition int (e.g. latch)
    or float (e.g. zlast) variables for each particle.  These
    arrays are now given fixed dimension MAXEXTRAS.

2.  Source had the temerity to actually assign any latch
    value read from the phsp file to the incident particle, not
    realizing that, at least in the case of egs_chamber, this
    variable is used for setting splitting no.  Now, latch is
    set to 0, similar to egs_phsp_source.  This partially fixes
    the problem that @ojalaj had in Issue #314.


ftessier:
Merely edited the commit message.

PR #327, opened 2017-08-17
Fix #326: Missing include file now fatal

### LINKED ISSUE #326: If file specified by "include file" doesn't exist, a fatal error should occur
Currently if a file specified in an `include file` input block directive does not exist a [warning is printed](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/egs_input.cpp#L969-L973) and execution continues.  Shouldn't this `egsWarning` be an `egsFatal`?  I can't think of any scenario off the top of my head where you would want to continue running if a file you meant to include was missing.  Conversely, we had an issue where an 'include file' path didn't exist but the simulation still ran to completion (albeit with a warning printed) producing potentially incorrect results.  A fatal error would have made it immediately clear something was wrong.

rtownson:
The `include file` input tag in `egs++` apps now calls `egsFatal` if the file does not exist, instead of `egsWarning`. This avoids the warning going unnoticed.

### DIFF
--- a/HEN_HOUSE/egs++/egs_input.cpp
+++ b/HEN_HOUSE/egs++/egs_input.cpp
@@ -962,14 +962,17 @@ int EGS_InputPrivate::addContent(istream &in) {
             what.assign(input,p,p2-p);
             string value;
             value.assign(input,p2+1,p1-p2-1);
-            for (int j=0; j<value.size(); j++)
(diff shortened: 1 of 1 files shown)


randlet:
Thanks Reid!

ftessier:
Juste lowercase "m" in commit title...

PR #331, opened 2017-08-21
Fix #330: allow input extension in egs++

### LINKED ISSUE #330: Cannot open .egsdat file for writing if using .egsinp on command line argument
This is a side effect from commit f93aa70effd3abae98ad6133e91445d5322f23fa.

If one leaves the `.egsinp` extension on the command line argument, then `outputData()` tries to open a `.egsdat` file with the double extension `.egsinp.egsdat`.

From `tutor7pp` with the example `test1.egsinp` file:

```
>tutor7pp -i test1.egsinp
(...)
EGS_Application::outputData: failed to open /Users/Marc/clrp/egs_home/tutor7pp/egsrun_10038_test1.egsinp_Moridin.local/test1.egsinp.egsdat for writing
```

rtownson:
Fix a crash that occurs when the egsinp extension is included when launching an egs++ application. Now the extension is stripped if it is included.

### DIFF
--- a/HEN_HOUSE/egs++/egs_application.cpp
+++ b/HEN_HOUSE/egs++/egs_application.cpp
@@ -269,6 +269,13 @@ EGS_Application::EGS_Application(int argc, char **argv) : input(0), geometry(0),
     sprintf(buf,"egsrun_%d_",egsGetPid());
     run_dir = buf;
     if (input_file.size() > 0) {
+
+        // Remove the .egsinp extension if it was included
+        size_t ext = input_file.rfind(".egsinp");
+        if (ext != std::string::npos) {
+            input_file = input_file.substr(0,ext);
+        }
+
         run_dir += input_file;
         run_dir += '_';
     }


HEN_HOUSE/egs++/egs_application.cpp
```diff
@@ -269,6 +269,13 @@ EGS_Application::EGS_Application(int argc, char **argv) : input(0), geometry(0),
     sprintf(buf,"egsrun_%d_",egsGetPid());
     run_dir = buf;
     if (input_file.size() > 0) {
+
+        // Remove the .egsinp extension if it was included
+        size_t ext = input_file.find_last_of(".egsinp");
```
ftessier: The search function should be `input_file.rfind(".egsinp")`, and then below `input_file = input_file.substr(0,ext)`; `find_last_of` searches the string for the last occurrence of any one of the characters in the character set `.egsinp`. The following code returns `myin` because the last character `e` is matched:
```
int main()
{    
    std::string str = "myinputfile";
    std::size_t ext = str.find_last_of(".egsinp");
    str = str.substr(0, ext-6);
    std::cout << str << "\n";
}
```

rtownson: Good eye! Fixed in the amended commit.

ftessier:
Just format with astyle.

PR #332, opened 2017-08-23
Add region labels to cavity and docs

rtownson:
Added region labels to `cavity`, similar to what was done for `egs_chamber` previously. Now you can use region labels anywhere you would otherwise type a list of global region numbers!

Also updated the documentation for egs_chamber and cavity to include some description of how to use region labels.


ftessier:
Simply added contributor name.

PR #336, opened 2017-08-31
Fix radionuclide atomic relaxations input

rtownson:
Fix a bug in `EGS_RadionuclideSource` where the relaxation input `atomic relaxations` required a `yes` or `no` input instead of what was described in the documentation. Now the accepted input is `ensdf`, `eadl`, or `off` with the default as `eadl`.

Also reduced the verbosity of radionuclide spectra.


mainegra:
@rtownson shouldn't there be an option to choose no relaxation?

rtownson:
Yes there should! I will add this..

PR #337, opened 2017-09-01
Adjust examin application to recent changes

mainegra:
- Print out new photoelectric cross sections by Sabbatucci and Salvat
- Turn ON relaxations if binding effects in Compton considered
- Print *.xsections file only if photons requested
- Reset some defaults

`egsnrc.mortran`: Only print photon binding energies if *.xsections file
output and eadl_relax is true.


ftessier:
Removed end-of-line white spaces.

PR #339, opened 2017-09-02
Fix rest mass and energy in DOSXYZ source 20

blakewalters:
If a BEAMnrc shared library geometry was interposed
between the source and the phantom and the source
in the BEAMnrc .egsinp was not a phase space or BEAM
simulation source, then the incorrect particle energy
was used to pass the particle through the shared lib.
geometry and, subsequently, on entry to the DOSXYZnrc
phantom.  Quite a dangerous bug.  Recommend patching this
as quickly as possible.


blakewalters:
Just to be more specific: If you had specified source 21, 23, or 24 in the BEAMnrc .egsinp, there was no issue.  Otherwise: 1) if the source specified in this .egsinp file was monoenergetic, then the rest mass was erroneously added to all charged particles before sending them through the shared library geometry or 2) if the source specified in this .egsinp file had an energy spectrum, the energy of all particles was erroneously reassigned using the energy spectrum before passing through the shared library geometry.

ftessier:
Updated commit title and text slightly. Cleaned up the merge bubble on this branch.

PR #341, opened 2017-09-05
OpenMP support in EGSnrc

edoerner:
## Parallel implementation of EGSnrc using OpenMP

This is a prototype for the parallelization of EGSnrc user codes using the OpenMP API.
It is currently implemented on the DOSXYZnrc user code. An advantage of this development
is that removes the necessity of the use of a job scheduler and/or lock-file mechanism to
handle parallel jobs and combination of results.

## Installation and usage

    1a. If you start from a fresh installation just install EGSnrc as usual. All the needed files 
    for OpenMP support are included in this fork.
    1b. Otherwise, once all the files are downloaded to HEN_HOUSE area, copy dosxyznrc folder 
    ($HEN_HOUSE/user_codes/dosxyznrc) to the user codes area (i.e. $EGS_HOME)
    2. In terminal, go to the dosxyznrc folder inside $EGS_HOME.
    3. If the provided standard_makefile is used, just type "make omp" to compile the user code
    with OpenMP support. It essentially compiles with -fopenmp flag (for gcc only). If you want to
    use another compiler, you must change this flag with the proper one inside standard_makefile.
    4. Execute the code as usual. The message "Number of threads in OpenMP parallel execution ..."
    should appear if OpenMP support was enabled

Due to the use of private variables (through the OpenMP THREADPRIVATE directive) it is needed
for some machines to increase the stack size of the OS. For example, on macOs or
GNU/Linux simply type on terminal "ulimit -s hard". In our case, we added that command on
the .bashrc configuration file.

We have tested sources 0 (Parallel Rectangular Beam Incident from Front) and 2 (Phase-Space
Source Incident from Any Direction), although with the notable exception of a BEAMnrc source
model the other sources should work without problems.

Also should be noted that the combination of results at the end of the simulation is
implemented only for the *.3ddose file. Information printed on console and output to
*.egslst file come only from the MASTER thread.

## Additional information

1. All relevant changes to the source files and makefiles are labeled with the "#OMP_EGS" tag.
2. In dosxyznrc $NBATCH was selected as 1, in order to start the parallel region inside 
the ibatch loop just one time. You can change it to the default value of 10 if you wish, although you should expect and additional overhead.
3. If dosxyznrc is compiled without OpenMP support (i.e. typing "make" in dosxyznrc user code folder) 
the original implementation is recovered. Therefore, it is possible to compare OpenMP execution 
with the parallelization using a BQS, for example.

### DIFF
--- a/HEN_HOUSE/makefiles/standard_makefile
+++ b/HEN_HOUSE/makefiles/standard_makefile
@@ -27,6 +27,9 @@
 #                   Jeffrey Siebers
 #                   Ernesto Mainegra-Hing
 #
+#  OpenMP support:  Edgardo Doerner
+#                   Paola Caprile
+#
 ###############################################################################
 #
 #  The main portion of a Makefile for the EGSnrc system. This file is included
@@ -162,6 +165,10 @@ debug:
 	@$(MAKE) -s WHAT=_debug OPTLEVEL_F="$(FDEBUG)" OPTLEVEL_C="$(CDEBUG)"
 #@($(MAKE) -s WHAT=_debug OPTLEVEL_F="$(FDEBUG)" OPTLEVEL_C="$(CDEBUG)" && echo $(ok_message)) || echo $(failed_message)
 
+# OMP_EGS - Build an user code with OpenMP support
+omp:
+	@$(MAKE) -s OPTLEVEL_F="$(FOPT) -fopenmp" 
+
 # Clean-up the user code directory.
 clean:
 	@echo "$(REMOVE) mortjob.mortran $(FORTRAN_FILE).$(FEXT) $(FORTRAN_FILE).mortlst"
@@ -180,4 +187,6 @@ execlean:
 # execlean have no list of prerequisits so that make can determine whether
 # they need to be remade, we declare them as phony.
 #
-.PHONY: opt noopt debug clean execlean
+# OMP_EGS - Added omp for compilation with OpenMP support
+#
+.PHONY: opt noopt debug omp clean execlean
--- a/HEN_HOUSE/omega/progs/beamdp/Makefile
+++ b/HEN_HOUSE/omega/progs/beamdp/Makefile
@@ -26,13 +26,19 @@
 #  Contributors:    Blake Walters
 #                   Frederic Tessier
 #
+#  OpenMP support:  Edgardo Doerner
+#                   Paola Caprile
+#
 ###############################################################################
 
 
 include $(EGS_CONFIG)
 include $(SPEC_DIR)beamnrc.spec
 
-SOURCES = $(MACHINE_MACROS) $(EGS_UTILS)phsp_macros.mortran $(IAEA_PHSP_MACROS)\
+# OMP_EGS: phsp_macros.mortran contains OpenMP constructs, defined at 
+# egsnrc.macros. Therefore it is added to SOURCES.
+SOURCES = $(MACHINE_MACROS) $(EGS_SOURCEDIR)egsnrc.macros \
+	$(EGS_UTILS)phsp_macros.mortran $(IAEA_PHSP_MACROS) \
           beamdp.mortran \
           $(EGS_UTILS)xvgrplot.mortran $(EGS_SOURCEDIR)lnblnk1.mortran
 
--- a/HEN_HOUSE/omega/progs/readphsp/Makefile
+++ b/HEN_HOUSE/omega/progs/readphsp/Makefile
@@ -26,13 +26,19 @@
 #  Contributors:    Frederic Tessier
 #                   Blake Walters
 #
+#  OpenMP support:  Edgardo Doerner
+#                   Paola Caprile
+#
 ###############################################################################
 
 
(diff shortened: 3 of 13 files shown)


ftessier:
Thank you for cleaning up this pull request! We will review and merge this in as soon as possible. In the mean time, can you add a version of the configure script that will compile with OpenMP support. This way we can test OpenMP configuration more easily and add it to the automated TravisCI checks (see `/.travis.yaml`). Start from `/HEN_HOUSE/scripts/configure.expect` and _send_ the required flags in the Fortran, C and C++ compiler sections.

edoerner:
Hi @ftessier , first of all sorry for the delay (University stuff, holidays, family, etc...). I have looked at the files that you pointed out and I have some doubts (I am not an expert, actually this was my first pull request ever, hehehe). So, the idea is that I modify the "configure" file in order to add OpenMP support, right?. 

For example, it seems that I would need to modify line 330 of the "configure" file (f_fflags=-fPIC) and add the required flag (for gfortran would be -fPIC -fopenmp). Thanks for your help!

ftessier:
@edoerner, that would be one way to do it... but a simpler alternative at this point would be to copy the `configure.expect` script which runs the configuration script unattended, i.e., sending the configure script the answers as if you typed them in at the keyboard. You can create a copy of this script which provides the right compiler flags etc. for your OpenMP version. You will need the `expect` package, which is not installed by default on all linux distributions.

ftessier:
Given that there is just an additional `-fopenmp` flag, I can do it of course, no worries! I thought perhaps the changes to compiler flags were more involved!

edoerner:
Ok, thank you!

crcrewso:
I'm reading your commits and it's leaving me a touch confused. Does this implementation provide both MPI and openMP handling? Or just openMP? If it's both could you please provide a bit about the expected distribution behaviour of the MPI control thread and the thread interaction model chosen. Ie how individual histories are dispatched and results returned.

edoerner:
Hi @crcrewso , I remember our last chat about this development. Actually, this pull request was originally intended for OpenMP support only. I started to work on MPI but during the process I merge the branch where I am working on MPI/OpenMP support with the 'develop' branch of my fork of EGSnrc, without knowing that the commits will be transmitted to this pull request. I need to clean and further check the MPI implementation part. Also I think that I over-abused of the use of THREADPRIVATE to privatize the COMMON blocks, so I need to optimize that, hehehe.

Maybe it will be just better to roll back to the original state of this pull request (@ftessier)?. I am sorry for the mess!.

About MPI support, this development is able to use OpenMP only mode, MPI only mode or hybrid MPI/OpenMP execution, depending on your configuration. I attached three configuration files as examples to enable each mode. For the hybrid approach the execution flow is the following: The histories are evenly divided among MPI processes. Inside each MPI process the histories are divided among OpenMP threads and simulated. Each thread creates a *.pardose file at the end of the simulation and then the "master" process combines them using the available functionality at DOSXYZnrc. With the different tests with OpenMP we found that this solution was more simple than trying to merge the results using OpenMP and/or MPI functionality like reduction functions. I have tested with gcc and Intel implementations, but I have not made a comparison between them.

We have an accepted Technical Note on Medical Physics for the OpenMP part, so as soon as they give me the link to the publication I can send it.

[osx_config_files.zip](https://github.com/nrc-cnrc/EGSnrc/files/1436116/osx_config_files.zip)

ftessier:
Indeed when you push to a pull request branch, the pull request is updated! This actually proves extremely useful to refactor the commits on a pull request until it is "just right" for merging.

Best thing to do at this point is to commit any pending change, and associate a new branch name to your latest commit (with all the newer MPI stuff), e.g.:
```
git checkout -b mpi-branch
```
Then go back on your develop branch and reset it to the latest openmp commit f75a934e7e6507a52ae12d38e6948cc7c374b0de (September 5), then push it up to roll back the pull request to its original state:
```
git checkout develop
git reset --hard f75a93
git push --force-with-lease
```

ftessier:
Note that your `develop` branch is for all practical purposes "locked" until we merge this in (which might take a while!), because as you found out anything pushed on your develop branch is added to this pull request! If you care about this, you can opt instead to close this pull request and reopen another one with a different branch name in order to free up your develop, e.g.:
```
git checkout -b feature-openmp f75a93
git push
```
and replace this pull request for your new branch `feature-openmp` instead of your `develop`

edoerner:
Ok, thanks for the help!, now I think that the pull request is clean... any question about the use of OpenMP on EGSnrc do not hesitate to ask me :).

ftessier:
Looks good, thanks! We will review and test as soon as we can. Comments from anyone willing to test this implementation will prove useful.

edoerner:
hi @crcrewso, I am sorry for the huge delay in my answer, but it has been a really difficult semester here. As you say, shared library simulation (i.e. BEAMnrc?) are not supported. I hope that I will have more time in the near future to tackle this...

edoerner:
Dear all, I will now close this pull request because I want to update my develop branch and also add some updates and support of shared library simulations.

PR #352, opened 2017-10-03
Fix #351: Infinite loop using beam source in egs++

### LINKED ISSUE #351: Infinite looping when using beam source in egs_chamber
I found that while running a BEAM shared library source in egs_chamber, I was getting an infinite loop (simulation would never finish, even with only 1 history).

After some sleuthing, my colleage Marc-Andre Renaud and I found that in egs_beam_source.cpp, there is a sanity check where the particle energy must be less than Emax:
       `  if (te > Emax) {
          ok = false;
      } //egsInformation("Emax rejection\n"); }`

We determined that if te equals Emax,  it can still fail the inequality te > Emax because of some funny business in comparing floating point numbers. This would occur for any particles in your beam model which have Emax and do not interact (i.e. their energy remains Emax).

Our workaround for this was to change the if statement to if (te > 1.01*Emax), but the more elegant solution would be to implement a robust, float rounding error proof comparison such as the epsilon constant.

rtownson:
Due to a floating point comparison, an infinite loop was possible in egs++ applications using a beam shared library source. This was fixed using the epsilon constant.

### DIFF
--- a/HEN_HOUSE/egs++/sources/egs_beam_source/egs_beam_source.cpp
+++ b/HEN_HOUSE/egs++/sources/egs_beam_source/egs_beam_source.cpp
@@ -206,7 +206,7 @@ EGS_I64 EGS_BeamSource::getNextParticle(EGS_RandomGenerator *, int &q,
             te -= EGS_Application::activeApplication()->getRM();
         }
         ok = true;
-        if (te > Emax) {
+        if (te > Emax + epsilon) {
             ok = false;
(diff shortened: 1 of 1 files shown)


ftessier:
Just tweaked the commit message.

vahx129:
Hi Reid, 

I believe I manually tried to implement that epsilon fix in EGS_2017 but didn't have any success. I haven't looked into this issue much since I haven't been actively working on the TrueBeam model but I can test an input file later on and update you if the issue persists.

PR #353, opened 2017-10-04
Fix #344: Add check for duplicate geometry names

### LINKED ISSUE #344: Warn when a geometry name is not unique
It would be nice if `egs++` warned users when geometry names are duplicated. If a user mistakenly duplicate a geometry name, results tend to be unexpected. This will eventually be much more of a problem when PR #103 is merged, since geometries can be defined in external files. I guess this may even warrant a fatal error, rather than a simple warning.

rtownson:
Add a check for duplicate geometry names, and do not allow them. Previously geometries with the same name could produce unexpected results.

**Please check this with some complex input files.** As far as I know geometries are never given the same name intentionally but I could be mistaken.

### DIFF
--- a/HEN_HOUSE/egs++/egs_base_geometry.cpp
+++ b/HEN_HOUSE/egs++/egs_base_geometry.cpp
@@ -164,10 +164,11 @@ public:
         if (!g) {
             return -1;
         }
-        for (int j=0; j<nnow; j++)
+        for (int j=0; j<nnow; j++) {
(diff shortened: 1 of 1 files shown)


mchamberland:
So far so good in initial testing with a few egs_brachy input files.

ftessier:
I agree no two geometries should have identical names. I amended the commit to format `egs_base_geometry.cpp` with astyle.

PR #360, opened 2017-10-11
Fix #334: recycle definition in egs++ phsp sources

### LINKED ISSUE #334: Inconsistent terminology in iaea_phsp_source
Maybe I'm misunderstanding, but it appears the IAEA phase-space source uses a terminology inconsistent with the rest of EGSnrc. It uses `Nrecycle` (in the code only, but that's still confusing to advanced users that look at it!) to mean the number of restarts and `Nreuse` (in the code) and `reuse` (as an input key) to mean the number of times a particle is recycled. See [here](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/sources/iaea_phsp_source/iaea_phsp_source.cpp#L293-L306).

I would suggest renaming to `Nrestart` and `Nrecycle` , respectively. The `reuse photons` and `reuse electrons` input keys could be renamed to `recycle photons` and `recycle electrons`. The old input keys should probably be kept for backwards compatibility.

rtownson:
Fixes the use of the term `reuse` in the `egs++` phsp sources, instead replacing it with `recycle`. Also fixes the implementation of recycling, so that 1 recycling means that the particle is used twice. 

**Previously, `reuse=1` actually meant that the particle was used once.**

### DIFF
--- a/HEN_HOUSE/egs++/sources/egs_phsp_source/egs_phsp_source.cpp
+++ b/HEN_HOUSE/egs++/sources/egs_phsp_source/egs_phsp_source.cpp
@@ -64,15 +64,15 @@ void EGS_PhspSource::init() {
     description = "Invalid phase space source";
     Nread = 0;
     count = 0;
-    Nrecycle = 0;
+    Nrestart = 0;
     Npos = 0;
     Nlast = 0;
     wmin = -veryFar;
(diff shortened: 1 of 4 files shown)


ftessier:
Just formatted through astyle.

PR #363, opened 2017-10-15
Add radionuclide alpha scoring options

rtownson:
Add the option to discard radionuclide alpha particles (the default) or include them as a local energy deposition. The is done using the `alpha scoring` input option for `EGS_RadionuclideSpectrum`. Also allow the user to use the input parameter `nuclide` or `isotope` interchangably.


ftessier:
Just ran through astyle. I see that astyle messes up the alignment of multi-line template variable declarations for some reason. We'll have to live with this or figure out how to tell astyle how to format this properly.

PR #365, opened 2017-10-20
Fix type mismatch in variable icase when compiling BEAM shared librar…

blakewalters:
Fix type mismatch in variable icase when compiling BEAM shared libraries.

Variable icase in all BEAM library routines calling beam_shower(icase) is
now $LONG_INT.  Gets rid of warnings.


ftessier:
Just formatted the commit message.

PR #366, opened 2017-10-21
Fix #349 Fix pegsless mode in DOSXYZnrc source 20

### LINKED ISSUE #349: DOSXYZnrc Source 20 simulations in pegsless mode do not work
I've started using pegsless mode. Simulations with traditional PEGS files work OK, i.e. I get:
```
About to initialise BEAM code..
 In beamlib_init!
 i_par =            0
hen_house = /home/EGS2017d/HEN_HOUSE/
egs_home  = /home/BEAM2017/
beam_code = MY_CODE
inputfile = MY_INPUT_FILE
pegs file = /home/EGS2017d/HEN_HOUSE/pegs4/data/521icru.pegs4dat
```

But in pegsless mode I get:
```
About to initialise BEAM code..
 In beamlib_init!
 i_par =            0
hen_house = /home/EGS2017d/HEN_HOUSE/
egs_home  = /home/BEAM2017/
beam_code = MY_CODE
inputfile = MY_INPUT_FILE
pegs file =

***************** Error:
 could not find pegs4 file named

```

The BEAM code input file that is used as shared library has the required lines (as does the DOSXYZ input file), i.e.,
```
:start media definition:
    ae = 0.521
    ue = 50.511
    ap = 0.01
    up = 50

material data file=/home/EGS2017d/HEN_HOUSE/pegs4/data/material.dat

:stop media definition:
```

blakewalters:
If you are using a BEAM shared library geometry with
source 20, it is assumed that the pegs data for the BEAM
geometry is the same as that for DOSXYZnrc.  Previously,
if you ran DOSXYZnrc in pegsless mode, this information was
not correctly passed on to the BEAM simulation.  Now it
is, and the BEAM simulation also runs in pegsless mode.
(Don't forget to define the media in the BEAM input file!)

### DIFF
--- a/HEN_HOUSE/user_codes/dosxyznrc/srcxyznrc.mortran
+++ b/HEN_HOUSE/user_codes/dosxyznrc/srcxyznrc.mortran
@@ -2063,22 +2063,6 @@ ELSEIF( (enflag = 2) | (enflag = 3) )["phase space input or full BEAM sim."
     "strip trailing blanks"
     the_shared_lib=the_shared_lib(:lnblnk1(the_shared_lib));
 
-    IF(the_shared_lib='0')[
-	"no shared library
-	SHLflag = 0;
-    ]
-    ELSEIF(the_shared_lib='particleDmlc') [
-	"particles must go through MLC code compiled as a shared library
-	MLCflag =1;
-	SHLflag = 1;
-    ]
-    ELSE[
-	"particles must go through BEAM shared library
-	SHLflag =1;
-	MLCflag=0;
-	OUTPUT61 the_shared_lib;(/' BEAM library: ',/A80);
-    ]
-
     the_phsp_file=FILNAM(:INDEX(FILNAM,',')-1);
     FILNAM=FILNAM(INDEX(FILNAM,',')+1:);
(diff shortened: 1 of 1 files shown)


mainegra:
@blakewalters does one have to set the input key 'pegs file' for a BEAM source to pegsless? Or is it enough to just use pegsless on the command line? In the latter case, what should one input in the BEAM source definition then?

blakewalters:
In general, @mainegra, if you want to run a BEAM simulation source in pegsless mode, then you have to provide the input "pegsless" in place of the name of a pegs data file in the source inputs, regardless of whether or not the code driving the source is being run in pegsless mode or not.  For DOSXYZnrc source 20, however, which is a phase space source that can be run through a BEAM shared library geometry before it reaches the phantom, it is assumed that the BEAM geometry uses the same pegs data as the DOSXYZnrc calculation, and so no input for the pegs data file (or "pegsless") to be applied to the BEAM geometry exists.  The bug was in not passing the pegsless status of the DOSXYZnrc simulation on to the BEAM geometry properly.

ftessier:
Adjusted the commit message and format.

PR #368, opened 2017-10-30
Fix variable declaration in subroutine replace_env

blakewalters:
Declared common block EGS-IO at top of subroutine replace_env--
the subroutine which allows $ and ~ to be used
at the beginning of file path names--so that i_log is declared for
output.  Don't know why this compiled with no complaints, but the
symptom was actually an incorrect "Job failed" output from
EXphantom runs, first noticed at the EGS/BEAM Workshop in Trieste, 2017.

### DIFF
--- a/HEN_HOUSE/src/egs_utilities.mortran
+++ b/HEN_HOUSE/src/egs_utilities.mortran
@@ -1408,6 +1408,9 @@ subroutine replace_env(fname);
 "full names or ~ with the full name of $HOME"
 "Assumes environment variable or ~ appears only at the beginning of the"
 "file name"
+
+;COMIN/EGS-IO/;
+
 character*(*) fname;
 character*256 dirname;
 integer indsep,ind1,ind2;


ojalaj:
I applied this change by copying the updated HEN_HOUSE/src/egs_utilities.mortran and remaking under BEAM and dosxyznrc directories in EGS_HOME, but still got 'Failed' status from slurm, when jobs finished (dosxyz source 20 run, through BEAM shared library). However, when looking at output files (.egslst, .egslog etc.) everything has run fine. This behaviour has been there already for a while, but I thought that this fix would have solved this 'Failed' issue with jobs that are finishing. There is always some fort.17 or fort.3 file around, e.g. this time:
***************** Error:
 job           41 : failed to rewind or lock the control file          -1

Should I remake somewhere else, or is this a separate issue, i.e., something is not finishing correctly in batch runs?

blakewalters:
Hi @ojalaj:

This is likely not related to the changes to replace_env.  Looks like a problem some of your parallel runs not finishing properly.

ojalaj:
OK. Anyhow, it seems that all of them finish 'enough', since .egslog and .egslst for the total run output fine, without errors.

ftessier:
Tweaked the commit message.

PR #369, opened 2017-11-02
Fix authorship and readme

ftessier:
Add Iwan Kawrakow as an author of the egs_chamber source code (instead of a contributor).

Fix the top-level readme to state the original release date of EGSnrc, and qualify it as a complete overhaul of EGS (instead of a extended and improved version). Remove the copyright transfer requirement note from the contributing section, because we will be moving to allow external copyrights to be included in the project.


ftessier:
@mainegra can you double-check again for the additional egs_cbct changes? Thanks!

mainegra:
@ftessier looks good! Thanks for updating!

PR #370, opened 2017-11-02
Fix isotropic emission into solid angle and egs_view alias

mainegra:
This PR affects only the isotropic source (and hence the Fano source) when used to emit particles into a solid angle defined by a minimum and maximum polar angle theta.

The expression to sample **cos(theta)** between **cos(theta_min)** and **cos(theta_max)** had the terms inverted. While `1-2*rand` produces the same distribution of **cos(theta)** as `2*rand-1` when sampling between 0 and 180 degrees or any angular interval centered around 90 degrees, this is not the case in general when restricting the emission to an arbitrary angular interval between **theta_min** and **theta_max**. One should use the correct expression:

`cos(theta) =  cos(theta_min) - rand * (cos(theta_min - cos(theta_max)))`

rather than

`cos(theta) = -cos(theta_min) + rand * (cos(theta_min - cos(theta_max)))`

In the second expression, if the angular interval lies entirely in the first quadrant (0 to 90) or in the second quadrant (90 to 180) the angles sampled will always be in the opposite quadrant which is not the desired outcome.

**egs_isotropic_source:**

- Fix emission into solid angle
- Remove incomplete comment at the bottom of the header file
  referring to the addition of a solid angle restriction
  by a user back in January, 2008

**egs_fano_source:**

- Fix emission into solid angle
- Add comments referring to the rejection technique
- Change authorship to add Hugo as author rather than contributor
  since I started from his initial version
- TODO: Derive it from egs_isotropic_source rather than being an
        almost verbatim copy of it.

**egsnrc_bashrc_additions:**

- Remove parentheses from the egs_view alias as it won't take any input
  parameters

### DIFF
--- a/HEN_HOUSE/scripts/egsnrc_bashrc_additions
+++ b/HEN_HOUSE/scripts/egsnrc_bashrc_additions
@@ -89,7 +89,7 @@ alias pprocess='$HEN_HOUSE/scripts/pprocess'
 # hard-coded run-time search path (no -rpath)
 #
 if test -d $HEN_HOUSE/egs++/dso/linux-static; then
-   alias egs_view='(LD_LIBRARY_PATH=${HEN_HOUSE}egs++/dso/linux-static/ ${HEN_HOUSE}pieces/linux/egs_view_64)'
+   alias egs_view='LD_LIBRARY_PATH=${HEN_HOUSE}egs++/dso/linux-static/ ${HEN_HOUSE}pieces/linux/egs_view_64'
 fi
 # OS X specific settings
 canonical_system=$( cat $EGS_CONFIG | grep "canonical_system =" | sed 's/canonical_system = //' )
--- a/HEN_HOUSE/egs++/sources/egs_fano_source/egs_fano_source.cpp
+++ b/HEN_HOUSE/egs++/sources/egs_fano_source/egs_fano_source.cpp
@@ -21,9 +21,10 @@
 #
 ###############################################################################
 #
-#  Author:          Ernesto Mainegra-Hing, 2016
+#  Authors:         Ernesto Mainegra-Hing, 2016
+#                   Hugo Bouchard, 2016
 #
-#  Contributors:    Hugo Bouchard
+#  Contributors:
 #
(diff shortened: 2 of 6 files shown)


rtownson:
Good find! Please make the same change in `egs_radionuclide_source.cpp`.

HEN_HOUSE/scripts/egsnrc_bashrc_additions
```diff
@@ -89,7 +89,7 @@ alias pprocess='$HEN_HOUSE/scripts/pprocess'
 # hard-coded run-time search path (no -rpath)
 #
 if test -d $HEN_HOUSE/egs++/dso/linux-static; then
-   alias egs_view='(LD_LIBRARY_PATH=${HEN_HOUSE}egs++/dso/linux-static/ ${HEN_HOUSE}pieces/linux/egs_view_64)'
+   alias egs_view='LD_LIBRARY_PATH=${HEN_HOUSE}egs++/dso/linux-static/ ${HEN_HOUSE}pieces/linux/egs_view_64'
```
ftessier: **Wow!** I did not know about this single-exectuable level of scoping for shell variables in bash! Leaving out the semi-colon after the variable definition implies the variable is strictly only available for the executable immediately following. Oh well, you learn some (bash) thing everyday. So there are **three**, not two, levels of scoping on the command-line; noting them here for reference:
```
key=value a.out    # key only defined for executable a.out (try: key=value env | grep key; env | grep key)
key=value          # key only defined in current shell
export key         # key defined in current shell and its sub-processes (through the environment)
```

ftessier:
Broke this down in two commits since the bashrc one is unrelated to the isotropic source. Also added Long Zhang to the list of contributors to the isotropic source, for his addition for the restricted polar angule sampling added in 2008, and corrected in this pull request.

mainegra:
@rtownson I added the fix for the radionuclide source as well! Time to review! :-)

ftessier:
Just squashed the two angular sampling commits. Will "Rebase and merge" into develop now.

PR #374, opened 2017-11-07
Fix #373: e_mutr is nan in rare case using g

### LINKED ISSUE #373: Application "g" does not output mass energy absorption coefficients for large numbers of histories
When I increase the number of particles from 10M to 500M, the mass energy absorption coefficient disappears from the G output. See attached comparison of the outputs.
screenshot at 2017-09-28 19 36 14

rtownson:
Fix a bug where e_mutr and e_muen were nan for a large number of histories and specific seeds. This was due to a rare case where gmfp evaluated to zero. A check was added to catch this case.


rtownson:
I changed this fix to check for `gmfp = 0` instead of `gmfp <= 0` as the fail case, since the gmfp seems to be legitimately negative in some cases (still trying to understand this). 

I also now use a `goto` to resample from the source without iterating `icase`. This was because fortran won't let me decrement the iterator within the loop.

PR #377, opened 2017-11-09
Fix #375: Fix multiple SYNC or DYN CMs of same type

### LINKED ISSUE #375: Cannot add two of the same SYNC/DYN CM to a BEAM accelerator
If you add two of any SYNC CM to the same accelerator, it does not compile. I believe this is due to the way the `USER-SYNCMLCE` type common blocks are included in `COMIN/USER/` and do not depend on the CM name. I believe this also applies to the DYN CMs.

The suggested workaround is to simply duplicate the CM code and give it a new name.

blakewalters:
The user can now compile acclerators containing multiple
instances of any of the SYNC or DYN CMs.  Runs have been
tested successfully for multiple SYNCJAWS and SYNCMLCE.

Also, fixed an annoying output error in DYNVMLC_cm.mortran.


ojalaj:
Hello @blakewalters, I noticed you have corrected some annoying error in DYNVMLC (and also in corresponding SYNC CMs?). May I ask what kind of error should it solve?

blakewalters:
Hi @ojalaj:  This was an output format statement causing a compilation error for any accelerator with DYNVMLC.  The remainder of these changes just relate to the ability to compile an accelerator having more than one of each type of these CMs.

ftessier:
Just cleaned up to remove the extra merges.

ftessier:
Simply tweaked the commit format and message.

HEN_HOUSE/omega/beamnrc/beamnrc_user_macros.mortran
```diff
@@ -86,8 +86,8 @@ REPLACE {$COVMIN} WITH {10}        "min. no. of particles crossing a scoring"
                                    "average energy/angle only)"
 REPLACE {$MXMED} WITH {12}         "max # of media
 REPLACE {$MXREG} WITH {3000}        "maximum number of regions (absolute)
-REPLACE {$MXSTACK} WITH {10000}    "maximum stack (much less if no brem split)
-REPLACE {$MAXBRSPLIT} WITH {2000}  "maximum number of bremsstrahlung splits
+REPLACE {$MXSTACK} WITH {1000000}    "maximum stack (much less if no brem split)
+REPLACE {$MAXBRSPLIT} WITH {500000}  "maximum number of bremsstrahlung splits
```
ftessier: @blakewalters can you confirm that committing these increases in default sizes is intentional?

blakewalters: Absolutely unintentional @ftessier; these are the settings I routinely use.  Thanks for catching this!  I'll return them to their original defaults.

ftessier: All right, just push the change as a new commit on top, as discussed!

ftessier:
Ironed out the merge bubble.

PR #379, opened 2017-11-14
Fix outdated/meaningless version numbers

blakewalters:
References in individual codes to version/revision numbers dating back to when revision was under CVS or, even earlier, RCS control have either been deleted or replaced by blank strings.

It is anticipated that this will be superseded by, perhaps, a single version/revision scheme for each official EGSnrc release on GitHub.


ftessier:
@blakewalters this commit failed on pegs4 compilation; can you look into it?

ftessier:
Updated commit message and removed EOL white space.

PR #389, opened 2017-12-01
Fix nleaf input in SYNCMLCE GUI

blakewalters:
Input did not exist!


ftessier:
Updated commit message.

PR #392, opened 2017-12-09
Fix radiative splitting csplit initialization

ftessier:
Initialize the `csplit` variable to 1 by default in the egs_chamber
application constructor; `csplit` is used to read in the radiative
splitting number from the input file.

Previously, `csplit` was only initialized if the input contained a
variance reduction input block. This caused a serious bug to be
introduced in commit 50e0d33, whereby `the_egsvr->nbr_split` was set to
the arbitrary (uninitialized) `csplit` number in the shower loop when
there was no variance reduction input block. If this number happened to
be a large positive integer (about half the time), any radiative event
immediately exceeded the stack size and aborted the simulation.

This probably went unnoticed for so long because one rarely runs
egs_chamber without variance reduction.


rtownson:
Good job hunting this down!

ojalaj:
To be sure, I'd like to confirm that has it been enough that there is a variance reduction block, but cs enhancement = 0, TmpPhsp = 0 and no RR, i.e., VRTs are turned off?

rtownson:
Correct @ojalaj, previously you would not have a problem if there was a variance reduction block regardless of whether anything was turned on.

ojalaj:
Thanks @rtownson for the confirmation!

PR #396, opened 2017-12-12
Fix #354: Update pirs-701 docs with new iarg

### LINKED ISSUE #354: PIRS701 table 6 needs updating re new IARG calls

rtownson:
Expand the pirs-701 table 6 to include recently added iarg calls for EII, sub-threshold relaxation photons and electrons.


ftessier:
Tweaked the commit message.

PR #397, opened 2017-12-12
Fix #384: Remove source types from g

### LINKED ISSUE #384: The source type options in 'g' should be removed
In the `g` code there are options for `source type=1` to set a beam radius and distance. Since the geometry is an infinite uniform medium, the source direction and location is meaningless. These input options should therefore be removed.

rtownson:
Remove source type options from g, because they are meaningless in an infinite space on a single medium. The options were likely copied over from an old code originally, and contained directional control of the source.


ftessier:
Tweaked the commit message.

PR #402, opened 2018-01-05
cavity e-/e+ fluence calculation redux

mainegra:
The charged particle fluence calculation in cavity was recently modified to provide the actual fluence rather than the product fluence times stopping power. To this end, the quantity scored at each energy bin was divided by the stopping power `Li` at the lower energy edge of the corresponding bin `i`. This is correct for the initial energy bin and the bins fully covered during the step. However at the final bin, the particle's energy at the end of the step doesn't always coincide with the lower energy edge of that bin and using that initial approach introduces an artifact. One could use the stopping power at the mid-energy of the bin, but it also introduces an artifact. Both cases can produce significant differences in the differential flluence and non-negligible differences in the total fluence. 

This is shown for the differential fluence in an air cavity for a 1.25 photon beam incident on a water phantom:
eflu_2bs_ik_vs_ref
The proper approach is based on the expression for the step length obtained integrating the inverse of the stopping power from the final energy `Ee` to the intial energy `Eb`. Based on the linear interpolation in a logarithmic energy grid as used in EGSnrc (i.e. `dedx = a + b*log(E)` ) and a power series expansion of the `ExpIntegralEi` function that is the result of the integration of the inverse of the stopping power with respect to energy one can derive this expression to either 3rd or 5th order.  Since both produce the same results in the examples tested, the `O(eps^3)` is set as default for being both, accurate and only 8% slower than the simpler FLURZ-mode approach B.

Pre-calculate values of `1/Lmid[i]` for faster evaluation  when using the `O(eps^3)` mode. `O(eps^5)` would require more pre-computed values than just `1/Lmid`. One requires for instance `lnEmid[i]` to get the `b` parameter and `eps[i]=1-E[i]/E[i+1]`. Not impossible, but seems unnecessary considering the excellent agreement with `O(eps^3)`, which should be always used.

Made algorithm work for logarithmic energy scale where the energy bin  width is not constant which must be accounted for when determining the bin fractions and scoring the total fluence. Pre-calculate bin widths `DE[i]` for the log energy scale total fluence scoring. Scores total fluence accounting for correlations in the uncertainty while it also integrates the differential fluence at the end assuming bin contributions to be independent of each other.

### DIFF
--- a/HEN_HOUSE/egs++/egs_interpolator.h
+++ b/HEN_HOUSE/egs++/egs_interpolator.h
@@ -230,6 +230,15 @@ public:
     EGS_Float getXmax() const {
         return xmax;
     };
+    /*! \brief Get i-th interval interpolation parameter a[i]. */
+    inline EGS_Float get_a(int i) const {
+        return a[i];
+    };
+    /*! \brief Get i-th interval interpolation parameter b[i]. */
+    inline EGS_Float get_b(int i) const {
+        return b[i];
+    };
+
 
 private:
 
--- a/HEN_HOUSE/user_codes/cavity/cavity.cpp
+++ b/HEN_HOUSE/user_codes/cavity/cavity.cpp
@@ -224,7 +224,6 @@ extern EGS_Float steps_ustepi[MAX_STEP], steps_ustepf[MAX_STEP];
 extern int steps_n;
 #endif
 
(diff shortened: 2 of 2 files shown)


ftessier:
Adusted commit messages.

PR #404, opened 2018-01-12
Fix #403: Fix flurznrc normalization of individual || jobs

### LINKED ISSUE #403: normalization of partial parallel run results in FLURZnrc
While doing some parallel runs with FLURZnrc I note that the results in the file_wn.egslst file (and other outputs such as file_n.spectra) are incorrectly normalized, apparently to the total fluence rather than the partial fluence in that individual run. The combined results are fine.  I checked DOSRZnrc and it doesn't do this, i.e. the results in any of the file_wn.egslst files are normalized properly. This FLURZnrc issue is not really a problem as long as one is aware of it. But it could cause errors to be made when jobs fail etc.

blakewalters:
For non-phase space sources, results from individual jobs were
normalized by the same total fluence (AINFLU) as the combined jobs.
This has been corrected so that AINFLU is adjusted to reflect the
number of histories run by a given job.  Somewhat frighteningly,
the original coding had AINFLU adjusted correctly but only if the
job couldn't execute within the user input max. CPU time.


ftessier:
Formatted the commit message.

PR #407, opened 2018-01-17
Add support for fish shell

marenaud:
Just converted the egsnrc_bashrc_additions script to work with the [fish shell](https://github.com/fish-shell/fish-shell) because I was tired of switching back to bash whenever I wanted to use BEAMnrc :)

Also added the proper detection flag in finalize_egs_foruser.


ftessier:
@marenaud I polished up the commit to make it ready for merging and pushed it to your branch.

PR #414, opened 2018-02-28
Add git hash, branch and compile time to egs++ output

rtownson:
Adds a record of the git hash, branch and configuration/installation and compile times to the standard EGSnrc output. 

This was done by passing the parameters using preprocessor compiler defines in fortran. In order to do this, **it was necessary to add the -cpp flag to the core fortran compilation**. The configuration time is passed through the `machine.mortran` file, but you couldn't do this with the other parameters because they need to be updated at application compile time (not configuration/installation time).

Also adds a warning if the local `${user_code}.cpp` or `.mortran` doesn't match the `HEN_HOUSE` version (doesn't check additional files like `.macros`).

Example output:

```
================================================================================
EGSnrc version 4 for x86_64-unknown-linux-gnu          Mon Mar  5 14:12:36 2018
================================================================================
configuration............................................test
configuration time.......................................2018-03-05 18:53:05 UTC
app compile time.........................................2018-03-05 18:54:21 UTC
git commit hash..........................................ef1a59a
git branch...............................................feature-git-log
application..............................................tutor7pp
pegs file................................................521icru on HEN_HOUSE
using host...............................................irs25
input file...............................................lab3.2
output file(s)...........................................lab3.2
================================================================================

               ***************************************************
               *                                                 *
               *                  tutor7pp                       *
               *                                                 *
               ***************************************************
```


ftessier:
I like it!! The compile time should be "configuration time", and there should be a way to get the compile time updated within the default application Makefile. We could also report whether the application's source files were modified compared to the repository version.

HEN_HOUSE/scripts/configure
```diff
@@ -2026,6 +2026,14 @@ str_header=$(printf "$str_header" | while read line; do
     printf "\"\n"
 done)
 
+GIT_HASH=" "
+GIT_BRANCH=" "
+CONFIG_TIME=`date -u +'%Y-%m-%d %H:%M:%S UTC'`
+if type "git" > /dev/null 2>&1; then
+    GIT_HASH=`git rev-parse --short=7 HEAD`
```
ftessier: Perhaps add a conditional `if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then`, because users might have git on their system yet will not be using the git clone for EGSnrc.

HEN_HOUSE/scripts/configure
```diff
@@ -2026,6 +2026,14 @@ str_header=$(printf "$str_header" | while read line; do
...
+    GIT_BRANCH=`git branch | grep "^\*" | sed 's/^..//'`
```
ftessier: An cleaner way to get the branch name: `git rev-parse --abbrev-ref HEAD`

mainegra:
Nicely done!

rtownson:
This PR is ready for review! The commits can be squashed I think.

ftessier:
Rebased on `develop`. Simple conflict in makefile for the name of the Windows make program, now set from the variable `$(make_prog)`, see c9cf98c.

rtownson:
Just a note for this one - users will have to run the configure script after this is merged when they pull the new version. Otherwise, applications will fail to compile. I think that's OK since it's what we recommend anyways (or better a fresh install).

rtownson:
Another note here - the git hash and branch are only updated when the users runs the configure script! This means that if they checkout a different branch and don't run the configure script then these parameters will be wrong. Is this acceptable since that's somewhat unusual?

rtownson:
I have now updated this PR so that the git branch and hash correctly update at compile time, rather than only when the configuration is run.

rtownson:
This PR is ready for review. It would be nice to get it into develop to take it for a spin before 2021 release!

ftessier:
@nrc-cnrc/egsnrc How do you feel about merging this in for Release 2021? Safer to merge it into `develop` just after the release, deferring to 2022 for official release?

mainegra (approved):

rtownson:
Sure, we can merge in after the 2021 release.

mainegra (approved):

blakewalters (approved):

ftessier (approved):
Just need to squash the (or some) commits before merging.

ftessier:
Multiple force pushes to fix a mix up while rebasing on the current develop, and tweaking commit title lenght and commit message fill. No code change.

PR #415, opened 2018-03-01
Fix #413: limit vacuum step size in magnetic field

### LINKED ISSUE #413: Magnetic field transport does not work in vacuum
For efficiency, transport in vacuum assumes that electrons necessarily travel in a straight line to the next region, but this is not valid when there is a magnetic field. The region and medium is updated to the next region without checking that `irnew != irold`, which is wrong when transport is limited by the B-field step size in vacuum. Moreover, transport stops after the first B-field step when the next region is `irnew=-1` (outside) due to a similar assumption in the howfar method.

ftessier:
Adjust the charged particle step routine for transport in vacuum when
there is an electromagnetic field (emf). The existing logic failed
because the emf step size restriction was not taken into account before
the howfar call. Hence the howfar method tracked the particle in a
single straight step across the entire vacuum region, causing erroneous
region changes and discards.

### DIFF
--- a/HEN_HOUSE/src/egsnrc.mortran
+++ b/HEN_HOUSE/src/egsnrc.mortran
@@ -1111,6 +1111,10 @@ $start_new_particle;
                     "          direction of motion at the start of the step
                     " Note that tustep and ustep are modified below.
                     " The above provide defaults.)
+
+                    " EM field step size restriction in vacuum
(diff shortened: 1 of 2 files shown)


mainegra:
Very nice @ftessier !

ftessier:
@blakewalters are we good to merge this in its current form?

blakewalters:
Yup @ftessier.

PR #417, opened 2018-03-14
Add Qt5 support

rtownson:
This replaces PR #263.

Adds compatibility with Qt5 to the Qt GUIs. Adds compile-time checks for the Qt version so that the correct functions or libraries are used; i.e. the GUIs are still compatible with Qt4.

Also removes some old tooltip code in egs_inprz that was not used any more, but that caused a visual bug on Windows with Qt5 (the word "toolTipTip" appeared in the centre of the GUI).

Tested on qt-4.8.6 and qt-5.10.1 on linux and qt-5.10.1 on Windows.


crcrewso:
One testing question. Have we tested if the installer breaks if it's rerun overtop of an existing installation after switching QT versions? I only ask because 2018 seems to be the year of QT4's depreciation.

rtownson:
@crcrewso no that shouldn't be a problem, I switched back and forth a number of times using an existing installation while testing the GUIs.

PR #418, opened 2018-03-16
Add egs_phsp_scoring ausgab object.

blakewalters:
This ausgab object allows the user to score phase space files in EGSnrc or
IAEA format for particles entering/exiting a predefined geometry.
See the documentation at the the top of egs_phsp_scoring.cpp for more
details.

It has been tested for a single scoring plane perpendicular to the Z axis.
However, testing for other and more complex geometries is encouraged and
required!

The commit also includes a small change to the check_byte_order() function
in iaea_header.cpp: The function, which is called every time an IAEA header
file is updated, no longer outputs the byte order (along with an outdated
list of associated architectures) to the screen.  This was resulting in
a lot of cryptic screen output when writing IAEA phase space files.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
@@ -28,7 +28,8 @@
 #  A phase space scoring class.
 #
 #  Scores phase space data for particles crossing all surfaces of a user-specified
-#  predefined geometry
+#  predefined geometry.  The phase space geometry must be a component of the simulation
+#  geometry or coincident with a surface of the simulation geometry.
 #
 #  Phase space data can be scored in one of 2 possible formats:
 #
@@ -89,17 +90,31 @@
 #  program their own concatenation routine.
 #
 #  Example:
-#  The following example input defines a phase space scoring plane perpendicular to the Z axis at Z=0.
+#  The following example input defines a phase space scoring plane perpendicular to the Z-axis
+#  positioned at the bottom of a volume of water.
 #
-#  First, the user must define the plane geometry (this could have already been defined as part
-#  of the simulation geometry):
+#  :start geometry definition:
+#      :start geometry:
+#          library = egs_planes
+#          type = EGS_Zplanes
+#          positions = 21.0
+#          name = scoreplane
+#      :stop geometry:
 #
 #      :start geometry:
-#        library = egs_planes
-#        type = EGS_Zplanes
-#        positions = 0.0
-#        name = scoreplane
-#     :stop geometry:
+#          library = egs_ndgeometry
+#          type = EGS_XYZGeometry
+#          name = water_box
+#          x-planes = -15.0  15.0
+#          y-planes = -15.0 15.0
+#          z-planes = 0.0 21.0
+#          :start media input:
+#             media = H2O521ICRU
+#          :stop media input:
+#      :stop geometry:
+#
+#      simulation geometry = water_box
+#  :stop geometry definition:
 #
 #  Then, the user must define the scoring plane ausgab object:
 #
@@ -109,7 +124,7 @@
 #          name = example
 #          phase space geometry = scoreplane
 #          output format = EGSnrc [IAEA]
-#          [constant Z = 0.0]
+#          [constant Z = 21.0]
 #          particle type = all
 #        :stop ausgab object:
 #    :stop ausgab object definition:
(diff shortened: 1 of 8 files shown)


blakewalters:
Excellent suggestion, @mchamberland!  I'll go ahead and put that in....

HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.h
```diff
@@ -0,0 +1,266 @@
...
+
+    int ocharge;           //particle type for output: 0--all; 1--photons; 2--charged particles
+
+    int scoredir;           //scoring direction: 0--on entry and exit; 1--on entry; 2--on exit
+
+    string phsp_fname; //name of phase space file
+
+    string phspoutdir; //output directory
+
+    const EGS_Float prm = 0.5109989461; //precise rest mass
```
rtownson: This was added a while ago! See `egs_phsp_source.cpp`.
```
EGS_Application::activeApplication()->getRM()
```

blakewalters: I figured there was something like that.  Thanks, Reid!  Other changes coming, too.

rtownson:
@blakewalters are you planning to add the ability to score the phase-space between pairs of regions? For example, to score particles moving from regions 2 to 4, 3 to 5, and 4 to 2 (could be done with region labels):
```
from regions = 2 3 4
to regions =  4 5 2
```

Also, in the current implementation particles could curve around and pass through the scoring planes twice (especially with magnetic fields). I feel like there should be an option to stop transporting the particle after it is scored.

HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
```diff
@@ -0,0 +1,718 @@
...
+#  along with EGSnrc. If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+#
+#  Author:          Blake Walters, 2018
+#
+###############################################################################
+#
+#  A phase space scoring class.
+#
```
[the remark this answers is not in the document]
blakewalters: Hi @rtownson:  One of my original ideas for this ausgab object was to use the setLabels function to determine whether particles are "inside" or "outside" of the scoring geometry, but I couldn't figure out a way to retroactively go back and set labels in the transport geometry once it had already been created.  Can you see a way to do this?  Even if region labeling can't be used, though, it should be no problem to score based on entering/exiting certain region numbers provided that the user knows the region numbers in the original geometry.  As for multiple passers, yes, in the current implementation, there's no way to tell if a particle has crossed a scoring plane more than once.  I can add an an option to: 1) eliminate particles once they've been scored, 2) flag multiple passers using LATCH in EGSnrc format phase space files or 3) simply not score > 1 pass of a given particle.  Finally, before merging this feature, yes, I'll transfer the description over to the .h file and format it for doxygen.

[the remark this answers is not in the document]
rtownson: I'm not sure I understand your `setLabels` idea, but could you use `isInside` on the scoring geometry for this? That could work if you want to score upon entering or exiting a given geometry I think. For the idea where the user provides a list of regions (or labels), then you can use labels the same as in egs_dose_scoring.

For multiple passers, it's always good to have options I guess I just don't understand the application of (3). I get confused by the creation of secondaries after the primary is scored - the primary may pass a second time and not get scored but create more secondaries that are scored. Then when you use the phase-space you would have the primary where it was first scored but also secondaries from transport that "hasn't occurred yet". So if you re-ran the simulation from the phase-space you would get a higher dose in the scoring region that in the original simulation.

Maybe there should be an option to not multiple-score the primary *and* its secondaries - this would allow you to continue transport after phase-space scoring (to score multiple phase-spaces for example) but not have signal duplication.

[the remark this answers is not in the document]
blakewalters: @rtownson, see the latest commit.

HEN_HOUSE/egs++/egs_application.h
```diff
@@ -1122,7 +1122,13 @@ class EGS_EXPORT EGS_Application {
     };
     virtual EGS_Float getRM() {
         return -1.0;
-    };
+    }
```
rtownson: This line is missing a semicolon.

blakewalters: Thanks, @rtowson.  There were some other issues as well.  Silly git!  See the amended commit.

maxorok:
Hi guys, great job on this feature! 

Having tested it with an EGS_Xplane at 0.0 for both IAEA and EGSnrc formats, it seems to work nicely. However, would there be any way to get constant values at a plane in the EGSnrc format in addition to that functionality in the IAEA phsp format? Is there a historical reason why or why not? 

Thank you!

blakewalters:
Hi @maxorok: By definition, EGSnrc format phase space files store particle X,Y positions.  Thus, they can only provide meaningful information for particles scored on a plane.  Generally, this plane is assumed to be perpendicular to the Z-axis, and the Z position of the plane is not written to the file.  This convention is traceable back to the original codes used to generate these phase space files, such as BEAMnrc, in which scoring planes always correspond to the bottom surfaces of accelerator components, these surfaces are perpendicular to the beam axis, and their Z location is known.  I suppose you could get meaningful 3D information from an EGSnrc format phase space file scored on an oblique plane if you knew the equation of the plane, but you'd have to recover the Z values using a separate routine.  Hope this answers your question.

Issue #385, opened 2017-11-23
egs++ tracks in vacuum on windows fails


rtownson:
On windows, if you enable particle tracks in egs++ applications and transport through vacuum, the application crashes.

PR #420, opened 2018-04-07
Fix egs++ particle tracks crash on Windows 10

rtownson:
Fix a crash that occurred often when using tracks on Windows 10. The crash started happening with a Windows update, and resulted in a segmentation fault when generating the first track in the simulation. The crash occurred due to accessing an array index out of bounds in the tracks generation (not sure why this didn't crash previously).

Fixes issue #385, from what I can tell.

This is an important functionality fix and should be added to master.

### DIFF
--- a/HEN_HOUSE/egs++/egs_particle_track.cpp
+++ b/HEN_HOUSE/egs++/egs_particle_track.cpp
@@ -187,6 +187,9 @@ void EGS_ParticleTrackContainer::addVertex(EGS_ParticleTrack::Vertex *x) {
 }
 
 void EGS_ParticleTrackContainer::addVertex(int stackIndex, EGS_ParticleTrack::Vertex *x) {
+    if (m_stackMap[stackIndex] < 0) {
+        return;
+    }
     if (!m_isScoring[m_stackMap[stackIndex]]) {
         return;
     }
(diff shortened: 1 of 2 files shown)


mainegra:
@rtownson can you visualize the tracks with egs_view?

rtownson:
Yes egs_view with tracks worked for me while testing this - I used the tutor7pp tracks1.egsinp example with spheres instead of a slab.

ftessier:
This bug should be trapped higher up. If stackIndex is less than zero (out of bounds) at some point, then so is app->Np, according to `egs_track_scoring.h`:
```
int np = app->Np;
if (m_pts->isScoringParticle(np)) {
```
Now `app->Np` is set by `egsAusgab` in in `egs_advanced_application.h`:
```
app->Np = the_stack->np-1;
```
So it seems that `egsAusgab` being called when  `the_stack->np == 0` is the underlying problem. Still looking into it.

ftessier:
Misread the proposed fix. All good! Should the other two instances of `[m_stackMap[stackIndex]]` (lines 284 and 305) be sanitized as well, for peace of mind?

rtownson:
Sure, those additions look good.

PR #427, opened 2018-04-17
Update egs_view layout

rtownson:
**New egs_view layout, settings file, dose viewing and more..**

- Added the ability to save & load a configuration file for `egs_view`. This includes saving the current look orientation, window size, track options, clipping planes, etc. You can load the config file from the GUI, or pass it as an argument (with or without ptracks file argument). E.g.

`egs_view myInput.egsinp myInput.egsview`

- Fixed various segmentation faults that occurred when reloading files. This includes a fix to clearGeometries() in egs_base_geometry so that composite geometries actually get cleaned up. Added an option to load a new input file.

- Changed the layout - now there are tabs to divide sections, and the more commonly used options are together. Also, clipping planes are more intuitive! The key-bindings can be found in the `Camera` tab, and there might be some there you didn't know about (plus a new `double click` functionality).

- Zooming has been modified - at higher zoom levels it switches to scaling with a power-law to slow it down (*let me know if you don't like it and I could make it optional*).

- Added region-by-region visibility, but only for geometries with <=1000 regions (for performance). Added color controls in the `Color` tab of backgrounds, tracks, etc. Also, you can now control the display of vacuum regions.

- The coordinate on the first surface hit by your mouse is now displayed in the viewer when region display is turned on. 

- Fixed a number of display bugs, like the axis flipping direction when the geometry was not near the origin. Added more hunting for an initial point inside the geometry so that it's less likely to crash when not near the origin or windowed properly.

- Added dose viewing (3ddose files). Replace `dosxyz_show` by loading the egsphant in an XYZ geometry and including an ausgab object for dose scoring that points to your 3ddose file. You can use any XYZ geometry. For speed, the dose is stored in an unordered_map, which requires `c++11`, so this was added as a compile flag since qt4 doesn't enable that compatibility by default. 3ddose example for egsphant files:
```
  :start geometry:
        name = my_phant
        library  = egs_ndgeometry
        type = EGS_XYZGeometry
        egsphant file = slice_files.egsphant
        ct ramp = 5MeV.ramp # this requirement will be removed in future PR
  :stop geometry:
  simulation geometry = my_phant
:stop geometry definition:
:start ausgab object definition:
  :start ausgab object:
        library = egs_dose_scoring
        name = example
        medium dose = no
        region dose = no
        :start output dose file:
            geometry name = my_phant
            file type = 3ddose
        :stop output dose file:
  :stop ausgab object:
```
- Added the ability to select the geometry that is being viewed. This includes geometries that are automatically generated, likes planes for an ND geometry, which may have no visual representation but allow for region analysis.

- Added track range selection. 
egs_view1
egs_view2
egs_view3
egs_view4
egs_view5

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.h
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.h
@@ -322,6 +322,11 @@ public:
         dose_geom= dgeom;
         file_type = ftype;
     };
+    bool getOutputFile(EGS_BaseGeometry *&dgeom, int &ftype) {
+        dgeom = dose_geom;
+        ftype = file_type;
+        return output_dose_file;
+    };
     void setUserNorm(const EGS_Float &normi) {
         norm_u=normi;
     };
--- a/HEN_HOUSE/egs++/egs_base_geometry.cpp
+++ b/HEN_HOUSE/egs++/egs_base_geometry.cpp
@@ -133,7 +133,7 @@ public:
         int j = 0, iloop = 0;
         while (nnow > 0) {
             if (geoms[j]->deref() == -1) {
-                delete geoms[j];
+                removeGeometry(geoms[j]);
             }
             else {
                 geoms[j++]->ref();
@@ -144,7 +144,7 @@ public:
 
                 for(int i=0; i<nnow; ++i) {
                     if(!geoms[i]->deref()) {
-                        delete geoms[i];
+                        removeGeometry(geoms[i]);
                     }
                 }
 
@@ -465,6 +465,14 @@ void EGS_BaseGeometry::clearGeometries() {
 
 EGS_BaseGeometry *EGS_BaseGeometry::getGeometry(const string &Name) {
     return egs_geometries[active_glist].getGeometry(Name);
+}
+
+EGS_BaseGeometry **EGS_BaseGeometry::getGeometries() {
+    return egs_geometries[active_glist].geoms;
+}
+
+int EGS_BaseGeometry::getNGeometries() {
+    return egs_geometries[active_glist].nnow;
 }
 
 void EGS_BaseGeometry::setMedium(const string &Name) {
--- a/HEN_HOUSE/egs++/egs_base_geometry.cpp
+++ b/HEN_HOUSE/egs++/egs_base_geometry.cpp
@@ -149,8 +149,8 @@ public:
                 }
 
                 if (iloop > 20) {
-                    egsWarning("~EGS_GeometryPrivate(): failed "
-                               "to delete all geometries after 20 loops!\n");
+//                     egsWarning("~EGS_GeometryPrivate(): failed "
+//                                "to delete all geometries after 20 loops!\n");
 
                     break;
                 }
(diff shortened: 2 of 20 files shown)


mchamberland:
No comments other than thank you and please hurry completing this PR! The load/reload fixes and save/load configurations will be great.

rtownson:
It's functional if you want to start using it (and you have Qt) - I'm just thinking about adding more features :)

victorMalkov:
Looks great, the saving feature will be really good! I guess the options to align with either the x,y, or z axes is still there (probably hidden in the camera tab)? One thought, and I am not sure how many people would find it useful, but maybe it would be possible to have a scale along the axes or else where as an option?

rtownson:
Thanks Victor! Yes I'm thinking about ways to add a grid or measurement tools.

mainegra:
Amazing job Reid! It was badly needed!

mchamberland:
The clipping plane checkbox column doesn't quite fit in the window on macOS and requires the user to scroll horizontally in the clipping plane box. Any way to increase the size of the window? Also, how about making the window re-sizable? Too tricky?

egs_view_mac

rtownson:
Commit 1598570 fixed the OS X layout, some other issues and added region visibility. I tried to get window resizing to work but it is tricky - no luck yet.

victorMalkov:
Hey Reid, the changes look awesome. Would it be possible to put in an option to select the number of particle tracks that you want to be displayed. I know that sometimes I may be viewing tracks and there could just be an overwhelming amount of photons (I still want to be able to see some) and very few electrons. Just a thought. 

Also, I guess it won't be an issue if there are thousand of regions for the regions tab?

mchamberland:
Things are looking really good, @rtownson !

@victorMalkov The regions tab gets greyed out when there are too many regions. I don't know how many is too many, though!

mchamberland:
@rtownson On gcc 4.4.7, the following header is required in `viewcontrol.cpp`:

`#include <algorithm>`

Also, older versions of Qt4 (< 4.7 or so) throw the following during compilation:

`main.cpp:128: error: ‘Monospace’ is not a member of ‘QFont’`

I think the best solution is for the user to upgrade their version of Qt... which I'm trying right now. I'll report back if I get more errors after upgrading.

mchamberland:
Disregard the 'Monospace' error. I was mis-matching Qt versions by mistake.

mchamberland:
Actually.... it turns out I was correct the first time: the 'Monospace' StyleHint is not available from QFont in Qt 4.6. But it's there in 4.7 and 4.8. At least, it is now documented here!

mchamberland:
Have you tested on Qt4? The buttons (load, save, etc.) don't show up for me, but I'm on Qt4.6 (RedHat Enterprise Linux 6.4, so all the dev tools are a little dated...).

rtownson:
@mchamberland I haven't updated the screenshots but now those buttons are in the new `File` menu. Does that work for you? Thanks for reporting the font error I can add a version check.

mchamberland:
@rtownson  I totally missed the File menu. Yep, that works. Thanks!

blakewalters:
Very very nice, @rtownson!  Love being able to see the dose results.  I know it wouldn't add anything quantitative, but might a future version have an option to display isodose lines as well?

rtownson:
@blakewalters I hope to add isodose lines in the future, just haven't put the time into it yet.

Issue #174, opened 2016-09-06
NaN with -ffast-math


mchamberland:
NaN (still) shows up when using -ffast-math with the Fortran compiler. This happens with gcc version 4.4.7 20120313 (Red Hat 4.4.7-4) (GCC).

Sample input file for `tutor7pp` below.

With -ffast-math flag:

`Medium            1  sige =   1.00000000362749373E-015                       NaN  monotone =  T T`

`Medium            2  sige =   1.00000000362749373E-015                       NaN  monotone =  T T`

Without -ffast-math:

`Medium            1  sige =   0.43671621424277174       0.43386623396694124       monotone =  T T`

 `Medium            2  sige =   0.24630246820211257       0.24325696800012028       monotone =  T T`

As far as I know, this does not produce any significant differences in the results of `tutor7pp` (or any other code), but I'm not entirely sure.

[material.txt](https://github.com/nrc-cnrc/EGSnrc/files/457483/material.txt)
[media_bug.txt](https://github.com/nrc-cnrc/EGSnrc/files/457484/media_bug.txt)

crcrewso:
May I ask, which input file were you using? Test1?

mchamberland:
The input file attached to the report: "media_bug.txt"

GitHub does not allow .egsinp extensions.

crcrewso:
Ahhh sorry it threw me not being an egsinp.

rtownson:
Is this still a problem on the current develop branch? I could not reproduce on centos. Thanks!

crcrewso:
Which gcc did you test with? I thought this was a 4.4 series issue.

rtownson:
Good point - I was on gcc 4.8.5.

Kawrakow:
Suggestion: discard the -ffast-math flag. Better (i.e., faster and less dangerous) outcomes with modern gcc versions are achieved using -march=native -mtune=native

ftessier:
Will do, thanks for the info!

Issue #85, opened 2016-05-18
NaNs in user-generated .pegs4dat file


ojalaj:
Hello.

With following `.pegs4inp` file I get NaNs in the generated `.pegs4dat` file:
```
MIXT
 &INP NE=5,RHO=7.8700E+00,RHOZ=0.0045,0.9875,0.0071,0.0004,0.0005,IRAYL=1, EPSTFL=0 &END 
AISI1045
C  FE MN P  S
ENER
 &INP AE=0.521,UE=55.511,AP=0.01,UP=55 &END
PWLF
 &INP  &END
DECK
 &INP  &END
```
Is this normal? At least it crashes the simulations. I've never seen this before, when using pegs4.

PS: Issue #67 also has an effect, when using from command line (at least I wasn't able to succeed in generating .pegs4dat with density corrections). I haven't tested yet if fix #80 help, but will do it, when merged to `develop` branch.

ftessier:
NaNs in pegs4 files can arise due to the `-ffast-math` compiler flag when compiling the `pegs4` executable. Go inside your `$EGS_CONFIG` file and remove `-ffast-math` from the Fortran compiler flags, then go in `$HEN_HOUSE/pegs4` to compile using `make clean` and `make`.

ojalaj:
I tried and NaNs are no longer present - thank you.

The reason why I haven't seen these before might be that I've previously generated all the pegs4 data in a Windows-based EGS/BEAM distribution, so this flag might not be there and/or something else has prevented NaNs from being generated. Since -ffast-math flag was suggest during configuration by default, could this kind of issue be worth mentioning in the documentation?

ftessier:
Good point. The flag is not suggested when using `$HEN_HOUSE/scripts/configure`, but I see it in the GUI configuration tool; @mainegra can you remove the `-ffast-math` default for the Fortran compiler in [`egs_tools.cpp`](https://github.com/nrc-cnrc/EGSnrc/blob/4fb7c06fb8b262834132d2860a8b015e03d7e5b3/HEN_HOUSE/gui/egs_configure/egs_tools.cpp)?

ojalaj:
[...]

PR #432, opened 2018-04-19
Fix #423: change -ffast-math to -mtune=native

### LINKED ISSUE #423: Floating-point exception when concatenating IAEA phase space files with addphsp
I get a floating-point exception (IEEE_INVALID_FLAG) when concatenating IAEA phase  space files. I re-compiled `addphsp` with the `-ffpe-trap=invalid` flag and no optimization to produce the backtrace below:
 
```
[~/Desktop/TrueBeam_v2_phsp_10FFF] Marc$ addphsp TrueBeam_v2_10FFF test 2 0 1 1

 Will sum from phsp file TrueBeam_v2_10FFF_w0.1.IAEAphsp
 to TrueBeam_v2_10FFF_w1.1.IAEAphsp
 And output result to test.1.IAEAphsp


 Adding TrueBeam_v2_10FFF_w0.1.IAEAphsp to test.1.IAEAphsp: 


 Header information for TrueBeam_v2_10FFF_w0.1.IAEAphsp:

  Warning: IAEA format phsp file does not store LATCH

            TOTAL NUMBER OF PARTICLES IN FILE:     45862111
                     TOTAL NUMBER OF PHOTONS:     45427146
THE REST ARE ELECTRONS/POSITRONS.
 
      MAXIMUM KINETIC ENERGY OF THE PARTICLES:       10.377 MeV
 # OF INCIDENT PARTICLES FROM ORIGINAL SOURCE:    324000000

                       Z AT WHICH PHSP SCORED:       26.700 cm



 Header information for test.1.IAEAphsp:


 First time writing to this file.
 No header data to display.


 BEGIN READING/WRITING PH-SP DATA .....


Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x10280c092
#1  0x10280b3b0
#2  0x7fff6a4bdf59
#3  0x1027fe4e6
#4  0x1027f967b
#5  0x1027d733a
#6  0x1027daaa8
#7  0x1027dd3e8
Floating point exception: 8

```

I'm not sure if I can provide anything more than that since those are the Varian phase space files which I'm not allowed to share. Let me know if I can help debug.

ftessier:
Update the default gcc optimization configuration to use -march=native
instead of -ffast-math. The latter causes various floating-point
exceptions on newer cpus and compilers. If the programs are run on a
different cpu, then one should use the corresponding -march option for
that architecture instead of "native", or else use the less aggressive
-mtune=native if the compiling and running cpus are in the same family.

The alternatives -march and -mtune were suggested by @Kawrakow in 
issue #174.

See also #85, #91, #174, #204 and #423.

### DIFF
--- a/HEN_HOUSE/doc/src/pirs509a-beamnrc/pirs509a-beamnrc.tex
+++ b/HEN_HOUSE/doc/src/pirs509a-beamnrc/pirs509a-beamnrc.tex
@@ -665,7 +665,7 @@ The options for {\tt make} are:
 \index{make!options}
 \begin{verbatim}
 make              Compile the accelerator executable with default optimization
-make opt          turned on.  Default optimization is level 3 (-O3).
+make opt          turned on.  Default optimization is level 2 (-O2).
 
 make noopt        Compile the accelerator executable with optimization
                   turned off.
@@ -10362,4 +10362,3 @@ explicitly mention the contributions of many others.
 \typeout{ add page throws if needed                                  }
 \typeout{************************************************************}
 \end{document}
-
--- a/HEN_HOUSE/doc/src/pirs702-egsnrc-codes/pirs702-egsnrc-codes.tex
+++ b/HEN_HOUSE/doc/src/pirs702-egsnrc-codes/pirs702-egsnrc-codes.tex
@@ -2186,7 +2186,7 @@ The options for {\tt make} are:
 \index{make!options}
 \begin{verbatim}
 make              Compile with default optimization
-make opt          turned on.  Default optimization is level 3 (-O3).
+make opt          turned on.  Default optimization is level 2 (-O2).
 
 make noopt        Compile with no optimization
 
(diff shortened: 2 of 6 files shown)


marenaud:
You should be mindful that this change will break some cluster installations of EGS. For example, many users on our cluster install/compile their own version of EGS in their home area, which is mounted on an NFS partition and shared across a heterogeneous cluster (5 nodes, 3 different generations of CPU architectures). Compiling with march=native will produce a binary that is only "valid" for the CPU architecture of the node that the user was logged in on, not other compute nodes.

ftessier:
Very good point @marenaud. Then I suggest settling for -mtune=native (or even -mtune=generic) which is backward compatible within cpu families, or else just leave out -ffast-math and -m options entirely in the default configuration (and provide some guidance about these optimization options in the installation instructions). At any rate the -ffast-math does not seem like a safe default any more.

marenaud:
Agreed, alternatively we could also look at which flag is causing issues. According to the gcc website, -ffast-math is equivalent to:
```
-fno-trapping-math, -funsafe-math-optimizations, -ffinite-math-only, -fno-errno-math, -fno-signaling-nans, -fno-rounding-math, -fcx-limited-range and -fno-signed-zeros
```
If we're lucky, maybe only one of these is causing all our trouble. Not sure I would count on that though...

ftessier:
I would not count on that either! I think `-mtune=native` is a reasonable default, since an environment with different cpu families is unlikely, and one must then _anyways_ handle multiple executables. I will update this pull request with `-mtune`.

ftessier:
Just added `-mtune=native` to the default fortran optimization options in the configure script.

mainegra:
@ftessier this must be also added to the config GUI!

ftessier:
I think it has, see changes in egs_tools.cpp. Can you check if I missed anything?

ftessier:
Also changed the default optimization level from -O3 to -O2

PR #444, opened 2018-05-09
Feature "minor" environment additions

mainegra:
This pull request introduces the tool used at NRC to create compressed and uncompressed archives with files to be distributed with the EGSnrc system. Since the move to github it only includes statically built Qt GUIs and required libraries, although in the past it was used to distribute the whole EGSnrc system.

Several fixes were made to make the process as straightforward as possible. However, one still has to build the Windows component on a Windows machine. Here are the fixes:

- Added or corrected Makefile targets 'realclean' for egspp and the GUIs
- Expanded 'clean' Makefile target for iaea_phsp to include all object files
- Fixed bug in file HEN_HOUSE/distribution/egs_guis_list whereby the egs_track_scoring.dll was not added to the pre-compiled egspp library making track visualization crash on Windows.

### DIFF
--- a/HEN_HOUSE/specs/windows.spec
+++ b/HEN_HOUSE/specs/windows.spec
@@ -23,7 +23,7 @@
 #
 #  Author:          Iwan Kawrakow, 2003
 #
-#  Contributors:
+#  Contributors:    Ernesto Mainegra-Hing
 #
 ###############################################################################
 #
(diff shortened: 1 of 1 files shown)


mainegra:
Resolved conflict in file $HEN_HOUSE/specs/windows.spec due the definition of gui_make. It was set to mingw32-make which is the most likely make utility, but users should not be restricted to MinGW. Using now 
`gui_make = $(make_prog)`
variable make_prog is set during the configuration phase.

rtownson:
Great work, this will save a lot of future hair-pulling! The c++ codes should be run through `astyle`.

mainegra:
Good idea @rtownson !

ftessier:
The distribution tools were moved to a new admin branch, in which develop can be merged before building distribution files.

PR #452, opened 2018-06-07
Fix #359: photon cs parameter not initialized

### LINKED ISSUE #359: Corrupted .egslog file if Photon cross section=  key is missing on input
If this key is missing on input, the code appears to run correctly with the default xcom, but in the log file the line specifying this setting has multiple non-ascii characters which causes some systems to declare the whole file to be binary so that utilities like grep and diff cannot read it.

rtownson:
Fix the photon cross sections transport parameters initialization, when the user does not specify a value.

Although the correct default value was used when this parameter was not specified, some corrupted text was output to the egslst file. **This minor bug did not affect results, only the display of the egslst file in some editors.** Now, the variable is initialized in the same way as the compton cross section variable.

### DIFF
--- a/HEN_HOUSE/src/get_inputs.mortran
+++ b/HEN_HOUSE/src/get_inputs.mortran
@@ -1182,8 +1182,7 @@ $LOGICAL  ecut_inregions,pcut_inregions,smax_inregions,
           pe_inregions,aux_inregions,photonuc_inregions;"Ali:photonuc"
 
 character*15 output_strings(14);"Ali:photonuc, increased by 1"
-character*16 p_xsections;
-save         output_strings,line, p_xsections;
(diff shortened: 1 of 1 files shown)


mainegra:
@rtownson I still don't get what the problem was! All I see in the changes is a switch to p_xsections from photon_xsections! I must be clearly missing something!

rtownson:
I revised the commit to have fewer changes but the same effect. This is safer because the global variable maintains the same name, thus better backwards compatibility.

mainegra:
Thanks @rtownson !

mainegra:
@ftessier can we merge this next week? Would be good to have it in before the Tampere course!

ftessier:
Just rebased on develop to test compilation.

Issue #442, opened 2018-05-08
Incorrect geometry check in SYNCHDMLC


ojalaj:
As originally described in https://plus.google.com/108156691265193140021/posts/7ZbSAQwi26W, code check W_(bottom of support rail) >= W_groove + W_tip should be W_(bottom of support rail) < W_groove + W_tip.

ftessier:
Thank you for reporting @ojalaj :medal_sports:

PR #455, opened 2018-06-17
Fix synchdmlc dimension check and medium bugs

### LINKED ISSUE #441: Feature request: Implementation of FULL leaves to SYNCHDMLC CM
I hope it is not a user error, but at least I can't include FULL leaves in my simulations. Even if I include them in my input files properly, they just seem to be neglected in the simulation. Actually there seems to be two different types of FULL/outboard leaves in Varian HDMLC, so should there be then e.g. FULL1 and FULL60 leaves in the CM? If this is a user error, I apologize and you can ignore this message.

blakewalters:
Fix #441 and #442: Fixes to SYNCHDMLC

1. Change restriction on width of bottom of support rail (W_railbot) for
    quarter and half target leaves from W_railbot >= W_tip + W_groove to
    W_railbot <= W_tip + W_groove in accordance with TrueBeam
    specs.  Accompanying changes to the BEAMnrc and the GUI.
2. Fix bug in full leaf in which much of the leaf cross section
     was left filled with air instead of actual leaf medium.


blakewalters:
Not quite sure why branch fix-call_vcusource_w_zsrcdum is also included with this, but there you have it.

deryayucel:
Sorry but on which file should I change these? I do not want to make a mistake and if you could tell the exact path, I would be glad.

rtownson:
@deryayucel I recommend waiting until this is at least merged into the `develop` branch and keeping that up to date using git. If you really want this fix now though, you could use `git checkout` for this pull request (google how to do that). Or just look in the `Files changed` tab above to see the changes and full file paths.

deryayucel:
@rtownson I am studying on HDMLC and this error forced me to write wrong data to start run. So, if I could do it, it would be better. But do you think it is risky to do it because it is not tested yet? I am not an expert and your opinion matters to me.

deryayucel:
@rtownson I checked Files Changed tab but there are many files which are to be changed. All of them are necessary to fix this? Or each one changes something else and I need only one of them?

rtownson:
@deryayucel I'm not sure these changes fix the error you reported on g+. I submitted that bug as issue #459.

I recommend waiting a couple weeks for these fixes to get merged into develop. If you can't wait, make sure to use `git` for any checkouts so that you can stay on top of changes.

blakewalters:
Hi @deryayucel and @rtownson:  Most of these changes are for the GUI and documentation.  If you just want to run with the corrected restriction on W_railbot and fixed FULL leaves, just apply the changes to the SYNCHDMLC_cm.mortran file.

ojalaj:
Does FULL leaf now allow the geometry of both types of full leaf in Varian HDMLC (#1/target and #60/isocenter)? Even if it would, the point is that since their geometries are different, they can't be included in the input file simultaneously.

blakewalters:
@ojalaj, the FULL leaf is intended to model the FULL type leaf in the VMLC (ie not the HDMLC).  There are no designated cross-sections for the OUTBOARD LEAVES #1 and #60.  In cases where including these leaves is important, my understanding is that people approximate them using FULL leaves.

rtownson:
@blakewalters from what I understand @ojalaj has the designated cross sections for the #1 and #60 outboard leaves. Is it possible that he has different specs? We can address this in a separate issue perhaps.

PR #457, opened 2018-06-19
Fix #447: Fix parallel runs with beam sim. source using phase space

### LINKED ISSUE #447: egs_chamber batch run with egs_beam_source does not work
See discussion:
https://plus.google.com/108037604399060498077/posts/61pJ4XE6AWE

blakewalters:
Previously, a parallel run using a beam simulation source that used a
phase space source either hung or crashed due the fact that the
variables defining the record no. to be read from the phase space
source (INPHSP) and the segment of the phase space source considered
(INPHSP_MIN, INPHSP_MAX) were undefined.  Now, each parallel job is
uses a segment of the phase space source restricted by:

(i_parallel - 1)*(NNPHSP/n_parallel)+1 <= INPHSP <=
(i_parallel)*(NNPHSP/n_parallel)

where n_parallel is the no. of parallel jobs, i_parallel is the job
number, and NNPHSP is the total no. of particles in the phase space
source.

### DIFF
--- a/HEN_HOUSE/cutils/load_beamlib.c
+++ b/HEN_HOUSE/cutils/load_beamlib.c
@@ -153,7 +153,7 @@ char *beamlib_maxenergy_name = beamlib_maxenergy_name1;
 
 DLL_HANDLE handle = 0;
 
-typedef void (*InitFunction)(const int *, const int *,
+typedef void (*InitFunction)(const int *, const int *, const int *,
                          const char *, const char *, const char *,
                          const char *, const char *, int,int,int,int,int);
 typedef void (*FinishFunction)();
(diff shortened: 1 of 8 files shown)


ojalaj:
We tested this with a 'develop' branch from yesterday. Now a parallel run using a beam simulation source that used a phase space source works, but the batch queueing system (slurm) sends an exit code 1 (failed) for each job. However, all the .egslog and .eo contain no errors nor warnings and the combined .egslog is generated. At the end of the simulation a fort.2 file is generated with:
***************** Error:
 job           38 : failed to rewind or lock the control file          -1

I've experienced this kind behaviour with DOSXYZnrc s20 simulations, but if I remember correctly, with them all the other jobs are finished with exit code 0 (OK), but the last one finishes with 1 (FAILED). Is is possible that the parallel jobs are trying to access something they can't find (anymore), i.e., .lock file?

Issue #383, opened 2017-11-21
Trapped particles in xyz geometry used as cd base, inside an envelope


ftessier:
When an xyz geometry is used as a cd base geometry, which is then inscribed inside an envelope, in some cases particles get trapped at a surface xyz edge. This can be reproduced in `egs_view` with the following simple example:

```ruby
:start geometry definition:

    ### big box
    :start geometry:
        name     = big_box
        library  = egs_box
        box size = 20
        :start media input:
            media = med_big
        :stop media input:
    :stop geometry:

    ### small box
    :start geometry:
        name     = small_box
        library  = egs_box
        box size = 10
        :start media input:
            media = med_small
        :stop media input:
    :stop geometry:

    ### xyz box
    :start geometry:
        name     = xyz_box
        library  = egs_ndgeometry
        type     = EGS_XYZGeometry
        x-planes = -2 2
        y-planes = -2 2
        z-planes = -2 1 2
    :stop geometry:

    ### cd box
    :start geometry:
        name          = cd_box
        library       = egs_cdgeometry
        base geometry = xyz_box
        set geometry  = 0 small_box
        set geometry  = 1 small_box
    :stop geometry:

    ### world
    :start geometry:
        name    = world
        library = egs_genvelope
        base geometry = big_box
        inscribed geometries = cd_box
    :stop geometry:

    ### simulation geometry
    simulation geometry = world

:stop geometry definition:
```

In this case particles get stuck on the `(x,2,1)` and `(x,-2,1)` edges, as reported by `egs_view`, e.g.:
```
...
ir=0 inew=1 x=(1.96,2,1) t=0
ir=1 inew=2 x=(1.96,2,1) t=0
ir=2 inew=0 x=(1.96,2,1) t=0
ir=0 inew=1 x=(1.96,2,1) t=0
ir=1 inew=2 x=(1.96,2,1) t=0
ir=2 inew=0 x=(1.96,2,1) t=0
ir=0 inew=1 x=(1.96,2,1) t=0
ir=1 inew=2 x=(1.96,2,1) t=0
ir=2 inew=0 x=(1.96,2,1) t=0
ir=0 inew=1 x=(1.96,2,1) t=0
ir=1 inew=2 x=(1.96,2,1) t=0
ir=2 inew=0 x=(1.96,2,1) t=0

```

(bug reported by Alexandra Bourgouin).

ftessier:
[...]

PR #462, opened 2018-07-12
Fix #388, #383: concave CD, XYZ inside CD, conestack, and box geometry errors

### LINKED ISSUE #388: CD geometry does not properly call howfar() against compound concave geometries
This issue is basically the same that was fixed with PR #95, however this time it occurs with two different geometries composing the CD geometry (e.g., a sphere and a cylinder with mismatched radii). In this case, when a particle leaves one of the geometries, the CD geometry thinks it cannot hit another inscribed geometry and fails to properly call howfar() against the other one (see [here](https://github.com/nrc-cnrc/EGSnrc/blob/develop/HEN_HOUSE/egs%2B%2B/geometry/egs_cd_geometry/egs_cd_geometry.h#L566-L568))

Perhaps a possible fix would be for users to explicitly set a is_convex() flag to false when constructing a CD geometry that they know is concave?

Simple geometry to reproduce the error:

```
:start geometry definition:

:start geometry:
library = egs_box
name = box
box size = 5 5 5
:start media input:
media =  AIR700ICRU
:stop media input:
:stop geometry:

:start geometry:
name        = my_cd_planes
library     = egs_planes
type        = EGS_Zplanes
positions   = -1. 0. 1.1
# No media required
:stop geometry:

:start geometry:
name        = sphere
library     = egs_spheres
midpoint = 0 0 0
radii = 1.
:start media input:
media = AL700ICRU
:stop media input:
:stop geometry:

:start geometry:
name        = cylinder
library     = egs_cylinders
type        = EGS_ZCylinders
radii       = 1. 2.
:start media input:
media = AL700ICRU
:stop media input:
:stop geometry:

:start geometry:
name            = cd_geom
library         = egs_cdgeometry
base geometry   = my_cd_planes
set geometry   = 0 sphere
set geometry   = 1 cylinder
:stop geometry:

:start geometry:
library = egs_genvelope
name = final
base geometry = box
inscribed geometries = cd_geom
:stop geometry:

simulation geometry = final
:stop geometry definition:
```

rtownson:
1. Fix a special case of geometry error where geometries that are convex individually result in a concave cd geometries when combined (issue #388). These geometry errors could be created using a radius-mismatched cylinder and sphere in a cd geometry. Also fix where a small step is taken in cylinder or sphere howfar routines to step by boundaryTolerance/2 instead of boundaryTolerance. This resulted in checks against the boundaryTolerance to fail in the cd geometry.

2. Fix a geometry error where particles could get stuck on an edge of an XYZ geometry used as the base geometry for a CD inscribed in an envelope (issue #383).

3. Fix an infinite loop that could occur in long simulations with a conestack with more than one layer. Previously, particles could get stuck on the surface between layers of the conestack. Example that would crash when simulating 1e7 particles:
```
:start geometry definition:
    :start geometry:
        library = egs_cones
        type = EGS_ConeStack
        name = inner
        axis = 0 0 -2.5545 0 0 1
        :start layer:
            thickness = .321
            top radii = 1.9
            bottom radii = 1.9
            media = air
        :stop layer:
        :start layer:
            thickness = 2.0585
            top radii = 1.9
            bottom radii = 1.9
            media = air
        :stop layer:
    :stop geometry:
    simulation geometry = inner
:stop geometry definition:
:start source definition:
    :start source:
        name    = isotropic_point_source
        library = egs_isotropic_source
        charge  = -1
        :start shape:
           type = sphere
            midpoint = 0 0 -2.5
            radius = 0.01
       :stop shape:
        :start spectrum:
           type = monoenergetic
           energy = 1.1
       :stop spectrum:
    :stop source:
    simulation source = isotropic_point_source
:stop source definition:
```

4. Fix a geometry error where a particle could get stuck in the corner of an egs_box. The error could be produced using:
```
:start geometry definition: 
     :start geometry: 
         library = egs_box 
         name    = the_world 
         box size = 50 50 50
         :start media input: 
             media = air 
         :stop media input: 
     :stop geometry: 
     :start geometry: 
      library = egs_box 
       box size =   2.8 2.8 2.8
          :start transformation:
            translation =    0, 9, 9
          :stop transformation: 
          name = my_box 
          :start media input: 
              media = water 
          :stop media input: 
      :stop geometry: 
      :start geometry: 
         library = egs_genvelope 
         name = the_scenario 
         base geometry = the_world 
         inscribed geometries =  my_box
      :stop geometry: 
simulation geometry =  the_scenario 
:stop geometry definition:
```

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
+++ b/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
@@ -565,11 +565,15 @@ public:
                 }
                 // Skip for concave geometries: particles can reenter the CD geometry after leaving.
                 // Check 2. below will catch these cases.
-                else if ((ixnew < 0 && tb <= boundaryTolerance) && (bg->isConvex()  &&
-                         (g[ibase] && g[ibase]->isConvex()))) {
-                    // (a) is true
-                    return ixnew;
-                }
+                // This follow has been commented out because it fails in cases
+                // where the inscribe geometries are themselves convex but
+                // form a concave shape. Instead, just continue to do more checks.
+//                 else if ((ixnew < 0 && tb <= boundaryTolerance) && (bg->isConvex()  &&
+//                          (g[ibase] && g[ibase]->isConvex()))) {
+//                     // (a) is true
+//                     return ixnew;
+//                 }
+
                 // If a particle approaching the geometry sits on a boundary, we look back to see
                 // if we just entered the geometry (the previous checks fail to catch this case).
                 EGS_Float tb_neg = veryFar;
--- a/HEN_HOUSE/egs++/egs_base_geometry.cpp
+++ b/HEN_HOUSE/egs++/egs_base_geometry.cpp
@@ -431,6 +431,8 @@ EGS_BaseGeometry::EGS_BaseGeometry(const string &Name) : nreg(0), name(Name),
     has_B_scaling(false), has_Ref_rho(false), bfactor(0), rhoRef(1.0),
     nref(0), debug(false), is_convex(true), bproperty(0), bp_array(0),
     boundaryTolerance(epsilon) {
+
+    halfBoundaryTolerance = boundaryTolerance/2.;
     if (!egs_geometries.size()) {
         egs_geometries.addList(new EGS_GeometryPrivate);
     }
@@ -716,7 +718,9 @@ void EGS_BaseGeometry::setBoundaryTolerance(EGS_Input *i) {
     int err = i->getInput("boundary tolerance", boundaryTolerance);
     if (err > 0) {
         egsWarning("EGS_BaseGeometry::setBoundaryTolerance(): error while reading 'boundary tolerance' input\n");
+        return;
     }
+    halfBoundaryTolerance = boundaryTolerance/2.;
 }
 
 void EGS_BaseGeometry::printInfo() const {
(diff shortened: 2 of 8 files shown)


HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
```diff
@@ -565,11 +565,15 @@ class EGS_CDGEOMETRY_EXPORT EGS_CDGeometry : public EGS_BaseGeometry {
...
-                    return ixnew;
-                }
+                // This follow has been commented out because it fails in cases
+                // where the inscribe geometries are themselves convex but
+                // form a concave shape. Instead, just continue to do more checks.
+//                 else if ((ixnew < 0 && tb <= boundaryTolerance) && (bg->isConvex()  &&
+//                          (g[ibase] && g[ibase]->isConvex()))) {
+//                     // (a) is true
+//                     return ixnew;
+//                 }
```
mchamberland: Are you sure this is the right solution? I'm fairly sure you need this specific check. Look what happens in egs_view with this pull request:

this does not look good

This is a simple conestack with mismatched radius. If you make the smaller one transparent in egs_view, it never sees the larger one.

mchamberland: Oh, I think I goofed! My conestack was the only geometry in the simulation, so when a particle exits the smaller cylinder, it leaves the geometry, which is why egs_view can't see anything behind it, I think. If you inscribe the conestack in a box, then it looks fine. My bad!

rtownson: Correct, that's normal behaviour. Thanks for testing!

PR #469, opened 2018-08-16
Fix: Set GUI defaults for ECUT and PCUT.

blakewalters:


mainegra:
@blakewalters I wonder if setting those values to 0 in the GUIs is a better approach. In that case ECUT and PCUT will be set to AE and AP by default. This seems more flexible as it would automatically match the user's intentions with the data. Whadda you think?

blakewalters:
That was along the lines of my original idea, @mainegra: leave the fields blank and then write in 0's for ECUT/PCUT when the .egsinp file is saved so that they get replaced by AE/AP at runtime.  I like the idea of having no GUI presets, inviting the user to be more aware of these values.  So I tried this out and found that BEAM choked with global ECUT (ECUTIN) set to 0.  This is because it uses ECUTIN to precalculate some range rejection parameters and, unlike the individual ECUT(i)'s--which get reset to the value of AE for the medium in region i--ECUTIN remains 0.  DOSXYZ may have a similar issue, although it does run with global ECUT set to 0.  I have to look more closely.  In any case, I backed away from making changes to the applications and just set these GUI defaults.  Based on your raising this concern, though, I think I'll go back to my original idea and make the required changes to the BEAM and DOSXYZ.  Whaddya think?

mainegra:
I like that! Finding in BEAM and DOSXYZ where ECUTIN is used/set and fix it there. The GUI should be agnostic about the physics. If the global ECUT (ECUTIN) is set to 0, and hence should be equal AE for the global medium (right?) then the range calculation for ECUTRR should be just the range down to AE.

blakewalters:
That's right, ECUTIN needs to get reset to AE for the default medium.  I'll go ahead and do this.  I just got a bit nervous about changing BEAM and DOSXYZ yesterday...

mainegra:
@ftessier it would be nice to have this pull request merged before the end of next week. This will avoid calculations bombing out in mysterious ways.

ftessier:
@rtownson can you review?

ftessier:
Just amended the commit title and message.

PR #472, opened 2018-08-25
Radiative splitting ausgab object (AO)

mainegra:
A first step for implementing the variance reduction technique (VRT) _radiative splitting_ via an AO making this VRT available to any C++ application. It simply turns on the intrinsic _uniform radiative splitting_ in `EGSnrc` by setting `nbr_split` to a user-defined value which can have the following effects:

- If `nbr_split > 1`: uniform radiative splitting (URS) is ON
- If `nbr_split = 1`: no radiative splitting
- If `nbr_split < 1`: Radiative events OFF !!!

This should make possible the efficient simulation of electronic brachytherapy sources. It will also allow modelling x-ray tubes and linacs albeit less efficiently than the DBS technique as implemented in `BEAMnrc`. The next step is to implement _directional radiative splitting_ which would increase the efficiency in the simulation of x-ray tubes and linacs.

The syntax to define this AO is simply:

```
:start ausgab object:
  library   = egs_radiative_splitting
  name      = some_name
  splitting = n_split
:stop ausgab object:
```

### DIFF
--- a/HEN_HOUSE/egs++/Makefile
+++ b/HEN_HOUSE/egs++/Makefile
@@ -62,7 +62,7 @@ shape_libs = egs_circle egs_ellipse egs_extended_shape egs_gaussian_shape \
              egs_line_shape egs_polygon_shape egs_rectangle egs_shape_collection \
              egs_voxelized_shape
 
-aobject_libs = egs_track_scoring egs_dose_scoring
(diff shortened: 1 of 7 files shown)


ftessier:
Rebased on develop, squashed commits and adjusted style and commit text.

ftessier:
@rtownson @blakewalters please review!

PR #482, opened 2018-10-02
Fix bug in atomic relaxation of some actinides

mainegra:
This PR fixes the following bug:

**Symptoms:** Simulations crashing (segmentation fault) for Pa, U, Np,
          and Cm when using EADL atomic relaxation.

**BUG:** Number of maximum transitions was fixed in egs_init_relax to 250
     via the local macro $MAXTMP. However some elements have a larger
     number of transitions, the maximum being 270 for curium (Cm).

**FIX:** Increase size of the arrays to 300.

**CHANGE:** Using now global macro $MAXTRANS

> Thanks to Jason Bolton from Nordion who encountered the segmentation fault and to @rtownson who tracked it back to being related to relaxations.


ftessier:
Moved the formatting-only changes to the small-stuff branch, and adjusted the commit message.

mainegra:
@blakewalters could you review this PR?

PR #483, opened 2018-10-02
Fix #176: relax C compiler selection and conf name

### LINKED ISSUE #176: Cannot specify alternative C compiler in configuration script
When running configure from `HEN_HOUSE/scripts`, I can select, say, gfortran4.9 (at least I think I can) and it will continue but fail to determine the version number. 

When I try to select gcc-4.8, or anything other than the generated list, the installer will not let me continue. 

I was hoping to be able to have a few config files each for a different compiler automatically generated using the installer but I will have to do so manually instead.

ftessier:
Update the C compiler selection logic (to match the fortran compiler
selection logic) which allows the selection of any C compiler available
on the search path, or via a full path. Previously, the C compiler name
was constrained to a hard-coded list.

Also adjust the filtering of the configuration file name to enforce a
single .conf extension. Previously, the file name was taken as is, and
any extension starting from the last dot was removed to yield the
configuration base name. This proved error-prone when trying to use
names with version numbers, such as gcc-4.8 for example.

### DIFF
--- a/HEN_HOUSE/scripts/configure
+++ b/HEN_HOUSE/scripts/configure
@@ -170,7 +170,7 @@ echo "yes ($abs_make)" >&2
 # Now see which fortran compiler is to be used.
 # First check the availability of fortran compilers
 echo >&2;
-printf $format "Looking for installed fortran compilers ... " >&2
+printf $format "Looking for available fortran compilers ... " >&2
 f_compilers=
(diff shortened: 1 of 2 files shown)


ftessier:
Updated `configure.expect` to match the modified query strings in `configure`.

PR #484, opened 2018-10-04
Fix #443: SYNCHDMLC zmin output and gui checks

### LINKED ISSUE #443: The level of START for SYNCHDMLC?
If I've understood correctly, the horizontal leaf dimensions are defined at the level of ZMIN of the CM, right? How about START? For VARMLC/DYNVMLC CM the manual text says ZMIN as well, but when I'm using SYNCHDMLC, in my ...._wX.egslog file there is a line with 'Input the starting X (ORIENT_MLC=0) or Y (ORIENT_MLC=1) position at z = XXX', where z > ZMIN, not z=ZMIN, as expected.

rtownson:
Fixes a mistake in the `SYNCHDMLC` output where the printed value of zmin was truncated.

Fixes a couple minor bugs in the GUI related to the MLC cross section checks that I came across while testing this.


ftessier:
@blakewalters please review when you have a chance. Thanks!

PR #492, opened 2018-11-01
Add a spectrum scoring application egs_phd

ftessier:
Provide a sample EGSnrc application which scores a pulse-height distribution in a set of regions specified by a label. Included is a sample input to model a germanium detector inside a lead shield, irradiated by a collimated
point source. Also included are egs_view settings and a viewer screen shot for reference.

ge-example


mainegra:
@ftessier, I wonder if egs_phd would be a more appropriate name to differentiate it from a differential fluence calculation.

ftessier:
Good idea, I will change the name to `egs_phd`.

HEN_HOUSE/pegs4/density_corrections/compounds/cellulose.density
```diff
@@ -0,0 +1,32 @@
+  cellulose
```
ftessier: Can you include a short reference? I find that in that with time this is often useful: knowing where the author got the data, or how it was derived. Same comment for the silicon dioxide file

elgoub: how can  get efficiencies  for this detectors in specific energy ??

blakewalters (approved):

ftessier:
Rebased on develop.

ftessier:
Tweak commits, fix `finalize_egs_foruser`, adjust format of HPGe sample input file. No functional changes.

ftessier:
Squash the commit that only added better descriptions in the supplied density correction files.

gweshitt:
I tested the egs_phd example. The output file ge-example-spectrum.dat contains the correct Emin, Emax and binning, but the 2nd and 3rd columns containing the spectrum data are all zeros. I've looked over the egs_phd.cpp file, but I can't see how this would happen.

mjukdisk:
In the icrm-hpge-example.egsinp you use a density correction file called icrm_dirt. Could you upload that? alternatively post the "formulae" to create it using ESTAR?

rtownson:
Thanks @mjukdisk, see PR #721

PR #509, opened 2019-01-26
Fix windows compilation bug

rtownson:
Fixes a bug where the egs++ library did not compile on windows. This bug did *not* make it into the `master` release, and has only been in the `develop` branch since commit 68a830d.

The failure was due to a typo in the radiative splitting ausgab Makefile, and due to some missing EGS_EXPORT flags in egs_ensdf for the radionuclide source.


ftessier:
Just updated the commit text slightly.

PR #510, opened 2019-01-28
Fix #485: string size mismatch in media inputs

### LINKED ISSUE #485: Density correction files not found after PR #445 on certain Linux environments
With the latest `develop` branch with default configuration options, I tested with EX16MVp an interactive run in pegsless mode and got:

**To the screen:**
```
PHOTONUCLEAR REQUESTED FOR MEDIUM NUMBER  4
 
PHOTONUCLEAR REQUESTED FOR MEDIUM NUMBER  5
 
PHOTONUCLEAR REQUESTED FOR MEDIUM NUMBER  6
 
  PEGSLESS INPUT.  CALCULATING ELECTRON CROSS-SECTIONS.
  density_file                                                                                                                                                                                                                               
 
***************** Error:
Error: Cannot open density correction file:
.density
 
 ```
 
**To the .mederr file under egsrun directory:**
 ```
  For medium: AIR521ICRU
  Composition specified in material data file
  Rho specified in material data file
  GASP specified in material data file
  Density correction file specified in material data file.
  Error: Density correction file.density                                                                                                                                                                                                    $
  does not exist in
  $EGS_HOME/pegs4/density_corrections,
  $EGS_HOME/pegs4/density_corrections/elements,
  $EGS_HOME/pegs4/density_corrections/compounds,
  $EGS_HOME/pegs4/density,
  $HEN_HOUSE/pegs4/density_corrections/elements or
  $HEN_HOUSE/pegs4/density_corrections/compounds.
 ```

**And to the .errors file under egsrun directory:**
```
*************ERROR*************
VALUE SOUGHT: bremsstrahlung correction
INPUT--><--NOT ALLOWED
OPTIONS ARE:
                                      KM
                                     NRC
                                    NONE
                          OFF IN REGIONS
                                    EADL
                                  SIMPLE
```

The culprit seems to be PR #445 and the 80 --> 256 changes therein. Reverting back to 80, everything runs as expected. Interestingly, on another cluster this issue cannot be reproduced. I have an access to both of the clusters and can provide more details on the systems/environments.

### PR DESCRIPTION
**Pull request by rtownson**
**Please review this PR carefully - will the `CHAR_VALUE` change cause problems elsewhere?**

Fixed a bug where simulations would not read the input for the density correction file if specified in a material file. The error looked like this:

```
***************** Error:
Error: Cannot open density correction file:
.density
 
```

This error only occurred on certain systems. It was due to a string assignment mismatch between 256 and 80 characters, resulting in an empty string. The bug only existed since PR #445, which extended one string to 256 characters.

The error can be produced by adding the following to the EX10MeVe.egsinp example, and running on a gcc 4.5.1 system.

```
:start media definition:
 AE=0.521
 UE=18.021
 AP=0.01
 UP=18.5
 material data file=/home/yourUsername/EGSnrc/HEN_HOUSE/pegs4/data/material.dat
:stop media definition:
```

### DIFF
--- a/HEN_HOUSE/src/egsnrc.macros
+++ b/HEN_HOUSE/src/egsnrc.macros
@@ -3387,7 +3387,7 @@ REPLACE {COMIN/GetInput/;} WITH
         ERROR_FLAG,                "0 for no errors, 1 for errors      "
         DELIMETER;                 "Name of the delimeter              "
    character ALLOWED_INPUTS*$STRING32,VALUES_SOUGHT*$STRING32,
-             CHAR_VALUE*$STRING80,DELIMETER*$STRING32;
+             CHAR_VALUE*$STRING256,DELIMETER*$STRING32;
    $REAL     VALUE,DEFAULT,VALUE_MIN,VALUE_MAX;
    $INTEGER  NVALUE,TYPE,NMIN,NMAX,ERROR_FLAG,ERROR_FLAGS,i_errors;
 }
--- a/HEN_HOUSE/src/get_inputs.mortran
+++ b/HEN_HOUSE/src/get_inputs.mortran
@@ -219,7 +219,7 @@ COMIN/GetInput,EGS-IO/;
 CHARACTER*$STRING256 TEXT;      "Used to read in each line of the input file   "
 CHARACTER*$STRING256 KEEPTEXT;  "Used to keep an old line of TEXT              "
 CHARACTER*$STRING256 ORIGTEXT;  "The text without conversion to upper case     "
-CHARACTER*$STRING80  TEXTPIECE; "Used to read a piece of TEXT                  "
+CHARACTER*$STRING256  TEXTPIECE; "Used to read a piece of TEXT                  "
 CHARACTER*$STRING40  DELIM_START;"Start of the delimeter                       "
 CHARACTER*$STRING40  DELIM_END;  "End of the delimeter                         "
 CHARACTER*$STRING32  VNAME;      "The name of the variable being extracted.    "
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by ojalaj** [comment]
@rtownson Do you mean the bug only existed since PR #445, not 485 (it was issue #485, where I reported it). Just for curiosity, why this occurred only _on certain systems_?

@blakewalters & @ftessier, sorry for not being able to reply back to you regarding this issue. I've been (or my cluster) very busy with certain simulations and I started doubting myself on this and would have wanted to perform further tests prior to replying.

**Comment by rtownson** [comment]
My guess is that it is related to compiler version, but I haven't tested that directly on the same system. It didn't crash on one system with gcc 4.8.5, but it did crash on a different system with gcc 4.5.1. It seems likely that the new compiler version is smarter about how to deal with mismatched string sizes.

**Comment by ftessier** [comment]
No worried @ojalaj, any issue you report proves very valuable, as you can see!

**Comment by rtownson** [comment]
While we're here, should we make similar changes in get_inputs.mortran? E.g. line 222, and subsequent uses of textpiece similar to get_media_inputs:
```
CHARACTER*$STRING80  TEXTPIECE;
```

**Comment by ftessier** [comment]
Yes the only consistency issue I see is indeed in `get_inputs.mortran`:
```
READ(TEXTPIECE,ERR=:GI1008:,FMT='(A80)') CHAR_VALUE(I,IVAL);
```
There are also a number of assignments to CHARACTER*16 variables, but given that these worked with 80 chars, I am not worried here.

crcrewso:
> My guess is that it is related to compiler version, but I haven't tested that directly on the same system. It didn't crash on a desktop with gcc 4.8.5, but it did crash on a cluster node with gcc 4.5.1. It seems likely that the new compiler version is smarter about how to deal with mismatched string sizes.

4.4 would be the more important gcc version to worry about as it was RedHat 6's default. I would not be concerned if it's only 4.5.x giving you grief. If you have specific test cases I can test against quite a stack of old GCC

rtownson:
@crcrewso I would always welcome more testing, but I am fairly confident now that this fix will solve the issue. I edited the original post to include how to reproduce the issue.

PR #514, opened 2019-02-17
Fix #489: Allow odd no. of TARGET/ISOCENTER leaves in SYNCHDMLC

### LINKED ISSUE #489: Simulation with SYNCHDMLC FULL leaves not possible
If one tries to use the SYNCHDMLC FULL leaf model to simulate Varian HDMLC, the following problem arises. 5 leaf groups has to be defined - 1 FULL leaf, 13 HALF leaves, 32 QUARTER leaves, 13 HALF leaves and 1 FULL leaf. However, for HALF leaf groups following problems is output:
```
 ***ERROR IN CM   5 (MLC)
 Leaf group   2 consists of TARGET/ISOCENTER pairs but
 does not have an even number of leaves.

 ***ERROR IN CM   5 (MLC)
 Leaf group   4 consists of TARGET/ISOCENTER pairs but
 does not have an even number of leaves.
```
This could be solved by having model for OUTBOARD/FULL leaf #1 and #60 separately and if they would together with HALF leaf groups form even number of leaves on both sides of QUARTER leaves. Or if this is not possible, then just allowing odd number of HALF leaves.

blakewalters:
Allows more accurate modeling of Varian HD 120 MLC.  Accompanying
changes to GUI and documentation included.

### DIFF
--- a/HEN_HOUSE/doc/src/pirs509a-beamnrc/inputformats/SYNCHDMLC.inp
+++ b/HEN_HOUSE/doc/src/pirs509a-beamnrc/inputformats/SYNCHDMLC.inp
@@ -14,9 +14,8 @@
              NGROUP_$SYNCHDMLC = number of groups of adjacent leaves where
                                all leaves in a group are:
                                1. FULL leaves
-                               2. HALF TARGET/ISOCENTER pairs with TARGET leaf
-                                  on the -X (ORIENT=0) or -Y (ORIENT=1) side
-                               3. QUARTER TARGET/ISOCENTER pairs
+                               2. HALF TARGET/ISOCENTER leaves
+                               3. QUARTER TARGET/ISOCENTER leaves
                                NGROUP_$SYNCHDMLC defaults to 3 if set <=0
              MODE_$SYNCHCMLC = 0 for single setting of leaf openings (static
                              field)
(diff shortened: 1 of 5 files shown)


ojalaj:
@blakewalters first of all I'd like to apologize that testing this PR for issue #489 has taken so long.

I'm able to run a short interactive test simulation with different leaf types defined in .egsinp:
1, 1
13, 2
32, 3
13, 2
1, 1

On screen output everything seems as expected, but why in .egslst I get (I've omitted the accurate span values):
  Leaves     type            Y rangeX coordinates of opening
                            spanned (cm)              NEG   (cm)   POS
   1-  1       1      -5.**** -   -4.*****        0.00000        0.00000
   2- 59       2      -5.****** -    4.******        0.00000        0.00000
  60- 60       1       4.******* -    5.*****        0.00000        0.00000

1) According to this there would only be leaves of type 1 and 2 (FULL and TARGET), even if ISOCENTER leaves are defined in .egsinp and correctly shown on run output on screen.
2) Leaf 2 seems to span to the range of leaf 1, but is this because tongue and groove?

Is all this right? For accurate simulation results we need to wait until the beginning of next week.

ojalaj:
Simulation results looked reasonable, however accurate simulations and comparison against measurements in the OUTBOARD/FULL leaf region are still not possible, since the SYNCHDMLC CM does not allow the modelling of two different types of OUTBOARD/FULL leaves (#1 and #60).

blakewalters:
Thanks for the testing on this fix, @ojalaj!  From the messy output you have in your .egslst file, it looks like your system may be having trouble interpreting some Mortran output statements.  In this case, I would trust the screen/.egslog output.  Yes, accurate simulation of O60 and O1 is currently not possible with the current FULL leaf design.  This may change in the future if we either: a) incorporate more leaf types(!) or b) split this into two CMs, one optimized for simulating the HD120 and one for the 120 MLC.

ftessier:
Rebased on `develop`.

blakewalters:
@ftessier: See my comment on PR #520.  There's no reason to hold off on approving/merging this.  Another PR is coming in which restrictions in leaf cross-section dimensions are relaxed, but this PR will be incorporated in it.

PR #516, opened 2019-02-27
Fix egs_view compilation

rtownson:
Fix a bug that caused `egs_view` to not compile, giving an error about including `vector.h`. This include statement has been removed since it is not necessary. This has been tested both on qt4 and qt5.


rtownson:
This PR needs to get merged so that we can update the admin branch for building the GUIs.

rtownson:
The line was added 3 years ago, and I'm not entirely sure why it started causing problems..

PR #518, opened 2019-03-07
Fix a few egs++ compiler warnings

rtownson:
Fix egs++ compiler warnings. This is just for tidyness and doesn't affect any behaviour. Also ran `astyle` on the codes I touched.


ftessier:
Rebased on `develop` and updated commit message.

PR #521, opened 2019-03-14
Fix radionuclide source Auger electrons for EADL relaxations

rtownson:
Fix a bug where the Auger electrons from atomic relaxations during `egs_radionuclide_source` source generation were not produced. This did not affect atomic relaxations in EGSnrc in general, only the radionuclide source when it was set to `EADL` relaxation mode. The Auger electrons were excluded because the relaxation code expected `ECUT` to be provided in kinetic energy instead of total energy. Since Auger electrons are low energy, they were all thus below the cutoff.

Also improve compatibility with the BNL and IAEA ENSDF databases.


ftessier:
Just tweaked the added commit title and text.

PR #525, opened 2019-03-29
Fix #522: More robust rotation checking in egs++ transformations

### LINKED ISSUE #522: isRotation is fragile
When I use transformations, I often get warnings like
```
getTransformation: the rotation specified by
   1 0 0
   0 0.707107 0.707107
   0 -0.707107 0.707107
 is not a rotation
```
In the code it is checked, that `A*A'` is exactly the identity matrix, see [here](https://github.com/nrc-cnrc/EGSnrc/blob/master/HEN_HOUSE/egs%2B%2B/egs_transformations.h#L220).

I think it would be more appropriate to check `norm(A*A' - I) < epsilon`. I can provide a PR if you like to change this.

jw3126:


blakewalters:
Please provide a more descriptive comment for this pull request, @jw3126, perhaps including why this fix is necessary.

rtownson:
@blakewalters there's more description in the issue, #522. Essentially the check against the identity matrix could fail due to floating point errors.

jw3126:
@blakewalters You want me to expand the commit message?

blakewalters:
@jw3126: Maybe the commit message just needs a reference to issue #522 (e.g. "Resolves issue #522").

jw3126:
@blakewalters `More robust roation checking (Resolves issue #522)` ok with that?

blakewalters:
Perfect, @jw3126.  Thanks!

ftessier:
Our commit title format for fixing issues is `Fix #nnn: (...)`; see https://github.com/nrc-cnrc/EGSnrc/wiki/Git-commit-messages. I will update and push to @jw3126, in view of merging this in `develop`.

ftessier:
Adjusted the commit message, and updated the logic to test component-by-component. This is more verbose, but it is clearer and on average more efficient, arguably!

HEN_HOUSE/egs++/egs_transformations.h
```diff
@@ -215,11 +215,71 @@ class EGS_EXPORT EGS_RotationMatrix {
...
-                ryz == 0 && rzx == 0 && rzy == 0 && rzz == 1);
+
+        EGS_Float d2;
+        EGS_Float eps2 = epsilon*epsilon;
+
+        // compute square distance to unit matrix, component-by-component;
+        // return as quickly as possible
+
+        // rxy
+        d2 = rxy*rxy;
```
jw3126: Small nitpick. It would be more straight forward to compare e.g. `abs(rxy)` to `eps` instead of squaring both.

ftessier: Which one is more efficient? (I don't know). I also side with using fabs since it is more explicit.

jw3126: I don't know which is more efficient, but I highly doubt the difference is measurable in any practical example.

ftessier: Just tested it, using a high-resolution c++ timer. With or without optimization, the latest version is slightly faster than the other two, but by no more than about 1%. So no worries...

jw3126: Nice! FWIF At first I thought both implementations would be "equivalent" and the compiler would be free to optimize them both into the same code. But this is not true, as `x^2 < y^2` does not imply `fabs(x) < fabs(y)` in floating point land:
```
x = 1e-200
y = 1e-300
x*x == y*y == 0.0
```

ftessier:
Cleaned up the branch and adjusted commit messages. Keep this as two separate commits on side branch when merging in, to preserve the original commit by @jw3126.

ftessier:
The comparisons with `0` and `1` in the original code are fine because those numbers have exact binary floating-point representations. But in general the (arbitrary) input numbers will carry roundoff errors, so it's worth adding epsilon guards. For the record, @jw3126, I think your error message in #522 arises because the determinant of your sample matrix `1.0000006188980002` and trips the first float guard in `isRotation()`.

PR #530, opened 2019-04-16
Fix DOSXYZnrc parallel runs

rtownson:
Fix a bug in `DOSXYZnrc` parallel runs when they were initiated using the `Run job in parallel` option in the GUI. The `_w` tag was not added to the pardose files, resulting in a crash when the code tried to combine them. This bug did not affect parallel runs using the `-j` command line argument.

### DIFF
--- a/HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
+++ b/HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
@@ -3002,7 +3002,7 @@ IF(IPARALLEL>1 | n_parallel>0)[
    "this file is written directly into the user code directory"
    parname=$cstring(egs_home) // $cstring(user_code) // $file_sep //
              $cstring(work_dir) // $cstring(output_file);
-   IF(n_parallel>0)[
+   IF(i_parallel>0)[
       "have to add _w[iparallel] if not controlled by script"
       parname=$cstring(parname)//'_w';
       call egs_itostring(parname,i_parallel,.false.);


blakewalters:
I don't understand how this bug could happen, @rtownson: The GUI won't submit a parallel job unless "Number of jobs">1, in which case n_parallel > 1 and it seems like there's no issue.  Unless someone's trying to get some functionality out of the GUI that I'm not aware of.

rtownson:
@blakewalters Currently when you set `NCPU` in the GUI, it does not seem to affect the value of `n_parallel`. I just gave it a quick try and couldn't find where `n_parallel` gets set. Currently if you set `NCPU=2` and `PARNUM=1` in the GUI, the pardose file created does not have `_w1` (but the .errors file does).

blakewalters:
Oh, I see, @rtownson: You're talking about using IPARALLEL and PARNUM in the .egsinp file to control the parallel run.  I think this is a bit of a legacy holdover from when we used to use the pprocess script to do parallel runs.  I think now it makes sense to just use the "run in parallel" option from the "Run" window (which, I think works okay) instead of selecting parallel run parameters amidst the main inputs.  What do you think?

rtownson:
@blakewalters Yes you are right, I guess we should instead remove this option from the GUI then instead of doing this "fix".

As a separate issue, maybe we should consider adding another `run in parallel` option to the Run window. Something like `submit to queue` and `run locally`. New users always encounter the `run in parallel` option failing because they don't have a queueing system and just want to use more cores on their single machine. Currently, we just recommend Fred's `egs-parallel` script instead.

blakewalters:
@rtownson, I agree: I think any parallel processing options should be removed from the "Simulation Parameters".  As for offering users the option of running in parallel without a batch queueing system (i.e. just submitting a user-specified number of parallel jobs to run in the background), what we make this the default if the user selects "run in parallel" together with "interactive run"?

rtownson:
@blakewalters Sure that sounds like a good solution.

rtownson:
Closing this PR because it was replaced with PR #545 which has been merged in.

PR #532, opened 2019-04-18
Fix #528: Windows PowerShell bug with .exe in app_name

### LINKED ISSUE #528: egs++ user codes fail to run because of ".exe" in name on Windows
On a fresh installation of EGS_2019, I noticed an issue that's persisted on Windows for at least the last few major releases where egs++ user codes including tutor7pp which I usually use to test EGS on new installs will initially fail with the following message: 

the input file: ...\EGSnrc\egs_home\tutor7pp.exe\tracks1.egsinp does not exist or is not readable

Obviously, the issue is that the directory of the egs++ user code has an extra .exe appended to it when it shouldn't. This is true of all egs++ user codes. 

I was given the workaround at the time of my personal installation but I thought I'd mention it as a bug so that other users wouldn't have this issue. It's also a pretty easy fix so I'll just reiterate what I was told here: 

1. Navigate to ...\EGSnrc\HEN_HOUSE\egs++\egs_application.cpp and open in a text editor

2. Wherever the variable "app_name" gets set, add the following line right below to correct the extra ".exe" characters from appearing (no quotes): 

"app_name = app_name.substr(0,app_name.find_last_of(".exe")-3);"

Then, simply recompile the egs++ directory in HEN_HOUSE and this should be fixed.

rtownson:
Fix a `Windows PowerShell` and `git bash` bug where the .exe extension on applications was included in the app_name variable. This resulted in a runtime error where the input file failed to be found. Now we check to see if there is a .exe at the end of the app_name, and remove it.

### DIFF
--- a/HEN_HOUSE/egs++/egs_application.cpp
+++ b/HEN_HOUSE/egs++/egs_application.cpp
@@ -245,6 +245,12 @@ EGS_Application::EGS_Application(int argc, char **argv) : input(0), geometry(0),
     //
     if (!getArgument(argc,argv,"-a","--application",app_name)) {
         app_name = egsStripPath(argv[0]);
+
+        // In Windows PowerShell, we need to remove a .exe extension
+        size_t exePos = app_name.find_last_of(".exe");
+        if (exePos != string::npos) {
+            app_name = app_name.substr(0, exePos-3);
+        }
     }
     if (!app_name.size()) egsFatal("%s\n   failed to determine application "
                                        "name from %s or command line arguments\n",__egs_app_msg1,argv[0]);


ftessier:
@rtownson there is a problem with this fix: using `find_last_of` find the last occurence in the string of *any* of the characters in the `.exe` character set. For example, for the string `egs_chamber`, the call `find_last_of(".exe")` returns `9`, matching the last `e` in `egs_chamber`. The method `rfind` should be used instead (see #331). I have reinstated the branch `fix-powershell-exe-bug`, can you update and submit as a new pull request? Thanks!

PR #545, opened 2019-07-27
Remove IPARALLEL and PARNUM options from main

blakewalters:
inputs in DOSXYZnrc GUI.

These are really only relevant for parallel runs using
the deprecated pprocess Unix/Linux script.  Moreover,
having these inputs accessible through the GUI never really made
sense anyway, because the script changed them automatically.


ftessier:
Adjusted format of commit message.

rtownson:
@kouadriaek yes, run that command once, and it will run all 6 jobs in parallel and automatically combine the results when they are done.

PR #547, opened 2019-08-29
Fix #9: Increase array defaults, decrease EPSEMFP

### LINKED ISSUE #9: Seg fault with BEAM shared library in DOSXYZnrc with modern gcc versions
if I run a shared library beam source in DOSXYZnrc on 64 bit Ubuntu 14.04, gcc 4.8.4, I get a seg fault in the photon routine of DOSXYZnrc. This occurs in all EGSnrc distributions from 2011 onwards (haven't tested earlier).

The seg fault disappears in the latest version if I compile with no optimisation, OR if I force 32 bit compatibility (and recompile all of HEN_HOUSE/cutils) using the -m32 flag for both C and fortran (also have to download gfortran-multilib and g++-multilib). The fact that it works fine with optimisations on if I force 32 bit compatibility hints to me that there's probably a pointer size problem or int/longint sizing issue or something like that in the interface between fortran and C for the shared libraries, but I don't really know how to pinpoint the problem further.

If this isn't a known problem, let me know and I'll send you a set of input files (BEAM and DOSXYZnrc) and accelerator that will reproduce the issue. I can't confirm that it happens with _every_ shared beam lib but it definitely does with our Novalis beam models.

Here's a short log of what happens, I can run roughly 1k histories before it seg faults (though that's probably random, but it definitely doesn't seg fault immediately).

```
 ***************************************************************

  Histories to be simulated for this run      1000000

  Histories to be analyzed after this run     1000000

 ***************************************************************
   Elapsed wall clock time to this point=       0.678 s


   CPU time so far for this run =       0.671 s


 BATCH #  TIME-ELAPSED  TOTAL CPUTIME  RATIO  TIME OF DAY  RNG pointers

     1          0.0            0.0      0.00    16:07:14   ixx jxx =   97  33 

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7FDADDB07777
#1  0x7FDADDB07D7E
#2  0x7FDADD243D3F
#3  0x43AAC7 in photon_ at dosxyznrc_linux.F:28919
#4  0x43B8EC in shower_ at dosxyznrc_linux.F:29302
#5  0x49B933 in MAIN__ at dosxyznrc_linux.F:1913
```

For reference, that specific line inside photon_ corresponds to:
`GBR1=GBR11(LGLE,MEDIUM)*GLE+GBR10(LGLE,MEDIUM)`

**After more testing: I can reproduce the seg fault with gcc/gfortran/g++ 4.8.4 and 5.1 BUT I can run everything just fine with 4.4.7**

rtownson:
This PR contains a number of suggestions that are up for discussion! There is a tradeoff for compatibility with old systems in moving to larger default array sizes.

1. Increase a number of array size defaults, since modern systems have considerably more memory on average than they used to.

2. **[Edit: I took this change out, so radiative compton corrections are off by default]** Enable radiative compton corrections in all applications. This one is up for debate! From what I understand it requires increasing the stack size, which I have done only for egs_chamber so far in this PR. Should I also increase the stack size for all applications, and do we want this on by default?

3. Decrease EPSEMFP and EPSGMFP, fixing issue #9.

These changes are mainly quality-of-life improvements for users who typically had to make the adjustments manually. Thanks to @ojalaj for providing a list of suggestions.


mainegra:
@rtownson all these changes should definitely be made. I can see that in egsnrc.macros $MXSTACK is only 40. We should definitely increase it although a matter of discussion would be how large. Although memory is readily available these days, we shouldn't be wasteful either.
Regarding the radiative corrections, I agree that right now it is slightly cumbersome to turn them on. If we can prove that having them on by default won't significantly affect simulation efficiency, then we turn them on by default. But if there is a significant penalty in CPU time, given that they have little impact on the results of most cases then keeping them as an option is the way to go. The idea of providing parameter sets for different scenarios comes to mind .... ;-)

blakewalters:
@rtownson: I like these changes to the defaults (max. no. of media = 7!  Phwaat?), but I'm with @mainegra on changing the default setting of radc_flag.  At what energies and for what simulation geometries is this required?

ojalaj:
Why not increasing MXSTACK also in egs_chamber?

Issue #24, opened 2016-01-05
EGS_AliasTable and EGS_SimpleAliasTable are susceptible to round off errors


mchamberland:
I'm not sure if anything can be done to guard against this, but EGS_AliasTable and EGS_SimpleAliasTable are susceptible to round off errors when building the table. [Lines 113-116](https://github.com/nrc-cnrc/EGSnrc/blob/6b1938d3186055017da09bfd37631c8b87c0af32/HEN_HOUSE/egs%2B%2B/egs_alias_table.cpp#L113-L116), and [lines 204-206](https://github.com/nrc-cnrc/EGSnrc/blob/6b1938d3186055017da09bfd37631c8b87c0af32/HEN_HOUSE/egs%2B%2B/egs_alias_table.cpp#L204-L206)  in egs_alias_table.cpp all involve comparing two float variables on which arithmetic operations were applied. 

This might be hard to test, but the following example produces an incorrect table on my system. Generate an EGS_SimpleAliasTable with:

N = 6
f = {1.028976E-5, 0.00019611, 1.028976E-5, 1.028976E-5, 0.00019611, 1.028976E-5}

Uncomment line 221 and a correct table would look like this:
alias table: 0 1 0.142459
alias table: 1 4 1
alias table: 2 1 0.142459
alias table: 3 4 0.142459
alias table: 4 4 1
alias table: 5 4 0.142459

However, on my system I get:
alias table: 0 1 0.142459
alias table: 1 1 1
alias table: 2 1 0.142459
alias table: 3 4 0.142459
alias table: 4 4 1
alias table: 5 5 -0.715083

which is clearly wrong (among other things, the last weight is negative).

Normalizing the weights to 1, i.e. f = {0.023743096, 0.452513809, 0.023743096, 0.023743096, 0.452513809, 0.023743096}, produces the correct table but I'm not sure that this will always be the case.

My compilers are:

```
Marc$ gfortran --version; gcc --version; g++ --version;
GNU Fortran (GCC) 5.2.0
Copyright (C) 2015 Free Software Foundation, Inc.

GNU Fortran comes with NO WARRANTY, to the extent permitted by law.
You may redistribute copies of GNU Fortran
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING

Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
[...]

PR #557, opened 2019-11-14
Fix source collection and simple alias table

rtownson:
1. Fixed phsp support in the `egs_source_collection`, that was due to not defining `setSimulationChunk`. It has been added to match what already exists in `egs_transformed_source`. The bug meant that parallel runs using phase-spaces in a source collection would not have the parallel "chunks" starting at different locations in the phsp. Instead, all chunks would start at the beginning of the phsp, effectively resulting in extra recycling of those particles. In highly parallel or low number of particle simulations, this would affect the statistics and potentially introduce latent phase-space variance artifacts.

2. Fixed a bug in `EGS_SimpleAliasTable` where certain series of weights did not result in correct sampling. For example, using weight = 10 10 1 1 in a source collection resulted in sampling of 10:5.5:1:5.5 instead of 10:10:1:1. The incorrect weighting was triggered only in a relatively rare combination dependent on the order of weights, their values and their sum. Fred describes the solution in the comments below.

`EGS_SimpleAliasTable` is used in `egs_source_collection`, `egs_voxelized_shape`, `egs_shape_collection` and `egs_atomic_relaxations`.

Thanks to Patty Oliver for reporting these issues.

### DIFF
--- a/HEN_HOUSE/egs++/egs_alias_table.cpp
+++ b/HEN_HOUSE/egs++/egs_alias_table.cpp
@@ -128,12 +128,10 @@ void EGS_AliasTable::make() {
         else {
             fcum[i] = 0.5*(fi[i]+fi[i+1])*(xi[i+1]-xi[i]);
         }
-        //egsWarning("i=%d fcum=%g x=%g f=%g\n",i,fcum[i],xi[i],fi[i]);
         sum += fcum[i];
         wi[i] = 1;
         bin[i] = 0;
         not_done[i] = true;
-        wi[i] = 1;
         if (type == 0) {
             sum1 += fcum[i]*xi[i];
         }
@@ -144,6 +142,7 @@ void EGS_AliasTable::make() {
                                   fi[i+1]*(xi[i]+2*xi[i+1]))/(3*(fi[i]+fi[i+1]));
     }
     average = sum1/sum;
+
     for (i=0; i<np; i++) {
         fi[i] /= sum;
     }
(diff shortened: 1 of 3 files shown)


ftessier:
@rtownson can I remove the debugging code which is commented out?

rtownson:
@ftessier you are welcome to remove it - I just left it in case this is a debug test we might want to do in the future.

ftessier:
I don't agree with the previous fix. It solves the sample failure case (10, 10, 1, 1), but this is coincidental and the code remains fragile, because the there is no distinction between failing to find a high (or low) bin, and exhausting the corresponding loop. Therefore extraneous aliasing might be performed, despite all bins being already equalized. In the failure case this triggers an aliasing of the last bin with itself, with biased results. I patched this in a minimally disruptive way by inserting an explicit index for the high (or low) bin that was found, and then breaking away when no such bin is found (alias complete). There is also no reference to the "i" index looping (n-1) times, which makes the algorithm difficult to follow.

I would much prefer a cleaner, more explicit code, e.g.,

```c++
    // normalize distribution
    sum = 0.0;
    for (i=0; i<n; i++) sum += p[i];
    for (i=0; i<n; i++) p[i] /= sum;

    // sort into big and small bins
    for (i=0; i<n; i++) {
        q[i] = p[i]*n;
        if (q[i] < 1) small[nSmall++] = i;
        else big[nBig++] = i;
    }

    // aliasing
    while (nBig > 0) {
        s = small[(nSmall--)-1];
        b = big[nBig-1];
        bins[s] = b;
         q[b] = q[b] - (1-q[s]);
         if (q[b] < 1) {
             nBig--;
             small[nSmall++] = b;
         }
    }
```

ftessier:
@rtownson can you review and try this patch, and let me know how it fares with the problematic examples you have considered? My last commit is meant to be squashed, so that this becomes a single commit to add on `develop`.

ftessier:
I ran the problematic example `(10, 10, 1, 1)` .egsinp file to confirm that the problem is resolved.

mchamberland:
Just chiming in to remind you of issue #24.  Does this fix also resolve #24?

ftessier:
Wow, Issue #24!!!! Thanks for the reminder @mchamberland! Good memory. Indeed I think the proposed fix will solve that issue for the simple alias table. I am reviewing the *regular* alias table in light of this, and glancing at the code is seems it might suffer the same issue. The floating point comparison operation is not the problem here: it is only used to select whether or not to alias the bin (and this decision has an `epsilon` impact in the end). Rather, it is the "uncontrolled" loop indices which trigger extraneous aliasing when, say, either `jh` or `jl` exhaust the loop but there is no bin left to alias. Note also that the `i` index is never used per se, which makes the algorithm harder to understand (the ` i` loop ensures there is enough bin search, and relies on breaks to escape early: fragile construct).

ftessier:
Rebases on `develop`.

ftessier:
Updated the commits. Will leave the 4-commit trail on the side, as it is useful to understand the change history on this issue.

PR #560, opened 2019-11-22
Reduce machine epsilon and add a distance epsilon

ftessier:
Reduce the value of the egs++ floating-point epsilon from 1e-10 to
a value that is 4 times (2 bits away) from the true machine epsilon, and
appropriately for single and double precision versions of the code.

Add a distanceEpsilon constant that is independent of the machine
epsilon, and much larger, but still entirely sufficient for physical
dimensions (well below atomic dimensions in double precision). By
default, the base geometry sets its boundaryTolerance to this
constant. Its value is close to the previous value of epsilon (1e-10),
which should curb any change in floating point error triggers in the
geometry library at this point.

Earlier in #177, the epsilon constant was introduced to remove
hard-coded floating point comparison guards. The 1e-10 value was chosen
because it matched most of the hard-coded values in the geometry
library. However, this epsilon is now starting to be used as a general
float guard (see #525 for example), where 1e-10 is not appropriate, as
it effectively reduces the floating-point precision of the computation.
Hence the need for independent epsilon values for computations and for
geometrical purposes.

### DIFF
--- a/HEN_HOUSE/egs++/egs_base_geometry.cpp
+++ b/HEN_HOUSE/egs++/egs_base_geometry.cpp
@@ -430,7 +430,7 @@ EGS_BaseGeometry::EGS_BaseGeometry(const string &Name) : nreg(0), name(Name),
     region_media(0), med(-1), has_rho_scaling(false), rhor(0),
     has_B_scaling(false), has_Ref_rho(false), bfactor(0), rhoRef(1.0),
     nref(0), debug(false), is_convex(true), bproperty(0), bp_array(0),
-    boundaryTolerance(epsilon) {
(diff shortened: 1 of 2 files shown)


ftessier:
This requires a review of uses of `epsilon` everywhere, and to a lesser extent the `boundaryTolerance` (a lesser extent because effectively its value has not changed much).

ftessier:
Oh wow, good catch, I don't think there is any harm in using `1ULL<<50`?

ftessier:
Rebased on develop and added fixed `1ULL` constant in bit shift operation.

blakewalters (approved):

ftessier:
Rebased on develop.

PR #565, opened 2019-12-13
Fix #554: Compile error when BEAMnrc EEMF macros are used

rtownson:
Fixes a compile error for EEMF beamnrc macros, due to an undeclared variable being used in an error message. Renames the BEAMnrc macros file to fix a typo in the filename. Removes a couple unused constants.

PR #571, opened 2020-01-19
Add eemf copyright

victorMalkov:
This is an update of the copyright to the EEMF macros. The changes are based on the fix that @rtownson introduced in the fix-eemf-compilation as that changed led to a change in file names. Please let me know if there is a better way of adding these changes.


ftessier:
@victorMalkov thank you for the copyright update. We will however keep the licence notice intact, i.e., only change the copyright line (and authors and contributors). I will adjust on the eemf-fix-compilation branch in a few minutes, please check and confirm it is good to go!

ftessier:
Changes have been spliced in #565.

PR #103, opened 2016-06-09
Additional egs++ geometries and shapes

randlet:
As discussed previously this PR introduces a number of new egs++ geometries and shapes outined below (example geometry files for the new geometries are also included).

One thing of note is that there are a couple of other GPL'd files included (gzstream.cpp, gzstream.h & sobol.cpp, sobol.h) that I did not write. Hopefully that doesn't cause any issues for distribution.

Hopefully everything is in OK shape but I recognize that this is a big PR so we may have to go through a few revisions (e.g. do you want this all squashed to a single commit?).

If you need a more formal statement of transfer of copyright than those contained in commit messages then please let me know!
### EGS_AutoEnvelope

An envelope geometry that inscribes a single geometry at one or more locations within a base geometry (think brachy seeds in a phantom). The primary difference between the regular egs++ geometry envelope and the auto-envelope is that, at initialization time the auto-envelope uses a Monte-Carlo routine to discover which regions of the base geometry are occupied by the inscribed geometry.  This allows us to skip bounds checking of the inscribed geometries when in regions of the base geometry that have no inscribed geometries in them.  This makes the transport in the phantom (roughly) independent of the number of inscribed geometries and can give a large gain in efficiency for some simulations.  This library also includes an additional EGS_ASwitchedEnvelope that allows user codes to enable / disable inscribed geometries arbitrarily (we use this for simulating inter-seed effects in TG43 calculations and stepped HDR sources).
### EGS_rz

A wrapper around ND_geometry to simplify the creation of RZ geometries.
### EGS_cSpheres

We've also modified the EGS_cSpheres geometry class to a) implement  getBound, getNRegDir,  and getMass and b) implement a spherical shell class (innermost sphere is considered to be outside the geometry). 
### EGS_glib

A very small shim library that allows you to load geometries from external files (similar to an "include" directive) like so:

:start geometry:
    library = egs_glib
    name = my_external_geom
    include file = /path/to/some/external/my_geom.geom
:stop geometry:

Where the external file is a valid `geometry definition` input block.  The advantages of this over a regular "include" is that you can now use egs_view to view the geometry in your external file and use "my_external_geom" in other composite geometries in your input file. This geometry can also be used to create an XYZ geometry from an egsphant file.
### egs_conical_shell_stack

Allows you to define a shape in a similar way to the egs cone stack geometry.
### egs_spherical_shell

Allows you to create a thin spherical shell shape.  This shell can also be truncated by a conical section of a given half-angle (we use it for simulating beta-emitting eye-plaques).

PR #575, opened 2020-01-31
Fix glib and egs_rz windows compilation

rtownson:
Fix compilation of egs_glib and egs_rz geometries on windows. This bug was introduced in PR #103 that added the geometries.


ftessier:
Good last-minute catch @rtownson! :medal_sports: :cold_sweat:

ftessier:
Sqashed into a single commit and adjusted commit message.

PR #576, opened 2020-01-31
Fix an egs_configure compilation error on Windows

rtownson:
Fix a type mismatch of a variable that resulted in compilation errors. Newer compilers are strict about the type matching correctly, though only the windows compiler mingw seemed to throw an error.


ftessier:
Tweaked commit title.

PR #588, opened 2020-03-23
Parallel job submission without *.lock files

mainegra:
**Brief:** Easy switch to a job submission approach which simply distributes the number of histories equally among the parallel jobs.

**Usage:** Pass command line arguments `-u` or `--urc` to a parallel job submission command.

**Pros:** 

1. No need to create control file for history distribution
2. Remove requirement for locking control file, known to be a source of errors (albeit recently significantly reduced)
3. Potentially slightly faster since no read/write operation to lock file by hundreds or thousands of parallel jobs.
4. No more calculation crashes due to lock file failures in different file systems

**Cons:** 
1. All jobs run the same number of histories, hence in heterogeneous environments the calculation is as fast as the slowest machines.
2. Needs algorithm to check whether all jobs have completed (_done! URC approach_)

# Uniform Run Control (URC) for Mortran apps

Very simple coding required to switch from balanced (lock-file) to uniform run control.
- Add logical blocks to routines `egs_pjob_control` and `egs_pjob_finish` to skip all lock file related operations.
- Added check in `egs_combine_runs` routine of `egs_utilities.mortran` to determine if as many output files (i.e., `*egsdat`, `*pardose`, etc) as parallel jobs were combined.
- If that is not the case, the last job sleeps for a given time, set to 1 s by default in `egsnrc.macros` (integer `$URC-SLEEP`), and repeats the combining process for a number of times, set to one time by default in `egsrc.macros` (integer `$URC-INTERVALS`).

**NOTE 1:** Unlike the `egs++ URCO` implementation, existing output files such as `*egsdat` and `*pardose` files are not deleted at the beginning of a parallel run. This could have undesired effects when re-running a previous calculation where changes have been made to the input and different results are expected. If not all output files are overwritten, or a different number of parallel jobs are requested, for instance, less, old and new output files available at the time the last job finishes will be combined into a final result, which might be far from the expected result.

**NOTE 2:** Although no changes were required in the Mortran apps, several changes were to the RZ apps and DOSXYZnrc to provide more information in the final output files after combining parallel runs.

**Tests:** Calculations using the URC approach were compared to parallel calculations using a locking file mechanism establishing that no differences in the results were introduced. Tested apps:

- BEAM_EX16MVp
- BEAM_EX10MeVe
- DOSXYZnrc using BEAM_EX16MVp as a source of particles
- All RZ apps

# The uniform run control object (URCO) for egs++ apps

_This RCO type could be useful in computing environments where the standard locking file mechanism is problematic. It was implemented to solve issues encountered with standard RCO types on an SGE HPC environment_

A uniform run control object (URCO) divides the number of histories equally among all parallel jobs like the [simple RCO](https://nrc-cnrc.github.io/EGSnrc/doc/pirs898/classEGS__RunControl.html#details), but allows choosing which job(s) will be tasked with combining the results from parallel runs upon their completion. To this end this URC class introduces _watcher jobs_ which check whether jobs have finished and if that is the case, combine them. By default, the last job is taken as the watcher job, but users can define an arbitrary number of them. The default URCO will check for jobs completing for 5 s at 1 s intervals. This behavior can be changed via input file.

A URCO can be requested via input file or by passing the -u or -urc flag on the command line. In the latter case a default URCO is created if no details are provided in the input file.

#### Input block syntax
The syntax to define an UCRO is shown in the example input block below: 
```yaml
:start run control:
   rco type = uniform           # Or simple, balanced
   check jobs completed = yes   # Defaults to 'yes'
   watcher jobs = j1, j2,...,jn # Defaults to last job
   interval wait time = t_ms    # Defaults to 1000 ms
   number of intervals = N_int  # Defaults to 5
   # Some common options for any RCO
   ncase = 20000000000
   calculation = first # analyze, combine
:stop run control:
```

#### Watcher mode
Once a _watcher job_ has finished, it switches to _watcher mode_ and checks for finished jobs an user-defined number of times (_check_intervals_) spaced in time by an user-defined time interval (_milliseconds_). The check for job completion is based on the existance of _file_name_w*.egsdat_ files in the application directory, where _file_name_ is the input or output file name without extension. For this reason existing _file_name_w*egsdat_ files **_are deleted_** at the begining of the simulation.

At each check interval, jobs that have finished are combined into a partial result. If no other job has finished at the next check, no action is taken and the _watcher job_ sleeps until the next check.

#### RCO type

This implementation allows to choose the RCO type via input file. A  job control file [(JCF)](https://nrc-cnrc.github.io/EGSnrc/doc/pirs898/classEGS__JCFControl.html#details) RCO is defined with the RCO type `balanced` and a simple RCO with the keyword `simple`. To use an URCO one must define the RCO type to be of type `uniform`. 

By default, the last job checks 5 times for finished jobs every second for a total time of 5 s:

```yaml
:start run control:
   rco type = uniform           # Or simple, balanced
   # Some common options for any RCO
   ncase = 20000000000
   calculation = first # analyze, combine
:stop run control:
```

In the above example, jobs finishing 5 s or later after the the last job finished are not combined. A manual combination can be requested by changing the `calculation` input key to _combine_ in the `run control` input block. Or the check time interval increased.

**Tip:** If `check jobs completed` is set to `no` the URCO is _identical_ to the simple RCO!

The `run control` input block below defines an URCO with the last job checking after it has finished for other jobs every 1 min for a total of 1 h:
```yaml
:start run control:
   rco type = uniform         # simple, balanced
   interval wait time = 60000 # 1 min in ms
   number of intervals = 60   # 1 hr total time
   ncase = 20000000000
   calculation = first # analyze, combine
:stop run control:
```
**Note:** If all jobs finish before the hour, the watcher job recombines them at the next check cycle.

# Current EGSnrc run control objects 

### Job control file object
The default approach for job submission used by EGSnrc relies on a locking file mechanism whereby the number of histories to be run, the number of histories already executed and the number of running jobs are kept in a job control file (JCF) which has to be locked while being accessed by one of the running jobs. This [run control object](https://nrc-cnrc.github.io/EGSnrc/doc/pirs898/classEGS__RunControl.html#details) (RCO) type is implemented in a [JCF class](https://nrc-cnrc.github.io/EGSnrc/doc/pirs898/classEGS__JCFControl.html#details). This class was introduced in a time when the NRC _computer cluster_ was composed of two very different types of CPUs. Simply distributing histories equally among all CPUs caused the total time of a simulation to be determined by the slowest goup of CPUs. As a consequence, jobs on the slower machines would continue running for a longer time than the jobs on the faster macines. Those CPUs that had finished earlier would seat idle, wasting resources and time. 

The locking file mechanism solved this _bottle-neck_ by balancing the number of histories to be run based on CPU speed. Each job would take smaller chunks of the histories by accessing the locking file until all hisotries had been dispatched. This way, faster CPUs would execute a larger portion of the simulation and the jobs were finished in about the same amount of time.

One _drawback_ of this approach is that ___jobs attempting to access the lock file abort if they have to wait too long___. This could happen in cases where many jobs are finishing at the same time or executing too fast. Another _drawback_ is that the first job **must** start before all other jobs since it is the one that creates the JCF. 

### A simple run control object
An alternative to the JCF class is the [simple RCO type](https://nrc-cnrc.github.io/EGSnrc/doc/pirs898/classEGS__RunControl.html#details) designed to split the simulation equally among all parallel jobs.  The last job is in charge of combining all parallel jobs into the final result. The only issue with using this RCO is that if the last job ___would___ finish before other jobs, those jobs would not be combined into the final result.

#### Using a simple RCO

Add -s or --simple-run to the end of the command for submitting parallel jobs, i.e., 

```
    exb app inp pegs4 p=N -s
```
### Issues with the simple and JCF RCOs
The locking file mechanism as tested in a [Son of Grid Engine](https://arc.liv.ac.uk/trac/SGE) (**SGE**) High Performance Computing (HPC) environment, produces unexpected results:
- Simulations finish early and the final result is based on a significantly lower number of histories. The reason for this is _not well understood_, but it would seem that the implementation of the locking mechanism does not work properly. **Update:** A bug in the URCO implementation was uncovered during the recent URC implementation for Mortran apps which caused the number of histories run by a JCFO to be reduced by the number of parallel jobs. In a large cluster the number of parallel jobs can be over 1000 which explains why jobs were finishing so early!!!. **Test needed to check if this makes the use of the JCFO approach in the SGE cluster possible**
- Many jobs abort reporting failure to access the JCF file. This can be somewhat mitigated by adding a delay between jobs. However there are still jobs failing when playing with delays between 0.1 s up to 1 s. Having to _play around_ with a delay time is not robust.

Another issue observed during the testing phase of the **SGE** environment was that jobs would not start nor finish synchronously as they were placed on the queue. Since both RCO types rely on the jobs being executed synchronously, i.e., the first job starts first and the last job is expected to finish last, this would cause the combined final result to miss some of the jobs. This later issue could be related to the existence of slightly different CPU architectures in a cluster.


blakewalters:
Excellent work (and documentation), @mainegra!  This makes parallel processing so much more portable.  What happens in a "uniform" run if .egsdat files get output at the end of each batch?  Or does egs_application not do that?

mainegra:
@blakewalters thanks! Only egsdat files from completed jobs are combined by the watcher job(s).

blakewalters:
@mainegra, what happens with a uniform RCO if the watcher jobs either don't get submitted properly or fail?  I guess if you're worried about this, it would be motivation to specify a number of watchers evenly distributed over 1-n_parallel.

mainegra:
@blakewalters thanks for reviewing and testing!
Regarding your question as to what is the best approach when jobs are failing or don't get submitted properly, I would say that selecting jobs 1, N/2 and N is a good starting point. I was having issues with when jobs were being executed by the Gridengine job scheduler because of the way I was initially submitting to the queue at the GPSC in Dorval. Once I got the job submission script "properly" working, these issues disappeared.
The URCO gives the user complete freedom in how to devise a working strategy that better suits their working environment.
The locking mechanism has not work properly at GPSC to date.

HEN_HOUSE/egs++/egs_run_control.cpp
```diff
@@ -810,10 +997,27 @@ EGS_RunControl *EGS_RunControl::getRunControlObject(EGS_Application *a) {
     }
     else {
         if (a->getNparallel() > 0) {
-            result = new EGS_JCFControl(a);
+           vector<string> allowed_types;
+           allowed_types.push_back("simple");
+           allowed_types.push_back("uniform");
+           allowed_types.push_back("balanced");
+           int rco_t = irc->getInput("rco type",allowed_types,2);
```
ftessier, line 1004: Can I change the "rco type" input key to simply "type"; I prefer that the input keys don't have cryptic abbreviations in them. Here "type" seems sufficient in the `run control` input block, consistent with "type" in other inputs.

mainegra: Sure!

mainegra: @ftessier could we merge this one?

ftessier:
Coming up, next one on the list!

mainegra:
@ftessier could we finally merge this? Please?

mainegra:
@ftessier any chance this could make it to develop? I thought it was already there, but when I tried at the NRC it wasn't there!

ftessier:
Oh yes, it should be merged, it is reviewed. I will rebase and let Travis run, then merge. Thanks for the heads up.

ftessier:
I see, I still had to go through the branch, that is why my review is held up. Will do.

mainegra:
> I see, I still had to go through the branch, that is why my review is held up. Will do.

I checked it out at NRC and found out that the simple way to pass `-u` or `--urc` does not work with the `run_user_code_batch` script since this argument gets passed as part of `$other_args` to `qsub`, not `egsnrc`. Not a deal (branch) braker, as one can still use it via input files, but the convenience of `-u` is lost. Need to patch the script run_user_code_batch, and potentially yours . It would be a simple fix though. **_Maybe hold back merging until I do that and test?_**

ftessier:
Ok, I will still _clean_ the branch, and we can adjust `run_user_code_batch`. Or, we can let that script rest ;-) and move on to `egs-parallel`, which will take the `-u` or `--urc` flags no problem, because it takes as input exactly the egsnrc command that you run on the command-line for interactive runs.

mainegra:
> Ok, I will still _clean_ the branch, and we can adjust `run_user_code_batch`. Or, we can let that script rest 😉 and move on to `egs-parallel`, which will take the `-u` or `--urc` flags no problem, because it takes as input exactly the egsnrc command that is issued on the command-line for interactive runs (for exactly the reason that we are encountering here, beyond the ability to copy/paste).

I like that idea! It will force me to finally move to the new, better scripts! Are they already in `develop`?

mainegra:
Are we merging this @ftessier ?

ftessier:
Yes, will clean and merge, I think definitely on the books to clean and merge for the imminent release 2021!

ftessier:
Rebased on develop. Still need to squash and cleanup.

ftessier:
Rebased on new `develop` tip; fixed and edited commit messages (alignment, simplify text, typos, etc.). No functional change.

ftessier:
@mainegra I would like to change two things on this branch:

1. cosmetic change the option `--urc` to `--uniform-run` to match `--simple-run` and avoid cryptic acronyms. Oh: the `cavity` manual still talks about `--simple`, but `egs_application.cpp parses `--simple-run`; will fix that too.

2. change how `egs_application.cpp` tracks the run control type. Right now there are two disjointed Booleans `is_simple_run` and `is_unifrom_run`; I suggest we leverage you new `RCO_Type enum` instead, pulling it out of the class and adding an `rco_` prefix to avoid name collisions (staying away from the modern scoped `enum class` declaration).

I pushed a new commit to show changes, let me know what you think.

ftessier:
Applied astyle throughout commits.

ftessier (approved):

ftessier:
@mainegra I cannot compile any mortran application on the `feature-urco` branch (compiler is `GNU Fortran (GCC) 4.8.5 20150623)`: it seems to be balking at the `#ifdef HAVE_C_COMPILER` preprocessor directive:
```
Mortran compilation for examin_dev.f
mortran3.exe: using following I/O units
  raw/hex data file: /home/ftessier/tmp/EGSnrc/HEN_HOUSE/bin/dev/mortran3.dat
  command line files: 
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/src/egsnrc.macros
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/utils/timing.macros
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/lib/dev/machine.macros
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/src/ranlux.macros
  -> appending examin.mortran
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/src/ranlux.mortran
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/utils/xvgrplot.mortran
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/lib/dev/machine.mortran
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/src/egs_utilities.mortran
  -> appending /home/ftessier/tmp/EGSnrc/HEN_HOUSE/src/egsnrc.mortran
  generated mortjob.mortran
  fortran output: examin_dev.f
  mortlst output: examin_dev.mortlst

  Mortran compiling ...  OK


Fortran compiling examin_dev.f using flags ' -fPIC -O2 -mtune=native' and extra objects/libs ' '
Warning: examin_dev.f:3737: Illegal preprocessor directive
Warning: examin_dev.f:3744: Illegal preprocessor directive
/tmp/ccdaJynh.o: In function `egs_combine_runs_':
examin_dev.f:(.text+0x489f0): undefined reference to `egs_sleep_'
collect2: error: ld returned 1 exit status
make: *** [/home/ftessier/tmp/EGSnrc/egs_home/bin/dev/examin] Error 1
```

I am guessing there is a change that has not been committed? The compilation works on branch develop.

ftessier:
Sigh 🙄 problems with `$PATH` while testing. It compiles fine now...

ftessier:
Compilation of Motran applications start failing on commit 67d4608.

ftessier:
It appears that the solution is to change the extension of the failing Mortran applications to `.F` (instead of `.f`), which invokes the C preprocessor:
```makefile
# We use #ifdef ... #endif for parallel processing implementation
# => Fortran file extension must me .F so that the C-preprocessor is
# automatically invoked.
#
FEXT = F
```

ftessier:
Fixed typo in last commit title.

ftessier:
Tweaked commit c71955d title.

ftessier:
Fixed a typo in the make file comments copied over from cavrznrc. I will fix the other files on the `small-stuff` branch.

PR #589, opened 2020-04-03
Increase MXSTACK for relevant apps

rtownson:
Increases the stack size to 900000 for any applications that allow for splitting. The stack uses a fairly small amount of memory for modern machines so I think this is reasonable. We could go up to 2M instead - that was the highest that I found in egs_cbct and egs_kerma (I didn't reduce them to 900k).

I did not change the stack size in:
- egs_cbct
- egs_kerma
- EDKnrc
- g
- SPRRZnrc
- tutor codes


mainegra:
@rtownson stack size was changed for BEAMnrc since  `HEN_HOUSE/omega/beamnrc/beamnrc_user_macros.mortran` is one of the modified files!

rtownson:
@mainegra you're right! I edited my original post.

ftessier:
I agree to increase `MXSTACK` in principle, although this means roughly an additional 30 MB _per array_ sized with this macro. Looking at the modified apps, none seems to define more than a couple of arrays of size `MXSTACK`, so I think this is fine. Why the 900k though, just curious...

rtownson:
I used 900k because that was the largest value already used, except for the 2M in apps that really need a huge stack. Shall I bump it up to 1M just to make it a round number?

ftessier:
Rebased on `develop` and tweaked spacing.

rtownson:
@Fisicomelo please post this question as a new discussion. If possible, include your BEAMnrc input file, or at least the DBS settings and what you've increased `MXSTACK` to. Also include your EGSnrc version.

PR #592, opened 2020-04-16
Add egs_circle_perpendicular shape

rtownson:
Adds a new circle shape where the angle of the surface of the circle depends on the position of the source particle directed toward it.

This shape is specified via
```
:start shape:
    library = egs_circle_perpendicular
    radius = the circle radius
    midpoint = Ox, Oy (optional)
    inner radius = the inner radius (optional)
:stop shape:
```
and delivers points uniformly distributed within a circle. The surface of this circle is always perpendicular to a ray based on the source particle position, when this shape is used as the target in a source that calls the `getPointSourceDirection()` function. For example, in a collimated source when a source shape and target shape are defined, the special properties of this circle only apply when it is used as the target shape. Otherwise, it behaves like an egs_circle.

For each source particle, when `getPointSourceDirection()` is called, the circle shape is rotated so that the circle surface is perpendicular to the vector between the source location, and the origin of the circle. In effect this means that the source particles see a sphere instead of a circle, except that for a given particle, the possible target locations are on the surface of a circle rather than within a sphere.

Thanks to Nathan Murtha for both requesting and helping to debug this feature!

### DIFF
--- a/HEN_HOUSE/egs++/Makefile
+++ b/HEN_HOUSE/egs++/Makefile
@@ -62,7 +62,8 @@ source_libs = egs_collimated_source egs_isotropic_source egs_parallel_beam \
 
 shape_libs = egs_circle egs_ellipse egs_extended_shape egs_gaussian_shape \
              egs_line_shape egs_polygon_shape egs_rectangle egs_shape_collection \
-             egs_voxelized_shape egs_spherical_shell egs_conical_shell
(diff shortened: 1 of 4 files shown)


blakewalters (approved):

ftessier:
Rebased on develop.

ftessier:
Removed EOL whitespace introduced in second commit.

ftessier:
Run astyle on header file.

ftessier (approved):

PR #598, opened 2020-05-14
Update and document the g application to calculate radiative energy loss calculations

mainegra:
This PR updates `g`, the `EGSnrc` application for radiative energy loss calculations. An extensive Doxygen based documentation is also provided which generates an html user manual and a pdf reference manual.


mainegra:
@ftessier this pull request has a commit that fixes issue #602 among many other things, mainly connected with the g app. The documentation is not finished though. I could make a separate PR just to fix PR #602. Perhaps that's what I'll do! I do need to finish documenting this!

ftessier:
As you please @mainegra, opening a new PR for the issue is fine, and indeed probably helps with issue tracking and resolution. If the commit in question is fairly orthogonal to the rest, then just `
`cherry-pick` it to a new branch and open the PR. No need to delete if from here (which commit is it?)

mainegra:
@ftessier I just created a new PR for this. But it is identical to commit 97efc1dd24963b44e4db5abbbea04fbf2529ba53 !

mainegra:
@ftessier , @rtownson , and @blakewalters I have finally finished this update! Feel free to review anytime! I will send @rtownson and Dave the documentation for their review as they have worked with g in recent times as well. Thanks!

mainegra:
@rtownson , yes it is meant to be there. In case one wants to build the documentation in a machine with no EGSnrc installation! The danger is that it could become slightly outdated, but it is minor.

ftessier:
Removed duplicate commits and merge, no diff with original.

ftessier:
Fixed EOL white space.

ftessier:
Adjust commit, reorder, squash, etc. No diff with original branch (except white space).

ftessier:
Rebased on `develop` before merging.

ftessier:
@mainegra I reduced the update to a couple commits, one for the code update, and the other for the documentation. Hope this is ok. No changes with your `g-update` branch, as confirmed by `git diff -w g-update`, diregarding EOL white space (still lots of those, again it must be that Windows editor, no worries!)

ftessier:
By the way, should be not change `accuracy` to `precision` everywhere in EGSnrc, when talking about the target uncertainty. Or maybe "target uncertainty" to be completely clear.

mainegra:
> By the way, should we not change `accuracy` to `precision` everywhere in EGSnrc, when talking about the target uncertainty. Or maybe `uncertainty`, why not, to be completely unambiguous.

Yes, uncertainty is more rigorous in this case. I should change that in the documentation and the code.

ftessier:
> Yes, uncertainty is more rigorous in this case. I should change that in the documentation and the code.

I suggest we create anothe pull request to change this across the board, in all applications. I will first check what the damage would be. It would be nice to capture this change everywhere in one logical commit.

ftessier:
Not so bad, only 148 instances of the string `accuracy` in 45 files, and some of those are probably legitimate cases.

ftessier:
> Not so bad, only 148 instances of the string `accuracy` in 45 files, and some of those are probably legitimate cases.

Here are about 100 cases in the whole repository, which I could quickly identify from `grep` as requiring a change. I'll open another pull request for this! So @mainegra change it right away in `g`, or wait for the system wide change, whatever you prefer.

[egs-accuracy.txt](https://github.com/nrc-cnrc/EGSnrc/files/5157548/egs-accuracy.txt)

mainegra:
image

ftessier:
Exactly!

mainegra:
But we use statistical uncertainty almost everywhere! All of us! Wouldn't a calculated or estimated statistical uncertainty be a measure of the precision of our calculations or measurements? There are different statistical descriptors that could be used to measure precision. What we calculate in our codes is an unbiased estimator of the sample standard deviation or statistical uncertainty (type A?).

dworogers:
Point well taken.  So wherever it says statistical uncertainty, that's
fine. But it just talking about uncertainty I would prefer precision. But I
don't feel strongly about this. As long as it isn't accuracy!

mainegra:
> Point well taken. So wherever it says statistical uncertainty, that's fine. But it just talking about uncertainty I would prefer precision. But I don't feel strongly about this. As long as it isn't accuracy!
> […](#)
> On Tue, 1 Sep 2020, Ernesto Mainegra wrote: But we use statistical uncertainty almost everywhere! All of us! Wouldn't a calculated or estimated statistical uncertainty be a measure of the precision of our calculations or measurements? There are different statistical descriptors that could be used to measure precision. What we calculate in our codes is an unbiased estimator of the sample standard deviation or statistical uncertainty (type A?). -- You are receiving this because you were mentioned. Reply to this email directly or view it on GitHub: [#598 (comment)](https://github.com/nrc-cnrc/EGSnrc/pull/598#issuecomment-685137763)

I have changed `accuracy` to `precision `in the manual and the code in the latest commit. @dworogers I have also fixed the output statements for `muen` and `mutr`. Give it a try and see if you find anything else that should be changed before committing to `develop`. BTW old input files using `accuracy` will still work! 🤓

ftessier:
I have a slight preference for _statistical uncertainty_, but `precision` is easier to work with and requires less qualification in colloquial text and inputs, so let's go with that! For example, you can say _"better precision",_ but not _"better uncertainty",_ and "decreased uncertainty" evokes a double-negative. I think the clunky `statistical accuracy sought` input key should become `precision goal`. I think by and large it is easier to understand `precision`, generically, even for someone who knows little about "uncertainty" in its different guises. Code _output_ ought to be specific and talk about "statistical uncertainty", or "standard deviation", etc., and perhaps include a clear statement about what that means in each context. I agree with @mainegra to keep input strings with `accuracy` for backward compatibility; I suggest we write a warning that such input keys are deprecated and should be updated in the input files.

(Careful if you decide do go ahead and create a pull request for changing accuracy throughout: there are about 50 cases where the term `accuracy` is legitimate, as when the docs talk about comparing the accuracy of EGSnrc compared to EGS4.)

mainegra:
> I have a slight preference for _uncertainty_, but `precision` is easier to work with and requires less qualification in colloquial text and inputs, so let's go with that! For example, you can say _"better precision",_ but not _"better uncertainty",_ and "decrease uncertainty" evokes a double-negative. I suggest that the clunky `statistical accuracy sought` input key becomes `precision goal`. I think by and large it is easier to understand `precision`, generically, even for someone who knows little about "uncertainty" in its different guises. Code _output_ ought to be specific and talk about "statistical uncertainty", or "standard deviation", etc., and perhaps include a clear statement about what that means in each context. I agree with @mainegra to keep input strings with `accuracy` for backward compatibility; I suggest we write a warning that such input keys are deprecated and should be updated in the input files.
> 
> (Careful if you decide do go ahead and create a pull request for changing accuracy throughout: there are about 50 instances where the term `accuracy` is legitimate, as when the docs talk about comparing the accuracy of EGSnrc compared to EGS4.)

@ftessier I changed accuracy to precision in the g app and its manual only when granted. Mind you, it was only at one spot! 😛

ftessier (approved):

mainegra:
@blakewalters could you review this so it can be merged to develop? There is a user wanting to use g and I rather help her with this version which provides much more support via the manual, to save time and efforts!

blakewalters (approved):
Yeehaa!  Way to go Ernst.

ftessier:
Rebased on `develop` ahead of merge. No other change.

Issue #602, opened 2020-05-25
mcdf-xcom reporting issue


teb5911:
For the Mortran codes when mcdf-xcom photon cross sections are selected the egslst and egslog files report photon cross sections as xcom (in the Electron/Photon transport parameter summary). The running text shows “Reading renormalised photoelectric cross section ……. Done” and testing gives different results when ‘xcom’ and ‘mcdf-xcom’ are selected so it seems to be an issue with the summary file rather than with the actual usage of the photon cross sections. It appears to be an issue in the 2019 and 2020 versions of EGS. The selection is reported correctly in egs_chamber output files and in files from a 2018 install.

PR #604, opened 2020-05-27
Print mcdf label in Mortran apps

mainegra:
Mortan apps were not properly reporting whether `mcdf-xcom`, `mcdf-epdl`, `epdl` or `xcom`
photoelectric cross sections were used. This fix was part of a larger PR (PR #598) that awaits review. 
It has been separated for speedier merge to develop. This PR fixes the recently posted issue #602.

Given how small the fix is, it should incorporated quickly into develop!


ftessier:
> Given how small the fix is, it should incorporated quickly into develop!

Ok, standing by for the 3 required reviews!

Issue #605, opened 2020-06-01
Suggestion: Add binary data storage options for EGS_ScoringSingle and EGS_ScoringArray


MartinMartinov:
I've been running simulations computing 3ddoses in 512x512x117 phantoms, where the actual particle simulation runs pretty quickly, but the all the storeState/setState calls (I suspect) are bottlenecking the simulation significantly.  I'm currently setting nbatch to 1 to avoid too many state calls which works well interactively, but when running several simulations in parallel accessing the same hard drive, it is still a huge efficiency killer when they all go to write egsdat files at once (once again, just a suspicion).

Would it be possible to write a parallel set of store/setState which save data in binary format (ie, storeBinState and setBinState) which usercodes that have massive arrays can use instead?  I don't mind giving it a try myself, I'm just not sure how well it would fit into the current architecture, so I wanted to post here first.

rtownson:
This would be a useful addition! Much of the work has already been done, in terms of comparing binary formats - see PR #405.

mchamberland:
I’d like to add my voice to @MartinMartinov. It is also my suspicion that I/O is a major bottleneck of  certain types of simulations. I had also come around to using nbatch = 1 and I had also noticed the bottleneck of all parallel jobs writing out their results at approximately the same time.

Also, this realization made me immediately suspicious of Monte Carlo timing studies (especially GPU vs CPU) that do not explicitly distinguish radiation transport vs all other tasks (initialization, I/O, etc.).

rtownson:
Another mitigation technique is to work in RAM, by having all I/O happen in /dev/shm for example. Then just copy over final results to a hard-drive.

PR #606, opened 2020-06-02
Add binary storeState and setState methods for efficient I/O

MartinMartinov:
This commit is a follow up to issue #605.

Added binary analogues to setState and storeState (setBinState and storeBinState) to be
implemented in applications using a data stream separate from data_out.  When
implemented in applications for outputting very large data arrays, it provides a 
significant speedup in downtime between batches.  Below is an example of an
application outputData implementation using the new functions.

```c++
int egs_mird::outputData() {
    // Invoke base class first
    int err = EGS_AdvancedApplication::outputData();
    if (err) {
        return err;
    }
	
    // Create new binary egsdat file (begsdat) for faster data storage
    string ofile = constructIOFileName(".begsdat",true);
    ostream *data_bout = new ofstream(ofile.c_str(), ios::binary | ios::out);
	
    // Store score
    if (!score->storeState(*data_bout)) {
        return 101;
    }

    delete data_bout;
    return 0;
}
```

WARNING:  All data is output as it is stored in memory, and can occupy several times as
much space as analogous text files.

### DIFF
--- a/HEN_HOUSE/egs++/egs_scoring.h
+++ b/HEN_HOUSE/egs++/egs_scoring.h
@@ -149,6 +149,20 @@ public:
         return data.good();
     };
 
+    /*! \brief Stores the state of the scoring object into binary stream
+	  \a data.
+
+      This function is analogous to storeState, but assumes istream is open
+	  with ios::binary.  This is intended to be used in a file
+	  analogous to the .egsdat file (suggest it be named .begsdat) 
+	  handled by a custom usercode so as not to interfere with the
+	  EGS_Application and EGS_AdvancedApplication implementations.
+    */
+    bool storeBinState(ostream &data) {
+        data << current_ncase << sum+tmp << sum2+tmp*tmp;
+        return data.good();
+    };
+
     /*! \brief Set the state of the scoring object from the data stream \a data.
 
       The data extracted from the stream is the 16 bit integer representing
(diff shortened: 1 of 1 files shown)


MartinMartinov:
After some testing on an EGS_ScoringArray holding 12,320,768 unique dose values, using the above binary reading functions, the total output time for nbatch=10 went from 55 s to about 9 s, which is far more reasonable for my simulations.  Some things to note:

- The ~10 MB egsdat file is now on the order of a few kilobytes, but the binary begsdat file I generated was ~60 MB.  So this method definitely requires way more space.
- My simulations were performed on a system with an SSD, simulating on an HDD might not offer the same speeds (especially with the increased file size).
- One upside to the massive filesize is that the doubles used for EGS_ScoringSingle maintain the same exact values as when they were stored in memory, as opposed to the (what looks like) 6 digits that a normal egsdat file uses.

ftessier:
@MartinMartinov the branch `patch-1` is not in your repo anymore (I know, it's been two years!). I can take it from here and create a new pull request with these changes. But let me know if you would prefer to reinstate `patch-1` at your end and that I continue working on it here.

MartinMartinov:
I'd actually abandon this one.  I could get it to work for my application egs_mird, but not any of the stock ones.  Unfortunately.

PR #610, opened 2020-06-19
Fix egs_kerma FD bugs and expand its capabilities

mainegra:
Introduces new features in `egs_kerma` and fixes several bugs in the _forced-detection_ algorithm. The `linear track-length` scoring has not been modified as it was working as expected.

**Bug fixes:**
- Kerma calculation with _forced-detection_ (FD) VRT: Use proper equation
based on the integral over the path across the scoring region.
- Account for the possibility of multiple entries to scoring volume when ray-tracing along the photon's directions (FD). This is relevant for hollow geometries such as spherical or cylindrical shells.
- Fix link to `Common input syntax` page from documented `egs++` apps manuals.

**New features:**
- _Forced detection_ can now be requested for individual calculation geometries.
- A default FD geometry can be also provided. In that case, calculation geometries
without their FD geometry will use this one. If no FD is to be used, this option
must be omitted.
- The rationale behind requiring an FD geometry is to guide the _ray-tracing_ towards the scoring regions (using a `howfar` call) which might only cover a small solid angle. In cases of sources embedded inside scoring regions the actual calculation geometry can be defined as the FD geometry. The logical checks to activate FD are:
  - Is there an FD geometry?
  - AND
     - Is particle inside a scoring region? OR 
     - Is particle outside FD geometry and directed towards it? OR 
     - Is particle inside FD geometry?
If the FD geometry is not judiciously chosen, time is wasted ray-tracing photons which
might no be crossing scoring regions.
- Scoring kerma _in more than one region_ is possible if their individual masses
are provided.
- Masses for each scoring region are required, unless one is only interested
in kerma in the whole volume, in which case only the mass of the whole scoring
volume required (original approach).
- Acccept _environment variables_ in file name:
  `E*muen` file name can start with an environment variable now.
- Dropped terms such as `cavity region`, `cavity mass`, `cavity geometry` in favor of
`scoring region`, `scoring region masses`, `FD geometry`, etc. Old inputs will fail
with a message instructing the user what input key should be used.
- Fluence is still calculated for the _whole sensitive volume_. Option for estimating
fluence in each scoring region might be added in the near future.
- Updated documentation reflecting the current status
- Add two new examples to egs_kerma
      - depth dose in water for 5 keV photon point source
      - photon fluence outside a thick lead shield for 20 MeV e- point source
- Added two new `E*muen/rho` files calculated with `g-app` and `mcdf-xcom` option:
      - water 1 keV to 1.5 MeV
      - air   1 keV to 20 MeV
- Renamed `emuen_icru90_1.5MeV.data` -> `emuen_rho_air_1keV-1.5MeV.data`


ftessier:
😲 Nice!

mainegra:
@rtownson this pull request requires updating the egs++ documentation!

mainegra:
@ftessier, @blakewalters and @rtownson I know this looks like a huge undertaking, but I have validated the results over and over again and the kerma and fluence results agree very well with other MC codes! 😉

ftessier:
Rebased on develop.

ftessier:
Working on the commits, not changing any code. Work in progress...

ftessier:
Cleaned up commits, reordered, squashed some small ones, etc. No changes to the files (except EOL white space removal and astyle formatting).

PR #615, opened 2020-07-17
Fix the addphsp usage text

rtownson:
Fix the addphsp usage text to describe the required input better. This is now consistent with phsps produced from egs++, instead of assuming BEAMnrc.


ojalaj:
Should the exclude section in parentheses in the changed line contain also the .IAEA extension?

ftessier:
Taking into account @ojalaj's comment, modified the text to `phsp file base name (excluding the trailing _w*.*)`.

PR #621, opened 2020-08-11
Fix minor bug in DOSRZnrc kerma calculation

rtownson:
Fix the DOSRZnrc kerma calculation so that sub-threshold fluorescent photons (`iarg=33`) are included in the kerma. This bug meant that `D/K` and `D/Kcol` were not equal to 1 for some cases, up to about a 0.8% difference. This makes the algorithm slightly different from the g app, where only primary photons are considered, because DOSRZnrc includes the kerma from secondary photons.

Add some output formatting improvements to DOSRZnrc. Also add the `IDC` or I Double Count variable to allow the user to avoid including k.e. released by any bremsstrahlung or 511 photons. Some comments are also added to improve the readability of the code, and a never-reached code block was removed.

Thanks to Dave Rogers and Ernesto for all their work on these improvements!

### DIFF
--- a/HEN_HOUSE/user_codes/dosrznrc/dosrznrc.mortran
+++ b/HEN_HOUSE/user_codes/dosrznrc/dosrznrc.mortran
@@ -1,3 +1,6 @@
+"Aug 2020 fluor discards included in K during_pe_compt"
+"during_pe_compt and during_eii   and IDC are here
+"also includes hack to print primary and scatter components for IFULL=3 "
 %C80
 "#############################################################################"
 "                                                                             "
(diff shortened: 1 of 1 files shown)


rtownson:
This Travis CI build failed on egs++ for some reason, and I don't seem to have an option to restart the build. @ftessier are you able to do this, so I don't have to push a new commit to trigger it?

ftessier:
Compilation fails, reportedly on building the egs++ library; don't know why, and no modification here related to egs++. I am able to build this branch without issue locally (?)

ftessier:
Relaunched the build and it passed.

PR #622, opened 2020-08-20
Fix egs_chamber usage of latch for splitting

rtownson:
Fix the egs_chamber russian roulette splitting algorithm so that the latch variable is no longer used. Instead, an extra stack is created with a splitting variable in it. This avoids allowing egs_chamber to overwrite the latch, in case other objects such as `egs_phsp_scoring` need to use it.

Previously, if you used russian roulette in egs_chamber and scored a phsp using egs_phsp_scoring, the multiple passer flag for scoring particles may have been incorrect. I think this would be a small effect.

### DIFF
--- a/HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp
+++ b/HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp
@@ -108,6 +108,16 @@ extern __extc__ void F77_OBJ(photo,PHOTO)();
                 the_stack->latch[np1] = the_stack->latch[np]; \
                 the_stack->ir[np1] = the_stack->ir[np]; \
                 the_stack->dnear[np1] = the_stack->dnear[np]; \
+                the_extra_stack->nbr_splitting[np1] = the_extra_stack->nbr_splitting[np]; \
+
+// Extra stack to hold variance reduction data per particle
+struct EGS_ExtraStack {
+    int nbr_splitting[MXSTACK];
+};
+
+extern __extc__ struct EGS_ExtraStack F77_OBJ_(extra_stack,EXTRA_STACK);
+static struct EGS_ExtraStack *the_extra_stack =
+                 &F77_OBJ_(extra_stack,EXTRA_STACK);
 
 /*! a class for storing the phase-space temporarilly */
 
@@ -115,8 +125,8 @@ extern __extc__ void F77_OBJ(photo,PHOTO)();
 class TmpPhsp {
 public:
     // constructor
-    TmpPhsp() : np(0), ntot(4), p(new EGS_Particle[4]) {};
-    ~TmpPhsp() { delete [] p; };
+    TmpPhsp() : np(0), ntot(4), p(new EGS_Particle[4]), nbr_split(new int[4]) {};
+    ~TmpPhsp() { delete [] p; delete [] nbr_split; };
     void grow() {
         int ntot_new = 2*ntot;
         if( ntot_new > MAXPHSP ) egsFatal("TmpPhsp::grow(): exceeded maximum size %d\n",MAXPHSP);
@@ -124,6 +134,11 @@ public:
         for(int j=0; j<np; j++) pnew[j] = p[j];
         delete [] p;
         p = pnew; ntot = ntot_new;
+
+        int *nbr_split_new = new int [ntot_new];
+        for(int j=0; j<np; j++) nbr_split_new[j] = nbr_split[j];
+        delete [] nbr_split;
+        nbr_split = nbr_split_new;
     };
 
     void set() {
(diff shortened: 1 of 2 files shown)


ftessier:
@rtownson Ok, waiting on the test results. Do you need precision simulations? If I read the changes correctly, simulations with these changes should be numerically equivalent to previous ones using latch, no? In this case, testing can be done with very short simulations.

ftessier:
I note that the check failed due to the egs++ failing to compile. This is a recurrent problem recently with Travis. It fails egs++ compilation, but upon relaunching the check it often succeeds (without changing the branch at all). So there seems to be something flaky on the Travis side, I have not figured it out yet...

ftessier:
@rtownson is this ready to go in your opinion, or do you seek more testing ahead of merging? If so, can you turn this into a draft pull request (and optionally add `[WIP]` in the title?

rtownson:
I do still want to take it for a spin, and haven't had a chance yet. I'll turn it into a draft..

ftessier:
Let's create a WIP label!!

rtownson:
@ftessier I do get different numbers, within the statistical uncertainties. My understanding was that any change to the code that requires recompilation can potentially produce a different result for the same seeds.

ftessier:
> @ftessier I do get different numbers, within the statistical uncertainties. My understanding was that any change to the code that requires recompilation can potentially produce a different result for the same seeds.

I will look into that. In principle yes, but in this particular case I fail to see why there would be a numerical difference, especially given that most of the changes here pertain to integer variables! The compiler optimization might create somewhat different code paths though, so I will try to compile with optimization off to check.

ftessier:
@rtownson I am testing this with a full exradin A12 chamber simulation depth dose, with TmpPhsp and xcse on. As you said, the results differ numerically, and I am trying to track this down. I have created a code where `the_extra_stack->nbr_splitting` is assigned and shadows `the_stack->latch`, but is never used for comparisons etc. I am perplexed, because at some point the two values diverge, but I have not yet found where or why this happens... Continuing to track this down.

ftessier:
Oh! While writing my previous comment it occured to me that latch is initialized from getNextParticle, but the_extra_stack->nbr_splitting _isn't!_ Yep, that seems to be the issue.

ftessier:
**Confirmed:** upon adding `the_extra_stack->nbr_splitting[0] = 0;` after `getNextParticle` in `simulateSingleShower`, this new version of `egs_chamber.cpp` produces identical numerical output as before. This is confirmed with a full dose distribution for an exradin A12 ionization chamber in water, using both TmpPhsp and cross-section enhancement.

rtownson:
I was just working on this, but you beat me to the solution! Thanks Fred!

HEN_HOUSE/user_codes/egs_chamber/egs_chamber.macros
```diff
@@ -52,6 +52,21 @@ REPLACE {$RANGE-DISCARD;} WITH {;
...
+  common/extra_stack/ nbr_splitting($MXSTACK), nbr_splittingI;
+  $INTEGER            nbr_splitting, nbr_splittingI;
+};
+APPEND {;COMIN/STACK-EXTRA/;} TO {;COMIN/STACK/;};
+REPLACE {$TRANSFERPROPERTIESTO#FROM#;} WITH {
+    X{P1}=X{P2};Y{P1}=Y{P2};Z{P1}=Z{P2};IR{P1}=IR{P2};
+    WT{P1}=WT{P2};DNEAR{P1}=DNEAR{P2};LATCH{P1}=LATCH{P2};
+    nbr_splitting{P1}=nbr_splitting{P2};
+};
+
```
ftessier, line 69: Should we add this for completeness, even though the `EXCHANGE-STACK` macro is never used, as far as I know:
```

REPLACE {$EXCHANGE-STACK(#,#);} WITH {
;
FDUMMY = U({P2});     U({P2})     = U({P1});     U({P1})     = FDUMMY;
FDUMMY = V({P2});     V({P2})     = V({P1});     V({P1})     = FDUMMY;
FDUMMY = W({P2});     W({P2})     = W({P1});     W({P1})     = FDUMMY;
FDUMMY = E({P2});     E({P2})     = E({P1});     E({P1})     = FDUMMY;
FDUMMY = WT({P2});    WT({P2})    = WT({P1});    WT({P1})    = FDUMMY;
IDUMMY = IQ({P2});    IQ({P2})    = IQ({P1});    IQ({P1})    = IDUMMY;
"LATCH IS NOW STANDARD"
IDUMMY = LATCH({P2}); LATCH({P2}) = LATCH({P1}); LATCH({P1}) = IDUMMY;
"add this for the extra stack
IDUMMY = nbr_splitting{P2};
nbr_splitting{P2}=nbr_splitting{P1};
nbr_splitting{P1}=IDUMMY;
}
;

File: HEN_HOUSE/user_codes/egs_chamber/egs_chamber.cpp [hunk 2]
```diff
@@ -108,22 +108,38 @@ extern __extc__ void F77_OBJ(photo,PHOTO)();
                 the_stack->latch[np1] = the_stack->latch[np]; \
                 the_stack->ir[np1] = the_stack->ir[np]; \
                 the_stack->dnear[np1] = the_stack->dnear[np]; \
+                the_extra_stack->nbr_splitting[np1] = the_extra_stack->nbr_splitting[np]; \
+
+// Extra stack to hold variance reduction data per particle
+struct EGS_ExtraStack {
+    int nbr_splitting[MXSTACK];
+    int nbr_splittingI;
```
ftessier, line 116: The `nbr_splittingI` is not used; leave it there nevertheless?

rtownson: It is used in `TRANSFERPROPERTIESTO#FROM#` on the mortran side, but indeed I tried removing it and it seems to both compile and run. I just copied what was done in cavity, but I'm pretty sure it can be removed.

blakewalters (approved):

ftessier (approved):

ftessier (approved):
Rebased on `develop` and squashed into a single commit.

PR #628, opened 2020-08-27
New egs-parallel set of scripts to replace run_user_code_batch

ftessier:
This pull request implements `egs-parallel`: **a new set of bash scripts to submit EGSnrc parallel jobs,**  with the following improvements over the legacy command `exb` (aliased to the `run_user_code_batch` script):
- Detailed account of the job submission is saved in a log file (and to the terminal with the `--verbose` option)
- Helpful messages and usage is provided when an error occurs during the job submission (or with the `-h` option)
- Script level synchronization significantly reduces errors and race conditions for parallel jobs and `.lock` files
- The EGSnrc command passed to the script is identical to the interactive one (copy and paste!)
- Options can be provided on the command-line in arbitrary order, in short (-) or long (--) form
- Single-letter short options don't require a space before their argument.
- The top-level `egs-parallel` script dispatches a specific sub-script for each method (e.g., `egs-parallel-pbs`)
- The `egs-parallel` script can launch jobs locally on a multicore computer, with the `--batch cpu` option
- Scripts for other schedulers are easily added as new sub-scripts, without having to modify `egs-parallel`
- A decent `egs-parallel-clean` script is provided to help tidying up intermediate simulation files and logs
- The implementation is orthogonal to legacy scripts, so it is backward compatible (you can continue to use `exb`)

```
    usage:

        egs-parallel [options] -c | --command 'command'

    options:

        -h | --help         show this help
        -b | --batch        batch system to use ("cpu" by default)
        -d | --delay        delay in seconds between individual jobs
        -q | --queue        scheduler queue ("long" by default)
        -n | --nthread      number of threads ("8" by default)
        -o | --option       option(s) to pass to job scheduler, in quotes
        -v | --verbose      echo detailed egs-parallel log messages to terminal
        -c | --command      command to run, given in quotes
```

Here is a sample invocation:
```bash
egs-parallel --batch pbsdsh -q short -n12 -v -c 'egs_chamber -i slab -p 521icru'
```

The scripts `egs-parallel` and `egs-parallel-clean` are placed inside a new `$HEN_HOUSE/scripts/bin/` directory, which is added to the path in the EGSnrc shell additions. The sub-scripts are only meant to be called from the top-level `egs-parallel`, so they are not placed inside this new `bin/` directory  to prevent calling them directly (they remain in `$HEN_HOUSE/scripts/` which is not added to the path).

Take this out for a spin if you will, but remain cautious (especially with the cleaning script!).


crcrewso:
This looks so much cleaner!!! 
What would it take to add support for Slurm?

ftessier:
@crcrewso For slurm, we need _someone_ to copy the `egs-parallel-pbs` script to `egs-parallel-slurm`, and adjust it accordingly!

ftessier:
I know the scripts are a little dense: I am trying here to resolve the lock file issue without changing the EGSnrc code, which is perhaps a bit contorted (see by comparison @mainegra's cleaner uniform run control object solution in #588). I found that separating the two roles of the lock file (indicating that the job is running or not, and serving the number of histories) resolved a lot of the issues. There is now a `.egsjob` file, accessed by the first job only, which serves to indicate when the job has started and when it is finished. So perhaps now I can go back an integrate this `.egsjob` file inside the EGSnrc code... On the other hand, when this file is created at the script level, _before the job is launched on the node,_ it curbs issues related to scheduler delay launching the first job. At any rate, I have found this approach more stable. And combined with @crcrewso's improvement on the lock file delay jitter!

There is also a lot of code repetition between the sub-scripts, which is not ideal for code maintenance: we have to remember to copy changes to all the sub-scripts. But this is intentional, as I wanted it to be easy to create new sub-scripts for other schedulers (slurm!) or to suit their particular needs, without imposing any logic beyond the arguments passed to the script. If code maintenance becomes an issue, I'll think of something else; I am not worried about this for now.

crcrewso:
I'm not entirely sure if I'll have a test system for slurm until these quarantimes are over but if I do I'll definitely submit something.

mainegra:
@ftessier are these scripts tied up to using the locking file mechanism? The reason I ask is that in HPC environments where the locking file does not work, one might want to resort to the URCO mechanism (uniform load on all jobs) and then these script will not work. But perhaps this not an issue since in those cases, one might have to use different scripts, such as the one I created for the GPSC. I will try these scripts on the GPSC and see what happens!

ftessier:
For the moment yes they are :-( (except the -cpu script, which does not try to synchronize with the `.lock` file). I have to add an option to skip this for a uniform run control object (urco)... In the end, the lock file checking etc. should be done in the code, because otherwise options are obscurely correlated between the input file (selecting the urco) and the script option to skip the lock file. I don't know yet what is the best way forward. Perhaps just leave the `.egsjob` file in the script, and let EGSnrc worry about the `.lock` file as before; maybe this is sufficient, I will try!

ftessier:
In the end, the lock file should only be checked and handled by either the script _or_ EGSnrc, not both, for exactly the reason you point out: we don't want to have to change the submit scripts if we change the lock file handling in EGSnrc.

ftessier:
@mainegra Could the first job still create an (empty) `.lock` file when starting with the uniform run control object (urco), even if it is never accessed again? This would be informative for the user, to know which job is running, etc. Perhaps the lock file could contain information on the total number of histories requested, how many jobs, and how many histories per job? The `.lock` file is useful on the submit script side to confirm that the job has indeed started.

mainegra:
@ftessier in practice it is possible, but the name (extension) would be misleading.

ftessier:
@mainegra Thanks for the suggested improvement! I removed the dependencies on the lock file in the egs-paralllel scripts. The script will still prevent launching the simulation if there is a `.lock` file though (or a `.egsjob` file), however this can now be overriden with a `-f | --force` option in the top level `egs-parallel` script.

mainegra:
> @mainegra Thanks for the suggested improvement! I removed the dependencies on the lock file in the egs-paralllel scripts. The script will still prevent launching the simulation if there is a `.lock` file though (or a `.egsjob` file), however this can now be overriden with a `-f | --force` option in the top level `egs-parallel` script.

@ftessier that's great news! That way we xould potentially use your scripts on the GPSC as well!

mainegra:
@ftessier I am giving these scripts a try! Quick question: Why is there a delay at the beginning? I turned on the verbose option and got a headache! 🤯

mainegra:
> @mainegra Thanks for the suggested improvement! I removed the dependencies on the lock file in the egs-paralllel scripts. The script will still prevent launching the simulation if there is a `.lock` file though (or a `.egsjob` file), however this can now be overriden with a `-f | --force` option in the top level `egs-parallel` script.

@ftessier I tried using the urco with this and it failed complaining about the lock file not being there ... are you sure the dependency is gone?

ftessier:
Indeed, the scripts still contain a `while` loop waiting for the lock file. Hmmm, strange. The only explanation I can come up is that I did not properly commit all changes back then... Let me try again!

ftessier:
Updated to _truly_ remove the lock file dependencies in the `egs-parallel` scripts, which was supposed to have been done (probably forgot to add these deletions to the commit earlier).

mainegra (approved):
Awesome!

rtownson (approved):

ftessier:
I found that after removing the lock file checking in the scripts, I still _occasionally_ encountered race conditions on the lock file, but the scripts failed "gracefully". So I will leave it as is, given the convenience of being able to use the uniform run control object _sans_ lock file, and the principle of not checking the lock file in two places. I have been using `egs-parallel` for months without any significant issue. I find the `-b cpu` very useful for working-up simulations on a multi-core cpu without a job scheduler!

ftessier:
@rtownson I am going to modify the commits about the leading `.` in job names: I will just trim all the leading dots. What do you think?

HEN_HOUSE/scripts/bin/egs-parallel
```diff
@@ -0,0 +1,251 @@
...
+        *)     cmd="";;
+    esac
+    verbosity="verbose"
+    log "$(basename $0): line $lineno: $msg"; $cmd; log "QUIT."; exit 1
+}
+
+### link file descriptor 3 to egs-parallel log file
+exec 3>egs-parallel-$$.log
+
+### default option values
```
crcrewso, line 88: Is it possible to define these in a more user centric way? A separate file?

HEN_HOUSE/scripts/bin/egs-parallel
```diff
@@ -0,0 +1,251 @@
...
+        -q | --queue        scheduler queue ("long" by default)
+        -n | --nthread      number of threads ("8" by default)
+        -o | --option       option(s) to pass to job scheduler, in quotes
+        -f | --force        force run, even if lock or egsjob file present
+        -v | --verbose      echo detailed egs-parallel log messages to terminal
+        -c | --command      command to run, given in quotes
+
+    example:
+
+        $(basename $0) --batch cpu -q short -n12 -v -c 'egs_chamber -i slab -p 521icru'
```
crcrewso, line 54: inconsistent use of Pipe, example does not have pipe in it.

HEN_HOUSE/scripts/bin/egs-parallel
```diff
@@ -0,0 +1,251 @@
...
+    cat <<EOF
+
+    usage:
+
+        $(basename $0) [options] -c | --command 'command'
+
+    options:
+
+        -h | --help         show this help
+        -b | --batch        batch system to use ("cpu" by default)
```
crcrewso, line 43: It would be great if supported batch systems were listed here. 
Am I misreading the code or can we no longer define some of these as global parameters in the bashrc file?

HEN_HOUSE/scripts/bin/egs-parallel
```diff
@@ -0,0 +1,251 @@
...
+
+
+### help function
+function help {
+    log "HELP"
+    cat <<EOF
+
+    usage:
+
+        $(basename $0) [options] -c | --command 'command'
```
crcrewso, line 38: inconsistent use of Pipe, example does not have pipe in it.

crcrewso (commented):
Slurm is growing in popularity with Canada  Compute infrastructure. Is it possible to get slurm supported? Do you need help with that?

HEN_HOUSE/scripts/bin/egs-parallel
```diff
@@ -0,0 +1,251 @@
...
+
+### default option values
+opt_batch="cpu"
+opt_queue="long"
+opt_nthread="8"
+opt_delay="0"
+opt_command=""
+opt_options=""
+opt_force="no"
+verbosity="silent"
```
crcrewso, line 96: Is it possible to define these in a more user centric way? A separate file?

HEN_HOUSE/scripts/bin/egs-parallel
ftessier: Good catch, this is not a pipe, it was meant as an "OR"; very confusing I will change it. Thanks.

HEN_HOUSE/scripts/bin/egs-parallel
ftessier: Good idea. Indeed, the `egs-parallel` scripts are completely orthogonal to the legacy `run_user_code_batch` method. Here, the supported batch systems are defined by scripts in `$HEN_HOUSE/scripts/` named `egs-parallel-*`, so if you create a new `egs-parallel-slurm` (ahem, for example 😉 ), them slurm becomes available. It is straightforward to check the scripts folder from `bin/egs-parallel` to list the available options though. Will not do this for this merge, but will work on it afterwards. 👍🏻

HEN_HOUSE/scripts/bin/egs-parallel
ftessier: These are not pipes but option "ORs". I think this is pretty standard and I find it clear (except in the example usage line, for sure!).

HEN_HOUSE/scripts/bin/egs-parallel
ftessier: Yes, that's a good idea. I won't change it for this merge, but will give it some thought for sure! I would prefer to gather all user settings in as few places as possible. Perhaps these egs-parallel defaults could go in `$EGS_CONFIG` (and even adjust the configuration script to define them when configuring EGSnrc)?

ftessier:
I thought `${jobname##*(.)}` would do it, but it turns out that the first character must be `[:alnum:]`, so I will have to rely on `sed` I guess to trim all leading non alphanumeric characters.

ftessier:
Can remove non :alnum: leading chars with `${jobname/+([^[:alnum:]])/}`, but I will use a `sed` pipe, which is POSIX compliant (I think!) and independent from Bash: `echo ${jobname} | sed 's/[^[:alnum:]]*//'`

crcrewso:
One comment I should have  made earlier, the format of -c matching the command to directly run the simulation is amazing! Biggest selling feature of the new system from a user perspective!

ftessier:
Agreed, I enjoy that too... it's the little details that make life easier. 😄

ftessier:
- Rebased on develop.
- Removed `|` pipe symbol in `egs-parallel` usage sample command (but not in list of options, stand for "OR").
- Adjusted job name to strip leading non-alphanumeric characters.
- Corrected typos in commit messages.

blakewalters (approved):

blakewalters:
@ftessier, these scripts aren't expected to work seamlessly on OSX, correct?  For example, on a (my) Mac, the line:
cpu_nthread=$(grep -c processor /proc/cpuinfo)
in egs-parallel.cpu doesn't actually return the no. of cpu's (I think it's "sysctl -n hw.ncpu"), resulting in a "unary operator expected" warning in the comparison on the next line of the script.  The job seems to be submitted okay, though!

ftessier:
Good point @blakewalters. I always forget that there is no `/proc` on BSD systems. I have added a small commit to trap the `Darwin` os name and issue the proper command that you provided. I don't know if these are going to work on Windows (my guess is no). Anyone willing to try, say, under Windows git bash or WSL?

Issue #634, opened 2020-09-17
Infinite loop in g appication if pegs4 data too low in energy


dworogers:
I accidentally used the wrong pegs4 file which only went up to 2 MeV when I was calculating for a 10 MeV photon beam.  The output .egslog just stops with no warning message but continues to use 100% of cpu. It needs a check.
I am using the feature-g-update branch version of g which worked fine when I switched to pegsless mode.

ftessier:
Nice find! I have not found the solution yet (well, beyond running with the proper pegs4 file 😉 ). Looking at this through the debugger, the code appears to be locked up in the `electr` subroutine...

ftessier:
A bit of investigating with `gdb` reveals that an electron ultimately gets stuck because `ustep .EQ. 0` in `IF ((ustep .EQ. 0 .OR. medium .EQ. 0))` on line 14174 (of the final `.f` Fortran code). So I know why the code hangs, but not yet why `ustep==0`. At any rate, gracefully trapping the error that the max energy is above the pegs limit at the outset would curb the issue; I would not try to trap this inside `electr`!

dworogers:
Hi Frederic,
I agree that we want to catch it on input. The rz codes must catch it
somehow. I didn't have the patience to track how to find the highest energy
in a spectrum, but presumably it is done somewhere.

Thanks for looking at this.

As you say, using the right pegs4 data is the best solution, or even just
pegsless.

Dave

ftessier:
@blakewalters can you suggest off-hand where we could add a check on the upper pegs energy, at the outset?

mainegra:
@dworogers this would also happen if one uses an energy above the energy limits of the data regardless of whether one uses pegs4 or goes pegsless, right? I looked inside g and everything needed to perform the check is there. We just have to put the check, perhaps right after the call to the inputs routine. I could do it if @blakewalters is not already on it! 😜

dworogers:
It seems to me we need to add a variable, perhaps to comin user (since
it is in subroutine ensrc already) which subroutine ensrc initializes
as the highest energy in a spectrum or EIN and then in main, when we
have access to UE,UP we can do the check.

ftessier:
> @dworogers this would also happen if one uses an energy above the energy limits of the data regardless of whether one uses pegs4 or goes pegsless, right?

[...]

PR #635, opened 2020-09-21
Check maximum source energy against production threshold UE, UP

blakewalters:
Addresses issue #634

### DIFF
--- a/HEN_HOUSE/user_codes/g/g.mortran
+++ b/HEN_HOUSE/user_codes/g/g.mortran
@@ -327,6 +327,7 @@ $INTEGER     nbatch, "$INTEGER and $REAL are defined in egsnrc.macros"
              ibatch, "they can be used (if employed consistently throughout"
              i,j,i_log;    "the user code) e.g. to switch to double precision"
 $INTEGER ifailed;    "Number of failed attempts to sample energy above AP"
+$REAL ekmax;
 
 "---------------------------------------------------------------------"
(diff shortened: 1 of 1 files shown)


blakewalters:
@mainegra, @ftessier, @rtownson: Here's a check to address the bug that @dworogers found when trying to run with Ekmax > UE, UP.  Perhaps a bit overengineered, but it should do the trick.

ftessier (approved):
Sounds good!

mainegra:
@blakewalters thanks for looking into this! 
There is only one issue with the `mono=0` case: Here one can enter one or more individual energy values and they _must not be in increasing order_, although most of the time people will enter individual energy values in an ordered manner. Hence we must check for the maximum energy during the input loop over all energies found in `subroutine source` in the block between lines 1471 and 1474.

blakewalters:
I hadn't accounted for that option with mono=0, @mainegra.  Good catch!  Look forward to another commit.

blakewalters:
Okay, @mainegra, @rtownson and @ftessier: I think I've now accounted for the option to input multiple energies under the mono-energy option.  Please take a look & approve or disapprove when you get the chance.

mainegra (approved):

rtownson (approved):

ftessier:
This pull request was closed automatically upon merging the target branch #598 `feature-g-update` into `develop`. I did not realize that this was the target of another PR. I have reinstated the branch and I am reopening this PR.

PR #645, opened 2020-11-02
Fix the dose values in egs++ 3ddose output

rtownson:
Fix the value of dose in 3ddose files generated from egs++ apps using `egs_dose_scoring`. The dose was not being scaled by the correct mass/density of materials, because the `getMass` function in the ND geometry was not functioning properly for XYZ geometries unless they were defined using an egsphant file or density file. In other words, the dose was wrong if the XYZ geometry boundaries were defined using "slabs" or "planes". 

This is a critical bug in the value of dose reported and has been an issue since the ability to generate 3ddose files was added to `egs_dose_scoring`. Any old dose in 3ddose files generated this way should be considered unreliable, due to the undefined value of density, and those simulations should be repeated.

Note that the doses reported by `egs_dose_scoring` by material, or by region were correct. Only the dose in 3ddose files from `egs_dose_scoring` has this issue. All 3ddose files from DOSXYZnrc are not affected.

Big thanks to our reddit user u/5xum for reporting this issue, and @marenaud for determining it was a bug.

### DIFF
--- a/HEN_HOUSE/egs++/egs_base_geometry.h
+++ b/HEN_HOUSE/egs++/egs_base_geometry.h
@@ -428,7 +428,7 @@ public:
 
     EGS_Float getMediumRho(int ind) const;
 
-    void setApplication(EGS_Application *app);
+    virtual void setApplication(EGS_Application *app);
 
     /*! \brief Does this geometry object have a B field scaling feature?
      */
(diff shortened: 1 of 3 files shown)


ftessier:
Rebased on `develop` and tweaked the commit message.

ftessier (approved):

ftessier (approved):
Rebased on develop and tweaked the commit message.

mainegra (approved):

blakewalters (approved):

ftessier:
Rebased on develop.

PR #648, opened 2020-12-01
Fix pegs4 compilation errors for gcc > 10.0

mainegra:
Since GNU gcc version 10.0 and up, `pegs4` compilation fails complaining about interface mismatch when using dummy arguments for functions. Previous versions (8.x and 9.x) used to report this only as warnings.

@joeydumont is credited with reporting this error and pointing to a solution based on `INTERFACE` blocks. This wasn't working due to Mortran interpreting `::` as empty labels.

The solution to the label issue is to use compiler directives `%M` and `%F` to switch between Fortran and Mortran code. Rather than using full INTERFACE blocks, the function type is included in the external statement for these functions. Subroutines are not affected.

A much simpler fix is to just add a type declaration statement for the dummy function arguments. Turns out @blakewalters  had already fixed this in his _pegs-less_ implementation avoiding Fortran 90 syntax which could compromise backwards compatibility with very, very old systems ... ;-)

Tested that `pegs4` data for some materials are identical to when using previous pegs4 versions. Compiled successfully with gcc 4.8.5, 9.3.0, and 10.2.0.

Added `clean` target to `Makefile`.

### DIFF
--- a/HEN_HOUSE/pegs4/Makefile
+++ b/HEN_HOUSE/pegs4/Makefile
@@ -41,3 +41,7 @@ $(PEGS4_EXE): $(FORTRAN_FILE)
 $(FORTRAN_FILE): $(sources)
 	@echo "Generating $(FORTRAN_FILE) from pegs4.mortran"
 	@$(MORTRAN_EXE) -d $(MORTRAN_DATA) -f $(sources) -o7 $@ -o8 pegs4_$(my_machine).mortlst
+
+clean:
+	@$(REMOVE) $(FORTRAN_FILE) $(HEN_HOUSE)/bin/$(my_machine)/pegs4.exe
+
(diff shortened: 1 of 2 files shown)


joeydumont:
Cool! I had actually started reading the MORTRAN source to see if it was possible to disable the parsing of `:{LABEL}:` in certain circumstances, but I still hadn't stumbled upon the `%M` and `%F` gates. 

Good work! :+1:

ftessier (approved):

rtownson (approved):

blakewalters (approved):

ftessier (approved):
Adjusted commit messages.

ftessier (approved):
Rebased on develop.

PR #662, opened 2021-01-12
Fix phsp output with photon splitting

blakewalters:
Photons were not included in the phase space data.  This is because $AUSCALL($TRANAUSA) (call to ausgab after particle step) was not included in the $SELECT-PHOTON-MFP macro that handles splitting in dosxyznrc.


ftessier:
Rebase on develop and disentangle from an earlier rebase on the now merged `egs-spectrum` branch.

mainegra (approved):
Wow! We had this bug for so long?

rtownson (approved):

ftessier (approved):

Issue #665, opened 2021-01-25
Bug in g when using multiple energies


mainegra:
When using multiple energies, the accuracy required is systematically reduced from one energy to the next. High precision calculations with several energies will never finish because the desired precision becomes virtually zero.
Reported by @dworogers

Issue #495, opened 2018-11-17
Subroutine egs_shellwise_photo wants to convert photon to an electron


ojalaj:
As reported in https://plus.google.com/117012414775310191025/posts/AJAj8sMBHs4, we are running an egs_chamber simulation on our cluster with 'develop' branch from September 25th (sorry, git log -1 returned fatal: Not a git repository (or any parent up to mount point /home) Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set)). It is a batch run simulating a dose profile with a realistic 'moving' detector. We get this kind of warning/error, which is new for us:
```
IAEA_PhspSource::setSimulationChunk(): illegal attempt to set the simulation chunk between 400000001 and 600000000 ignored

Running 200000000 histories

Batch CPU time Result Uncertainty(%)

1 Subroutine egs_shellwise_photo called with E = 1.0000000253235597E-003 which is below the current min. energy of 1.00000005E-03 keV!
Converting now this photon to an electron,
but you should check your code!
24836.59 2.78508e-22 5.00
2 Subroutine egs_shellwise_photo called with E = 1.0000000464604173E-003 which is below the current min. energy of 1.00000005E-03 keV!
Converting now this photon to an electron,
but you should check your code!
```

When generating our iaea_phsp_source, we used ECUT=ae=0.521 and PCUT=ap=0.01 and for this run we are using ECUT=0.521=ae and PCUT=ap=0.001, with:
```
 Global ECUT= 0.521
 Global PCUT= 0.001
 Global SMAX= 1e10
 ESTEPE= 0.25
 XImax= 0.5
 Skin depth for BCA= 3
 Boundary crossing algorithm= EXACT
 Electron-step algorithm= PRESTA-II
 Spin effects= on
 Brems angular sampling= KM
 Brems cross sections= NRC
 Photon cross sections= mcdf-xcom
 Electron Impact Ionization= On
 Triplet production= On
 Radiative Compton corrections= On
 Bound Compton scattering= On
 Pair angular sampling= KM
 Pair cross sections= NRC
 Photoelectron angular sampling= On
 Rayleigh scattering= On
 Atomic relaxations= On
 Photonuclear attenuation= On
 Compton cross sections = default
 Photonuclear cross sections = default
```
Have you any tips, where to start trying to find the reason/solution for this?

mainegra:
[...]

PR #666, opened 2021-01-25
Fix bug in precision goal in g and 1 keV round-off

mainegra:
This pull request addresses a bug in `g.mortran` in a type 1 calculation for multiple energies whereby the precision goal is reduced at every energy step potentially making the calculation last very long. Bug reported by @dworogers. 

Two other fixes are included in this pull request. See the list below for more information:

- For `calculation type 1` set desired uncertainty `accu0` at the first step of the energy loop. The value of `accu`
  must be updated for each energy since it is different for the `mutr` and `1-g` calculation blocks. Fixes issue #665.

- Round-off errors produce wrong results for a 1 keV energy value. The quick fix is to add an epsilon equal to or larger than 1.0D-18 MeV. Using here a slightly larger epsilon of 1.0D-10 MeV to account for machine precision differences. This should still be negligible in the keV energy range:

```
  REPLACE {$E_THRESHOLD} WITH {1.0D-3}; "Minimum energy in EGSnrc"
  REPLACE {$E_EPS}       WITH {1.0D-10};"Shifted by small epsilon"
```

- Finally set `$RELAX-CUTOFF` in `egsnrc.macros` to 0.001D0 to make it a double precision constant fixing the annoying warning messages from `subroutine egs_shellwise_photo`. Fixes issue #495.

### DIFF
--- a/HEN_HOUSE/src/egsnrc.macros
+++ b/HEN_HOUSE/src/egsnrc.macros
@@ -771,7 +771,7 @@ X-OPTIONS/;
 " use the a lower threshold.
 "***************************************************************************"
 "============================================================"
-REPLACE {$RELAX-CUTOFF} WITH {0.001"threshold energy for outer shells"}
+REPLACE {$RELAX-CUTOFF} WITH {0.001D0"threshold energy for outer shells"}
 REPLACE {$MXPESHELL} WITH {16} "K,L1..L3,M1..M5,N1..N7 + outer shell"
 REPLACE {$MXNE} WITH {500}     "number of energy points per shell "
 "============================================================"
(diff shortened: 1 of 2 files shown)


mainegra:
@ojalaj you might be interested in this pull request!

ojalaj:
@mainegra Thanks :) ! Now I don't need to (remember) change it manually everytime upgrading.

mainegra:
@ftessier , @rtownson , and @blakewalters could we review this one soon? It is really only small and I don't want users running g stuck for ever due to the bug!

rtownson (approved):

blakewalters (approved):

ftessier (approved):
Fixed commit message typos and reflowed paragraphs.

ftessier:
Removed EOL whitespace in g.mortran.

PR #672, opened 2021-03-03
Change radionuclide source to use position/direction sampling from any other source

MartinMartinov:
The implementation of egs_radionuclude_source is changed so
that it takes an input 'base source' instead of defining a
collimated or isotropic source.  This way you can use any
of the egs sources or a custom egs source to define the
location of the radionuclide in the simulation.  The change
in particular was made towards fast simulations targeted
radionuclide therapy across a whole patient phantom.

### DIFF
--- a/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.cpp
+++ b/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.cpp
@@ -23,7 +23,7 @@
 #
 #  Author:          Reid Townson, 2016
 #
-#  Contributors:
+#  Contributors:	Martin Martinov, 2021
 #
 ###############################################################################
 */
@@ -41,11 +41,7 @@
 
 EGS_RadionuclideSource::EGS_RadionuclideSource(EGS_Input *input,
         EGS_ObjectFactory *f) : EGS_BaseSource(input,f),
-    sourceType(""), shape(0),
-    geom(0), regions(0), nrs(0), min_theta(0), max_theta(M_PI),
-    min_phi(0), max_phi(2*M_PI), gc(IncludeAll),
-    source_shape(0), target_shape(0), ctry(0), dist(1),
-    q_allowed(0), decays(0), activity(1) {
+    baseSource(0), q_allowed(0), decays(0), activity(1), sCount (0) {
 
     int err;
     vector<int> tmp_q;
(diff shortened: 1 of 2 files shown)


rtownson:
This change makes sense and improves the versatility of the radionuclide source.

I haven't thought about it too hard yet - are you confident about the getFluence() being correct for all source types?

MartinMartinov:
I went through all the algebra on paper when I first did the implementation and I think it should work.  I wasn't super clear on phase space sources, but plugging in the numbers I think it worked out.  I haven't actually performed testing on the fluence outside of comparing to your original source using an isotropic setup, though.

Robarts-Lee-Lab:
After doing some further testing, I found there is an issue in the store/add/set/reset calls when it comes to parallel runs (my previous testing was all in console...), so please don't merge this until I find the issue.

ftessier:
@MartinMartinov I find that I cannot push to the PR branch (located on your fork). Can you [check your settings?](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) Normally, by default the admin of the repo against which you open the PR can push to your branch (yikes, I know!). This is useful to polish up the branch before merging, etc. If your setting is as expected, let me know, I will investigate further...

ftessier:
@Robarts-Lee-Lab any update regarding the store/add/set/reset calls?

MartinMartinov:
image
Is "Allow edits by maintainers" the right option?  It seems to be checked.

Also, I consulted Reid on the store/add/set/reset calls and they should be all good now.  I've done some more robust testing and it looks good.

ftessier:
Yes that is the setting. Thank you for checking, I will try a test push!

ftessier:
Ok, that worked. It must have been my ssh keys before, who knows? I will rebase on the current `develop` and apply some formatting etc.

ftessier:
Rebased on `develop`.

ftessier:
@MartinMartinov the branch `feature-radionuclidesource-sourceimport` is not in your fork anymore. I can create a new pull request with your proposed changes. But let me know if you would prefer to reinstate `feature-radionuclidesource-sourceimport` at your end and that I continue working on it here.

MartinMartinov:
A new pull request would probably be easier at this point. :)

ftessier:
Reinstated in #896. Closing this pull request since the original branch was removed on the fork.

PR #702, opened 2021-04-21

seirios:
This is a resubmission of https://github.com/nrc-cnrc/EGSnrc/pull/292 since my develop branch was removed.
The original message was: "I noticed a lack of the RANLUX random number generator in egs++, so I implemented a class EGS_Ranlux for it based on GSL's integer implementation (gsl_rng_ranlux), which is pretty much the same as the original Fortran implementation in EGSnrc. I haven't done extensive testing, but it seems to be working fine."

### DIFF
--- a/HEN_HOUSE/egs++/egs_rndm.cpp
+++ b/HEN_HOUSE/egs++/egs_rndm.cpp
@@ -249,7 +249,7 @@ void EGS_Ranmar::setState(EGS_RandomGenerator *r) {
     copyBaseState(*r);
     EGS_Ranmar *r1 = dynamic_cast<EGS_Ranmar *>(r);
     if (!r1) {
-        egsFatal("EGS_Ranmar::setState: attampt to set my state by a non EGS_Ranmar RNG!\n");
+        egsFatal("EGS_Ranmar::setState: attempt to set my state by a non EGS_Ranmar RNG!\n");
     }
     ix = r1->ix;
     jx = r1->jx;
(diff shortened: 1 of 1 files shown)


ftessier:
Thank you @serios! 🙏🏻

ftessier:
Will look into the PR again. May I push to your `develop` branch (on EGSnrc-1) if I need to make changes?

seirios:
Sure! I opened that repo specifically for this. No problem, I'm happy to contribute to this great piece of software!

ftessier:
@serios thanks for the invitation on the repo, but I will do it from here, i.e., push to the pull request branch you just re-created. By default, admin of the repo against which you create a PR can push to the PR branch on your repo. I was just asking if that was ok, as it might come across as a little invasive if you are not expecting it. 😄

seirios:
OK, no problem, I wasn't aware of that :)

mxxo:
I wanted to note that C++11 added RANLUX to the standard library: https://en.cppreference.com/w/cpp/header/random (`ranlux24`, `ranlux48`). It might be worth comparing the performance of both implementations, perhaps the standard has some tricks up its sleeve.

HEN_HOUSE/egs++/egs_rndm.cpp
```diff
@@ -404,6 +404,230 @@ void EGS_Ranmar::fillArray(int n, EGS_Float *array) {
...
+
+    for (j=0; j<24; j++) {
+        k = seed / 53668;
+        seed = 40014 * (seed - k * 53668) - k * 12211;
+        if( seed < 0 ) seed += 2147483563;
+        u[j] = seed % twop24;
+    }
+
+    i = 23;
+    j = 9;
```
mxxo, line 583: There appears to be a variable shadowing bug here: The local `j` is set to 9, not `EGS_Ranlux::j` as likely intended (AddressSanitizer complains about an out of bounds read for `u[j]` in `incrState` later on with the current version). Changing the local `j` to `jj` seems to run OK.

mxxo (commented):
There appears to be a small issue in the implementation. I think the overall structure looks good but more testing and QA is needed. Comparing performance to the Fortran version of `ranlux` (the C++ version of `ranmar`, etc.) would also be nice to have.

ftessier:
I think we ought to provide hooks in egs++ to call on the `<random>` generators provided as part of STL (since C++11). I don't look favorably upon maintaining our own generator code inside EGSnrc.

In cursory testing, I found that the ranlux24 generator is ~20 times slower than the default ranmar. the Mersenne Twister about twice as slow as ranmar. The latter is thus a good alternative (and perhaps a good default), while ranlux would only be practical for investigating the quality of the faster generators.

For my own reference, here is a boiler plate for the C++11 STL random:

```c++
#include <random>
#include <iostream>

int main() {
    auto seed = 12345;
    std::mt19937 mt {seed};
    std::uniform_real_distribution<double> uniform(0.0, 1.0);

    for (int i=0; i<10; i++) {
        std::cout << uniform(mt) << "\n";
    }
}
```

 I also remain curious about xoshiro and pcg64 generators to boost simulation speed.

PR #637, opened 2020-09-23
Add egs_mesh msh file parser

mxxo:
Add a basic Gmsh msh file parser for the `egs_mesh` class. The idea is to avoid requiring the `libgmsh` dependency to use `egs_mesh`. This was spun off from the original `egs_mesh` PR (#556). 

The single public API function of the msh_parser library is `EGS_Mesh parse_msh_file(std::istream& input)`, which parses an `EGS_Mesh` object from an input stream. In future PRs we will move `EGS_Mesh` to its own header and make it a subclass of `EGS_BaseGeometry` to actually run simulations. 

The parser can handle `msh` version 4.1, the latest stable format version. The format [roadmap](http://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format) does not predict any changes to the sections we need. The sections `egs_mesh` uses are: 
* `$MeshFormat`: msh file format information (only v4.1 is supported)
* `$PhysicalNames`: Arbitrary group information, we are using it for EGSnrc medium names
* `$Entities`: 3D volume information and their associated `PhysicalGroups`
* `$Elements`: tetrahedrons, the volumes they belong to and constituent nodes 
* `$Nodes`: node coordinates

There is a new parser namespace `msh_parser` and a test suite file (`msh-parser-tests.cpp`).

PR #703, opened 2021-04-22
Update compiler flags to require c++14 standard

ftessier:
The `egs_view` gui already specifies c++11, and the new mesh geometry also requires c++11. These changes are extracted from #637 to create a distinct commit, and updated to c++14 while we are at it.

### DIFF
--- a/HEN_HOUSE/egs++/view/view.pro
+++ b/HEN_HOUSE/egs++/view/view.pro
@@ -93,7 +93,7 @@ unix {
 #QMAKE_CXXFLAGS+="-ggdb3"
 #QMAKE_LFLAGS+="-fsanitize=address"
 
-QMAKE_CXXFLAGS+=-std=c++11
+QMAKE_CXXFLAGS+=-std=c++14
 UI_DIR = .ui/$$my_machine
 MOC_DIR = .moc/$$my_machine
 OBJECTS_DIR = .obj/$$my_machine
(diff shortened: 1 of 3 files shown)


rtownson:
Note that on older compilers (like the one on my linux box), c++14 won't work. They would have to change it to c++1y. This is probably fairly rare but we could add a check.

ftessier:
I would stick to official releases of the standard (if I understand correctly `c++1y` referred to the draft of `c++14`). What kind of error message do you get on your old box? Is there any clue to the user that this flag is causing the problem?

rtownson:
The error is pretty clear:
```
g++: error: unrecognized command line option ‘-std=c++14’
make: *** [dso/test/egs_input.o] Error 1
```

ftessier:
Ok, then I propose to leave it as is and move on. I mean, the default option is presented during the configuration (script or gui), so after an unsuccessful run one could recognize this in the configuration and adjust. I will add a note on the installation wiki.

ftessier:
Added the folllowing paragraph on the wiki pages of installation instructions, in the **Compiler** section:

> Note that by default, some parts of EGSnrc require a compiler that supports standard C++11 (g++ option -std=c++11), and will soon require C++14 (g++ option -std=c++14). If you have an older compiler that does not recognize this option, please adjust or remove it from the list of proposed flags during the EGSnrc configuration.

mxxo:
I just wanted to mention, if for any reason we can't jump right to C++14, that is not a blocker for me. I have been compiling with `-std=c++11` and so haven't been using any C++14 features yet.

ftessier:
Thanks for the note, I still think we ought to move one to C++14 as soon as possible, and then retrograde before release 2022 if anything falls apart...

rtownson (approved):

blakewalters (approved):

mainegra (approved):

PR #705, opened 2021-04-23
Fix #573: wrong I-value for aluminum alloys

### LINKED ISSUE #573: Ivalue, aluminium alloy
In the file:
EGSnrc/HEN_HOUSE/pegs4/density_corrections/compounds/aluminumalloy_type6061.density

The I-value is 188.02901 eV while is should be something close to pure aluminum I-value of 166 eV.
NIST ESTAR program issue

ftessier:
Recalculate the aluminum alloy density corrections files with `estar.f` program, after it was pointed out in #573 that the I-value in those files was far from the pure aluminum value of 166 eV (ICRU Report 90), although the alloys typically contain only trace amounts of impurities. This also affects `aluminumoxide.density`, so I will recompute that as well. Other compounds containing `Al` are also likely affected, but the highest concentration elsewhere is 6%. I will check if the density correction data is affected.


mainegra (approved):
Surprising how many materials were affected! Should we look into a co-op project to integrate this to EGSnrc?

ftessier:
Yes, I am not sure what to do with this... I recomputed some materials, and the I-values are different. For example, concrete type 3 goes from 166.479 eV to 151.4 eV, with a significant effect on density corrections. Since I am not certain about the source of this change, I will **only update the aluminum alloys here** (which are "clearly" wrong), and leave out aluminum oxide or other compounds. The issue with the set of  the compounds from ICRU37 is that in some cases there might be I-values that are determined independently from other sources (?), and a diligent review of the ICRU reports is required, etc.

I will close #573 with this pull request, and open a new issue for a broad review of density effect correction data.

ftessier:
Removed the aluminum oxide change for now.

rtownson (approved):

blakewalters (approved):

ftessier:
> For example, concrete type 3 goes from 166.479 eV to 151.4 eV, with a significant effect on density corrections.

@rtownson can you double-check aluminum alloys and other aluminum containing compounds in the comparison with the new C++ version of estar? I would expect these would have shown large discrepancies in the test of all compounds.

rtownson:
> @rtownson can you double-check aluminum alloys and other aluminum containing compounds in the comparison with the new C++ version of estar? I would expect these at least would have shown large discrepancies in the test of all compounds, but they have not?

In the comparisons so far I used the I-value from the density correction file, essentially assuming that the chosen I-value was intentional. But I'll do up a test to compare with the default I-value.

rtownson:
When I ran the test it seems there are a lot of materials where the I-value in the file doesn't match the "default" as calculated by ESTAR. I think a difference of 0.3 in the density correction is still pretty small - was this difference actually detectable in simulations?

```
Material                        Max absolute difference
aluminumalloy_type1100.density	0.266
aluminumalloy_type2014.density	0.269
aluminumalloy_type6061.density	0.264
aluminumoxide.density           0.183
```

ftessier:
Ok, thanks for checking. We ought to review the I-values calculated by estar against the ones currently in all density files, and flag the ones that differ significantly and why.

Issue #681, opened 2021-03-26
EGS_XYZGeometry documentation: x/y/z-planes vs. x/y/z-slabs?


ojalaj:
In https://nrc-cnrc.github.io/EGSnrc/doc/pirs898/classEGS__XYZGeometry.html it suggests that:

The other new possibility to define a XYZ geometry is

```
:start geometry:
    library = egs_ndgeometry
    type = EGS_XYZGeometry
    x-slabs = Xo  Dx  Nx
    y-slabs = Yo  Dy  Ny
    z-slabs = Zo  Dz  Nz
:stop geometry:
```

which should be self explanatory.

However, in a simple simulation, using above with x/y/z-slabs resulted in very long simulation times (e.g. 44-fold) compared to x/y/z-planes, as in other examples. Why is this happening and which one is right or both?

ftessier:
Hi @ojalaj, that is indeed quite strange. Are you sure that your two inputs define the same geometry? Can you share a small sample input that reproduces the problem?

ftessier:
As far as I can tell the implementation is essentially the same, except for the constructor variant:
```c++
new EGS_PlanesX(xslab[0],xslab[1],nx,"",EGS_XProjector("x-planes"))   // for x-slabs
new EGS_PlanesX(xpos,"",EGS_XProjector("x-planes")                    // for x-planes
```

ftessier:
Implementation of the slab constructor seems fine to me:
```c++
    EGS_PlanesT(EGS_Float xo, EGS_Float dx, int np, const string &Name,
                const T &A) : EGS_BaseGeometry(Name), a(A) {
        if (np < 1) egsFatal("...",np);
        if (dx <= 0) egsFatal("...",dx);
        p = new EGS_Float [np+1];
        p[0] = xo;
        for (int j=0; j<np; j++) {
            p[j+1] = p[j] + dx;
        }
        nreg = np;
        p_last = p[nreg];
        n_plane = np;
        is_uniform = true;
        dp = dx;
    };
```
If you have a simple example input on hand, I will take a look at it.

ftessier:
Note by self-explanatory, what the author meant was "`Xo` is the position of the first x plane, `Dx` is the uniform slab thickness, and `Nx` is the number of slabs". Perhaps you interpreted "self-explanatory" differently? 😉

ojalaj:
Hi @ftessier ! Actually, the reason is that! Sorry about this - I got confused with the notations :(

[...]

PR #707, opened 2021-04-23
Change endl to \n in egs++ scoring to improve performance

rtownson:
Change a few occurances of `endl` to `\n` in egs++ mid-simulation scoring routines, to increase performance. The reason for this is that `endl` performs a buffer flush each time, which I believe is not necessary. The takeaway here is to never use `<< endl` unless you actually want the flush.

The performance hit was huge for large arrays, like a 300x300x300 3ddose file. On linux this was a 50% speedup on the total simulation time, on windows 400%. For smaller scoring arrays (i.e. most cases) the speedup will be minimal.

There is an outstanding problem that even with these changes, egs++ simulations with large 3ddose files are many times slower than dosxyznrc (2-30x slower), entirely due to the storeState function. This might be fixed if we remove ostream and use fprintf instead, but that would probably require updating the storeState functions across all of egs++. Or if we changed the egsdat file to be binary... This is under discussion here: https://github.com/nrc-cnrc/EGSnrc/discussions/688

Note that a partial work-around is to use `nbatch=1`.

Thanks to @ojalaj for reporting this!

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
@@ -421,7 +421,7 @@ void EGS_DoseScoring::outputDoseFile(const EGS_Float &normD) {
         int ny=dose_geom->getNRegDir(1);
         int nz=dose_geom->getNRegDir(2);
         //output no. of voxels in x,y,z
-        df_out << nx << " " << ny << " " << nz << endl;
+        df_out << nx << " " << ny << " " << nz << "\n";
         //use single precision real for output
         float bound, dose, doseun;
         //output voxel boundaries
@@ -429,17 +429,17 @@ void EGS_DoseScoring::outputDoseFile(const EGS_Float &normD) {
             bound=dose_geom->getBound(0,i);
             df_out << bound << " ";
         }
-        df_out << endl;
+        df_out << "\n";
         for (int j=0; j<=ny; j++) {
             bound=dose_geom->getBound(1,j);
             df_out << bound << " ";
         }
-        df_out << endl;
+        df_out << "\n";
         for (int k=0; k<=nz; k++) {
             bound=dose_geom->getBound(2,k);
             df_out << bound << " ";
         }
-        df_out << endl;
+        df_out << "\n";
         //divide dose by mass and output
         for (int i=0; i<nx*ny*nz; i++) {
             doseF->currentResult(i,r,dr);
@@ -447,7 +447,7 @@ void EGS_DoseScoring::outputDoseFile(const EGS_Float &normD) {
             dose=r*normD/mass;
             df_out << dose << " ";
         }
-        df_out << endl;
+        df_out << "\n";
         //output uncertainties
         for (int i=0; i<nx*ny*nz; i++) {
             doseF->currentResult(i,r,dr);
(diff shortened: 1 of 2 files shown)


MartinMartinov:
That's great news!  I didn't realize there were so many endl calls.  I think the only place they are important is in debug output, to make sure the buffer is flushed before a segmentation fault occurs.

ftessier:
Good find! 🏅  This `endl` issue has been a real curse on the modern world; it is probably responsible for a lot of wasted time in a lot of software. Great that this was caught here. All those C++ books that had it in sample code has burned this in every programmer out there! I myself _still_ sometimes forget to _not_ use it, force of habit. I will scan the EGSnrc code base for it, come to think of it (of course it remains useful when you _do_ want to flush the ostream).

ftessier:
There are a few instances in `egs_dose_scoring` too, might be worth adding them to this pull request.

ftessier:
All other instances seem pretty inconsequential, or necessary.

rtownson:
> There are a few instances in `egs_dose_scoring` too, might be worth adding them to this pull request.

Sure, though I think those ones are pretty safe! The problem was that in `storeState` it puts a new line between every quantity, e.g. every one of the _27 million voxels_ has dose **on its own line**!

ftessier (approved):

ojalaj:
Great that you got this solved @rtownson ! Some background for this case is that I was very surprised, when for a veryvery simplified case, DOSXYZnrc, even without HOWFARLESS, was ~30x faster than egs_chamber (both with nbatch=10). Reducing nbatch=1, as suggested by @rtownson, helped a lot, making DOSXYZnrc "only" ~2x faster. (@ftessier this is related to #681, when I interpreted the documentation incorrectly, but also wondered why some simulations took soooo long)

mainegra (approved):

rtownson:
I've added another commit that takes the improved speed of this PR from 1.5x to 2.1x by using `FILE*` instead of `ofstream` (on linux). It's still 16x slower than DOSXYZnrc when using 10 batches, entirely due to `storeState`.

I also confirmed that the results are identical.

blakewalters (approved):

HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
```diff
@@ -469,20 +469,41 @@ void EGS_DoseScoring::outputDoseFile(const EGS_Float &normD) {
...
+    // Convert the ostream to FILE * before writing data
+    // This might seem like a hack, but improves efficiency for writing
+    // very large arrays (like 3ddose) to egsdat
+    // It's otherwise fine (or better) to use ostream elsewhere
+
+    // Get the position from the stream
+    long pos = data.tellp();
+    data.flush();
+
+    string ofile = app->constructIOFileName(".egsdat",true);
```
ftessier, line 481: While I appreciate the gain in performance in writing this via a separate `FILE *`,  it makes the code fragile here. There is no guarantee that the hard-coded filename (which cannot be recovered from the stream) will match that of the stream, and the code relies on perfect synchronization between the two separate file handles, which is not guaranteed and  even OS or compiler dependent.

I reckon the only place where the ostream proves really slow for large arrays is the repeated calls inside the scoreSingle method. I think we can improve it _there_, by buffering (or double-buffering) the write loop in the scoring array storeState method and realtively few `ostream::write` calls.

I propose to merge in only the innocuous `<< endl;` changes at this time, and keep the proposed `FILE *` on this branch for further testing along the lines above.

ftessier (approved):

ftessier:
Merged in the `<<endl;` changes, and saved the `FILE *` changes on the same branch name `egspp-scoring-speed`.

PR #709, opened 2021-04-23
Increase floating point-precision for rho output in dose scoring object

### LINKED ISSUE #708: Density output in egslog file for egs_chamber and cavity
The number of digits reported in the .egslog file for the density of each medium is only 2, therefore, the reported density of air is 0.00.

ftessier:
This resolves #708. At the same time I have realigned the output slightly. While not perfect in regards to all cases, I find it provides a better default alignment. Sample output:

```
==> Summary of region dosimetry (per particle)

ir  medium  rho (g/cm^3)  Volume (cm^3)   Edep (MeV)                   D (Gy)                   
------------------------------------------------------------------------------------------------
 0  air      0.00120479  270000.000000    1.7576e-02  +/-  0.697  %    8.6558e-15  +/-  0.697  %
 1  water    1.00000000     100.000000    1.9265e+00  +/-  0.262  %    3.0862e-12  +/-  0.262  %
 2  air      0.00120479     900.000000    6.9888e-02  +/-  0.496  %    1.0326e-11  +/-  0.496  %
------------------------------------------------------------------------------------------------
```

I am growing weary of the rather inconsistent output format across EGSnrc (including this commit, for example!). Should we not start using proper table output, at least on the egs++ side. The header-only [tabulate library ](https://github.com/p-ranav/tabulate) comes to mind. There must be a number of solutions out there, but surprisingly I cannot find much! Could code our own little output stream class? Suggestions welcome.

The `git column` command might provide some inspiration.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
@@ -135,6 +135,7 @@ void EGS_DoseScoring::setApplication(EGS_Application *App) {
     char buf[32];
     int count = 0;
     int imed=0;
+    max_medl = 6;  // length of the string "medium" is the minimum
     for (imed=0; imed < nmedia; imed++) {
         sprintf(buf,"%s%n",app->getMediumName(imed),&count);
         if (count > max_medl) {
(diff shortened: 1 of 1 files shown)


ftessier:
This also increases the precision of rho higher up in the egslog file, where the dose scoring object outputs its description.

mainegra (approved):

rtownson (approved):

blakewalters (approved):

PR #713, opened 2021-04-29
Increase the list of colours for the BEAMnrc gui

rtownson:
Change the default colour list for the BEAMnrc gui to include 269 colours instead of 44. They are from a list of maximally distinct colours. Vacuum still defaults the grey, and black for anything beyond the 269. 

The list I found here:
https://graphicdesign.stackexchange.com/revisions/3815/8


mainegra (approved):
Neat!

ftessier:
I recall looking into the problem of generating maximally "distant" colours algorithmically when I was adding the region colour swatches in `egs_view`. Not an easy problem! Your link led me to a [discussion of the solution](http://godsnotwheregodsnot.blogspot.com/2012/09/color-distribution-methodology.html), neat! A colour table is probably the way to go; once you have the algorithm, you might as well just generate and include the table :-).

rtownson:
I should dig up the silly algorithm I wrote for "EGSweb" and see how it compares :)

blakewalters (approved):

ftessier:
@rtownson there are only 269 colors in the list (including the last one black). Did you mean to include `#FFFFFF` white?

rtownson:
I must have just miscounted - I did not mean to intend white and the original list also has 269. I'll fix the typo.

ftessier:
Squashed and rebased on develop.

PR #716, opened 2021-05-14
Fix egs-parallel job name with invalid leading character bug

rtownson:
Fix a mistake in the previous fix for non-alphanumeric characters in the job name in egs-parallel jobs resulting in a crash. The previous fix removed non-alphanum characters from the start of the original job name, but actually we needed to remove from the start of the trimmed 14 character job name.

### DIFF
--- a/HEN_HOUSE/scripts/egs-parallel-pbs
+++ b/HEN_HOUSE/scripts/egs-parallel-pbs
@@ -123,6 +123,7 @@ jobname=$(echo "${basename}[$nthread]" | sed 's/^[^[:alnum:]]*//')
 trim=$(( $(echo $jobname | wc -c) - 14 ))
 if [ $trim -gt 0 ]; then
     jobname=$(echo $jobname | cut -c $trim-)
+    jobname=$(echo $jobname | sed 's/^[^[:alnum:]]*//')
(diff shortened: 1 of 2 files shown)


ftessier:
Good find, I had forgotten that the job name was trimmed from the left!

ftessier:
But it seems it would have to be done in both places, because now the non-alphanumeric characters are only removed if `trim > 0`.

HEN_HOUSE/scripts/egs-parallel-pbs
```diff
@@ -119,10 +119,11 @@ verbosity=$8
 exec 3>>$basename.egsparallel
 
 ### set scheduler job name (skip leading non alnum chars, maximum 14 characters)
-jobname=$(echo "${basename}[$nthread]" | sed 's/^[^[:alnum:]]*//')
```
ftessier, line 122: Leave it here too, in case `trim <= 0`?

rtownson: Sure, fair enough!

mainegra (approved):

blakewalters (approved):

PR #717, opened 2021-05-16
Fix `ReadCT_DICOM()` to work with slices of arbitrary size

mpayrits:
As outlined in issue #718, `ReadCT_DICOM()` currently works only with slices of size 512-by-512. This is a quick fix to make it work with smaller sizes as well.


rtownson (approved):

mainegra (approved):

ftessier (approved):

ftessier:
Added contributor name in header.

PR #719, opened 2021-05-17
Fix egs++ 3ddose output

### LINKED ISSUE #720: egs++ 3ddose output is _still_ wrong
Dose values in `.3ddose` files produced by `EGS_DoseScoring` over an `EGS_XYZGeometry` defined through an `.egsphant` file, density matrix, or interfile are wrong. They are off by a voxel-dependent factor equal to the default density of the medium in that voxel. This is contrary to the claims in pull request #645 which only fixed the issue for `EGS_XYZGeometry`'s explicitly defined in the input file through slabs or planes.

I'm [attaching](https://github.com/nrc-cnrc/EGSnrc/files/6493383/3ddose_density_bug.zip) a few simple input files that demonstrate the issue. `germanium_copper_cube.egsphant` is a simple handcrafted `.egsphant` file representing a cube split into two equal-volume halves along the z-direction. The lower-z half consists of copper, CU521ICRU, and the upper half of germanium, GE521ICRU. These materials were chosen because they're both in the default `$HEN_HOUSE/pegs4/data/521icru.pegs4dat` PEGS4 file, they're both elements, and their default densities are `8.9333 g/cm3` and `5.3600 g/cm3`, respectively, different enough between themselves and both different enough from `1.0 g/cm3` to be able to clearly demonstrate the bug. For actual voxel densities I've chosen densities 9.0 and 5.3, close but not exactly equal to the default densities, which is the general case in simulations. I've also attached the ramp file needed by egs++ and `.egsinp` files for egs++ and dosxyznrc simulations of a 1 MeV isotropic point source in the center of the cube with a million primary histories. I also include results I've obtained on my machine in the `results/` subfolder.

Suppose you've followed the egs++ sections of the [getting started guide](https://nrc-cnrc.github.io/EGSnrc/doc/getting-started.pdf) and have compiled the `myapp` application whose source lives in `$EGS_HOME/myapp`. Then do the following to observe the bug:
* Copy `germanium_copper_cube.egsphant` to `$EGS_HOME/myapp` and `$EGS_HOME/dosxyznrc`
* Copy `germanium_copper_cube_egspp.egsinp` and `germanium_copper_ramp.dat` to `$EGS_HOME/myapp`
* Copy `germanium_copper_cube_dosxyznrc.egsinp` to `$EGS_HOME/dosxyznrc`
* Run the dosxyznrc simulation:
```
cd $EGS_HOME/dosxyznrc
dosxyznrc -p 521icru -i germanium_copper_cube_dosxyznrc.egsinp
```
* Run the egs++ simulation, noting the final dose output on the console:
```
cd $EGS_HOME/myapp
myapp -p 521icru -i germanium_copper_cube_egspp.egsinp
```

Observe that the doses reported in `$EGS_HOME/dosxyznrc/germanium_copper_cube_dosxyznrc.3ddose` and `$EGS_HOME/dosxyznrc/germanium_copper_cube_dosxyznrc.egslst` are the same. Observe that the doses output on the console by `myapp` are also _approximately_ equal to that. The doses in `$EGS_HOME/myapp/germanium_copper_cube_egspp.3ddose`, however, are considerably larger. The ratio of the `myapp` and `dosxyznrc` `.3ddose` doses is approximately the density of copper in `g/cm3` for the first voxel and approximately the density of germanium in `g/cm3` for the second voxel.

The issue lies in `EGS_XYZGeometry::getMass()` returning the volume times relative density, i.e., the absolute density divided by the default density of the medium, rather than just the absolute density. I've submitted pull request #719 that deals with this.

I'm however still unsure what the intended purpose of `getMass()` is. Its name implies regular mass, but the Doxygen above its declaration in `EGS_BaseGeometry` proudly declares that it's the volume times the relative density. If you ask me, that's not a sensible quantity and even has units of volume. So, what is the intended purpose? In the pull request I've taken the liberty of updating the Doxygen to state it returns the (regular) mass.

The documentation is also not very helpful on what the default density, the third number on each line of the `ct ramp` file, actually is. A user might be particularly mystified by it because, from their perspective, egs++ should already know the default density of a medium. My impression is that having to re-state the default density of a material is a workaround for the fact that the PEGS data is not yet "hatched" when calling `EGS_Application::initGeometry()`. It is certainly not a very user-friendly workaround and offers ample opportunity for confusion and error when collecting and copying the densities. It might be a good idea to add a post-hatch hook (i.e., after the `initCrossSections()` call in `EGS_Application::initSimulation()`) and set media-dependent things there down the line. For starters, I think explaining in Doxygen why default densities are needed and why they can't just be read from PEGS data would greatly clarify what they actually are. I've updated the Doxygen in the pull request.

There's also the smaller issue of the dose reported on the console not taking into account density scaling. And right now it can't do so straightforwardly - `EGS_Application` defines many functions that delegate to its `geometry` member, such as `getnMedia()` and `getMediumName()`, but `hasRhoScaling()` is not one of them. Either that and `getRelativeRho()` would have to be added to `EGS_Application` or, and I'm more in favour of this, it would have to be possible to retrieve its `geometry` member through a new `EGS_BaseGeometry *getGeometry() const;`.

mpayrits:
This pull request addresses the problems outlined in issue #720. `EGS_XYZGeometry::getMass()` now returns the actual mass of the voxel. The description of the default density in the Doxygen has been extended. I have changed the Doxygen of `EGS_BaseGeometry::getMass()` to state that it returns actual mass. But I did not fix the description saying that it is only implemented in `EGS_XYZGeometry`. It seems to also be implemented in at least `EGS_cSpheres` and `EGS_cSphericalShell`, but I'm not familiar enough with those and whether they're considered stable.
There's also the issue of vacuum voxels. In the previous implementation, encountering a vacuum voxel, with a medium index of -1, resulted in at least two kinds of undefined behaviour. The first is accessing `the_media->rho[-1]` in `EGS_AdvancedApplication::getMediumRho()`. In practice this probably mostly points into zeroed out memory and hence returns 0.0. This then causes float division by zero in several places in `EGS_DoseScoring`. I have addressed the first UB instance by adding a check to `EGS_AdvancedApplication::getMediumRho()` and I'll let someone else tackle the instances in `EGS_DoseScoring`.
I would have moved `EGS_XYZGeometry::getMass()` and most other functions' implementations from the header into the source file, but I've left it as it is for now for some sort of consistency.
I have also overridden `getMass()` in `EGS_DeformedXYZ` to return 1 since the volume calculation there should be different, which is outside the scope of this pull request. I see `EGS_DeformedXYZ`, which can't be constructed from an input file, is either a geometry in the making or, and I find this more likely since it hasn't been touched in 5 years, has been abandoned. As such I think you should remove it as it's dead code. And if you ever pick it up again, I don't think deriving from an `EGS_XYZGeometry` is a good idea since it's composed of tetrahedra. Presently it also doesn't override `getType()` so would be accepted for 3ddose output by `EGS_DoseScoring`, possibly leading to strange results.
Fixes #720.

### DIFF
--- a/HEN_HOUSE/egs++/egs_advanced_application.cpp
+++ b/HEN_HOUSE/egs++/egs_advanced_application.cpp
@@ -1171,7 +1171,7 @@ void EGS_AdvancedApplication::resetRNGState() {
 //************************************************************
 // Returns density for medium ind
 EGS_Float EGS_AdvancedApplication::getMediumRho(int ind) {
-    return the_media->rho[ind];
+    return ind < 0 ? 0.0 : the_media->rho[ind];
 }
 // Returns edep
 EGS_Float EGS_AdvancedApplication::getEdep() {
--- a/HEN_HOUSE/egs++/egs_base_geometry.h
+++ b/HEN_HOUSE/egs++/egs_base_geometry.h
@@ -231,7 +231,7 @@ public:
      */
     virtual EGS_Float hownear(int ireg, const EGS_Vector &x) = 0;
 
-    /*! \brief Calculates the volume*relative rho (rhor) of region ireg.
+    /*! \brief Calculates the mass of region ireg.
 
       Currently only implemented in EGS_XYZGeometry
     */
--- a/HEN_HOUSE/egs++/geometry/egs_nd_geometry/egs_nd_geometry.h
+++ b/HEN_HOUSE/egs++/geometry/egs_nd_geometry/egs_nd_geometry.h
@@ -28,6 +28,7 @@
 #                   Frederic Tessier
 #                   Reid Townson
 #                   Randle Taylor
+#                   Matjaz Payrits
 #
 ###############################################################################
 */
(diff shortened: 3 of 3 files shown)


rtownson:
Thank you for this thorough and thoughtful contribution! You rightly point out that there is a lot of work to be done in terms of improving the design, consistency and functionality of egs++. I'm not sure what the plan for `EGS_DeformedXYZ` is, I'll let others comment on that.

rtownson (approved):
I'll address the outstanding issues with `egs_spheres` and `egs_dose_scoring` in a separate PR.

blakewalters (approved):
Looks solid to me!

rtownson:
PR #759 implements most but not all of these fixes, along with some changes. We'll have to merge these.

rtownson:
Superceded by #759.

PR #721, opened 2021-05-19
Add the dirt density file used by icrm-hpge-example

rtownson:
Add the dirt density correction file that was used in the icrm-hpge MC comparison. Also update the materials in the example input file to refer to the correctly named density files.


mainegra (approved):

ftessier:
@rtownson I don't understand why this material is called `dirt`, if it is only `Si` and `O`. I presume it was the name used in the ICRM comparison, but the generic `dirt` seems misleading out of context. Thoughts?

rtownson:
You're right, I'll rename it to silicondioxide.density.

ftessier (approved):

ftessier:
@blakewalters please review when you have a chance.

ftessier:
@blakewalters please review this one when you have a minute.

ftessier:
Rebased on develop.

blakewalters (approved):

ftessier:
Squashed the commits.

PR #725, opened 2021-05-21
added -mcmodel=medium as a default compiler flag

crcrewso:
added -mcmodel=medium as a default compiler flag to default optimization flag lists
This targets issue #723 along with other similar issues

### DIFF
--- a/HEN_HOUSE/scripts/configure
+++ b/HEN_HOUSE/scripts/configure
@@ -343,7 +343,7 @@ for test_arg in -v; do
             junk=$(echo $f_test_output | grep "GNU Fortran")
             if test "x$junk" != x; then
                 f_version=$(echo $junk | grep -o -m1 'GNU Fortran [^\s]* version [0-9.]*'| head -1)
-                f_oflags="-O2 -mtune=native"
+                f_oflags="-O2 -mtune=native -mcmodel=medium"
                 f_dflags="-g"
                 if test ! x$is_x86_64 = x; then
                     f_fflags="-fPIC"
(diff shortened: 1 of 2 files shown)


ftessier:
Great! Note to self. Note to self: update the configure gui accordingly.

crcrewso:
> Great! Note to self. Note to self: update the configure gui accordingly.

Could you add the updated configure (maybe as an alternate branch)? I'd like to get the LLVM (and update PGI/ICC) compilers tested and configured and it just makes sense not to duplicate the efforts.  

After the success with ARM and current indications that we're going to start seeing more [ARM based clusters](https://en.wikipedia.org/wiki/Fugaku_(supercomputer)) we'll need to support those too.

ftessier:
> Could you add the updated configure (maybe as an alternate branch)? I'd like to get the LLVM (and update PGI/ICC) compilers tested and configured and it just makes sense not to duplicate the efforts.

I don't understand what you mean: you want me to create a new branch off `develop` with this PR's changes, for testing? But you can do that locally at your end, so I think I am missing the point...

crcrewso:
> > Could you add the updated configure (maybe as an alternate branch)? I'd like to get the LLVM (and update PGI/ICC) compilers tested and configured and it just makes sense not to duplicate the efforts.
> 
> I don't understand what you mean: you want me to create a new branch off `develop` with this PR's changes, for testing? But you can do that locally at your end, so I think I am missing the point...

Sorry, I really I mean a new branch with Your configure, unless it's a different filename and can live in develop in parallel.

rtownson (approved):

ftessier:
Rebased on develop and added the `-mcmodel=medium` option to the the default flags in the egs_configure gui as well.

PR #732, opened 2021-06-10
Fix getMass and dose scoring density scaling

### LINKED ISSUE #720: egs++ 3ddose output is _still_ wrong

rtownson:
Fix the `getMass` definitions in `egs_spheres` and `egs_rz` to return the regular mass. Also fix divide-by-zero situations in `egs_dose_scoring` for vacuum, and add proper accounting for density scaling. This is an important change for users who use the input `set relative density =` along with `egs_dose_scoring` - the reported dose from the ausgab object was not scaled by the density scaling factor.

The changes to `getMass` do not affect any standard EGSnrc output, but may have impact in other applications such as `egs_brachy` (@randlet @mchamberland) or power-users of `egs_autoenvelope`. These functions were added somewhat recently and are not used anywhere but in `egs_autoenvelope`. Similar, more critical fixes to `egs_nd_geometry` are handled in PR #719.

Fixes #720. Thanks to @mpayrits for outlining these changes.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
@@ -340,7 +340,7 @@ void EGS_DoseScoring::reportResults() {
                     continue;
                 }
                 int imed = app->getMedium(ireg);
-                EGS_Float rho = app->getMediumRho(imed);
+                EGS_Float rho = app->getMediumRho(imed) * app->getGeometry()->getRelativeRho(ireg);
                 EGS_Float mass = vol[ireg]*rho;
                 dose->currentResult(d_reg_index[ireg],r,dr);
                 if (r > 0) {
@@ -349,8 +349,12 @@ void EGS_DoseScoring::reportResults() {
                 else {
                     dr=1;
                 }
+                EGS_Float finalDose = 0;
+                if(mass > 0) {
+                    finalDose = r*normD/mass;
+                }
                 egsInformation("%*d %-*s %7.3f   %8.4f %10.4e +/- %-7.3f%% %10.4e +/- %-7.3f%%\n",
-                               irmax_digits,ireg,max_medl,app->getMediumName(imed),rho,vol[ireg],r*normE,dr*100.,r*normD/mass,dr*100.);
+                               irmax_digits,ireg,max_medl,app->getMediumName(imed),rho,vol[ireg],r*normE,dr*100.,finalDose,dr*100.);
             }
         }
         egsInformation("%s\n",line.c_str());
@@ -362,7 +366,7 @@ void EGS_DoseScoring::reportResults() {
             if (app->isRealRegion(ir)) {
                 imed = app->getMedium(ir);
                 EGS_Float volume = vol.size() > 1 ? vol[ir]:vol[0];
-                massM[imed] += app->getMediumRho(imed)*volume;
+                massM[imed] += app->getMediumRho(imed) * app->getGeometry()->getRelativeRho(ir) * volume;
             }
         }
         if (normE==1) {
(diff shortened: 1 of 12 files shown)


rtownson:
This PR will be superseded by PR #759

rtownson:
Closing this because I have confirmed it will be handled by PR #759 instead.

Issue #745, opened 2021-07-06
tutor7pp error: Track 2026 has length -1717986918 < 2.Segmentation fault


Vitigis:
Hello Everyone,

I´m new to EGSnrc and I´m trying to recreate the task of the "getting started"-PDF. In Chapter 5.4 "Generate particles tracks" there is a task where you simulate ptracks in an air box with an water cylinder in it (tutor7pp -i test2.egsinp -p 521icru). When I want to see the result in _egs_view_ i get the error message: "Track 2025 has length 0 < 2.Track 2026 has length -1717986918 < 2.Segmentation fault". When I change the material of the air box to water instead it works properly.
So does anyone has a clue why the ptrack-lenght is negative and how to fix it? The .egsinp-File is attached.

Greetings and thanks for help

[test2.txt](https://github.com/nrc-cnrc/EGSnrc/files/6769465/test2.txt)

ftessier:
Thank you for reporting this @Vitigis, we will try to reproduce this behaviour at our end!

rtownson:
Hi @Vitigis, it looks like the problem is that you have included two `egs_track_scoring` ausgab objects. Even though they have different names this actually isn't supported. So they both tried to output to the same ptracks file and corrupted it. Just comment one out and you should find that it works!

mxxo:
Hi @rtownson, sorry to barge in on this issue. Is this the case for all ausgab objects or just tracks? Maybe we could add a check in `EGS_Application::addAusgabObject` to catch this? 

```diff
diff --git a/HEN_HOUSE/egs++/egs_application.cpp b/HEN_HOUSE/egs++/egs_application.cpp
index c4159c3d..bc840ef9 100644
--- a/HEN_HOUSE/egs++/egs_application.cpp
+++ b/HEN_HOUSE/egs++/egs_application.cpp
@@ -929,6 +929,12 @@ void EGS_Application::addAusgabObject(EGS_AusgabObject *o) {
     if (!o) {
         return;
     }
+    for (int j=0; j<a_objects_list.size(); ++j) {
+        if (a_objects_list[j]->getObjectType() == o->getObjectType()) {
+            egsFatal("error: only one ausgab object of type '%s' is allowed\n",
+                o->getObjectType().c_str());
+        }
+    }
     o->setApplication(this);
     a_objects_list.add(o);
     //int ncall = 1 + (int)AugerEvent;
```

mxxo:
[...]

PR #751, opened 2021-07-15
Guard against more than one track_scoring object in egsinp

mxxo:
As discussed in https://github.com/nrc-cnrc/EGSnrc/issues/745, ensure that multiple `EGS_TrackScoring` objects are not defined as this can lead to bugs at write time when the objects try to write to the same file (i.e. if the `file name addition` is blank). More than one track scoring object is not seen as a required feature, since photons, electrons and positrons can all be toggled in `egs_view`.

```
:start ausgab object:
    library = egs_track_scoring
    name    = some_name
    score photons   = yes or no # optional, yes assumed if missing
    score electrons = yes or no # optional, yes assumed if missing
    score positrons = yes or no # optional, yes assumed if missing
    start scoring   = event_number # optional, 0 assumed if missing
    stop  scoring   = event_number # optional, 1024 assumed if missing
    buffer size     = size         # optional, 1024 assumed if missing
    file name addition = some_string # optional, empty string assumed if missing
:stop ausgab object:
```

### DIFF
--- a/HEN_HOUSE/egs++/egs_application.cpp
+++ b/HEN_HOUSE/egs++/egs_application.cpp
@@ -1001,6 +1001,16 @@ void EGS_Application::addAusgabObject(EGS_AusgabObject *o) {
     if (!o) {
         return;
     }
+    // only one track scoring object is allowed, otherwise there can be bugs
+    // if both objects try and write to the same file.
+    if (o->getObjectType() == "EGS_TrackScoring") {
+        for (int j=0; j<a_objects_list.size(); ++j) {
+            if (a_objects_list[j]->getObjectType() == "EGS_TrackScoring") {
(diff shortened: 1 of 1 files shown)


rtownson:
@mainegra is there a debugging feature here where you might output tracks in limited "start scoring/stop scoring" ranges that would make it useful to have multiple track ausgab objects? I had forgotten 'file name addition' was an option - maybe we just need to defend against them having the same output file..

mainegra:
Fellas, my understanding was that the possibility to modify the file name was just for that: having a ptrack file name different from the default name used based on the input file name. So I agree that one ptrack file should be enough.

ftessier (approved):

blakewalters (approved):

mainegra (approved):

PR #757, opened 2021-08-08
Add Vagrantfile file

ifurther:
Add this file and it can use Vagrant. Maybe it can more easy to use VirtualBox .


ftessier:
Good point, I like the idea! I have no looked into vagrant for a while, thanks for this reminder!!

ifurther:
I think use container like singularity more quickly.

PR #787, opened 2021-10-20
Fix a mistake in the getting-started solutions

rtownson:
In the solution to one of the questions in the getting-started manual, the total dose was calculated as the sum of dose to regions, instead of appropriately weighted by the relative volume of each region. This was corrected, as well as another typo in the instructions.

Thanks to @mjukdisk for reporting this (https://github.com/nrc-cnrc/EGSnrc/discussions/780).


blakewalters (approved):

mainegra (approved):

ftessier:
Added the "reported by" note in commit message.

PR #806, opened 2021-12-02
Fix #803 a quiksyncdef error from DYNJAWS in BEAMnrc GUI

rtownson:
Fix #803 a GUI error that occurred in DYNJAWS jaw definitions when clicking
the button to calculate the jaw settings from the field size and SSD.
The button triggered a function that was duplicated with the same name
in SYNCJAWS, leading to the error. As the fix, the syncjaws version of
the function has been renamed. This bug has existed since SYNCJAWS was
added to the GUI.

### DIFF
--- a/HEN_HOUSE/omega/progs/gui/beamnrc/syncjaws.tcl
+++ b/HEN_HOUSE/omega/progs/gui/beamnrc/syncjaws.tcl
@@ -326,12 +326,12 @@ proc define_syncjaws { id } {
 	}
     }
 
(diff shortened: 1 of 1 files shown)


ftessier:
That was quick! Thank you @rtownson.

ftessier (approved):

blakewalters (approved):
Sorry it took so long for me to get to this, @rtownson!

mainegra (approved):

Issue #844, opened 2022-02-07
material definition order causing dosimetric difference when using xcom-mcdf photoelectric cross sections


iymadmansour:
Hello! 

There seems to be an issue specific to the xcom-mcdf photoelectric cross-sections which can cause significant dosimetric differences when using otherwise identical media definitions which vary in the order of elements presented. The issue was initially identified when calculating key metrics related to miniature X-ray tubes. To illustrate this point, I've created a test case (attached) which has a very simple point source within a stainless steel cylinder. I've "modified" the stainless steel composition (by simply switching the order of the materials) and compared the resulting 3ddose files. For completeness, I've run the test case using both xcom-mcdf and xcom photoelectric cross-sections. 

When using the mcdf-xcom cross sections, relatively large dosimetric differences are measured. The same differences are not present when using the xcom cross sections. I’ve reviewed PIRS-701 section 7 (regarding pegsless mode) and I have not seen any mention of a specific order requirement for defining material (please forgive me if this is in fact not the case!). 

Questions: 
	-Is there a specific material definition template that should be followed (I noticed there was another issue that seems to touch on this, but no elaboration is provided: https://github.com/nrc-cnrc/EGSnrc/issues/785)? If so, why?
	-Why is this issue I am having specific to mcdf? 
        -Would an error message be a good idea to introduce if the defined material is not consistent with the desired material definition format? 

Ancillary information:
	-All data are calculated using egs_brachy pulled from the EGSnrc_CLRP repository with most recent commit 134c731
	-Track length scoring is used for all calculations. Muen data are provided. 

I don't know if there is (are) any additional documents/information which I can provide, but if so - please let me know!

[SimulationFiles.zip](https://github.com/nrc-cnrc/EGSnrc/files/8017443/SimulationFiles.zip)

![ToyModelComposition1-MCDF](https://user-images.githubusercontent.com/43008613/152841584-ee208f67-5e5f-4143-8ef1-5d745a17428f.png)
![ToyModelComposition1-XCOM](https://user-images.githubusercontent.com/43008613/152841587-6fa2a43e-37a8-40b5-ba09-4b1f85246c13.png)

ftessier:
Thanks for taking the time to report this issue @iymadmansour!

> Is there a specific material definition template that should be followed (I noticed there was another issue that seems to touch on this, but no elaboration is provided: https://github.com/nrc-cnrc/EGSnrc/issues/785)? If so, why?

[...]

PR #845, opened 2022-02-08
Fix serious bug in MCDF-XCOM PE routine

mainegra:
**Unexpected behavior:** Results depend on the order in which the material
composition is defined. This is clearly a bug as  elements can be entered in 
an arbitrary order to define compounds and mixtures.

**Reported by:** Iymad Mansour (issue #844) who observed that `egs_brachy` simulations of
an irradiation with an isotropic 35 keV photon point source embedded
in a steel applicator, produced 5% differences in the surface dose of
a water phantom when the order of the steel components was changed.

Confirmed with `egs_kerma` for a 35 keV photon point source surrounded by
a 1 mm steel encapsulation embedded in a spherical water phantom. Kerma
at the spherical water shell next to the encapsulation differs by 4%.

**BUG:** Sampling element and atomic shell for photo-electric interaction
     with MCDF-XCOM shell-wise cross sections using unsorted array `zelem` rather than
     `pe_zsorted.` The use of a sorted array was introduced with the MCDF-XCOM option 
     for efficiency purposes when sampling the interacting element, by checking from heaviest 
     to lightest element.

Using `pe_zsorted` array fixes the problem.

### DIFF
--- a/HEN_HOUSE/src/egsnrc.mortran
+++ b/HEN_HOUSE/src/egsnrc.mortran
@@ -5247,7 +5247,7 @@ IF( iedgfl(irl) ~= 0 ) [" User requested atomic relaxations "
       $RANDOMSET br; logE = log(peig);
       "DO k=1,nne(medium) ["
       DO k=nne(medium),1,-1 [
-          iZ = int( zelem(medium,k) + 0.5 );zpos = pe_zpos(iZ);
+          iz = int(pe_zsorted(k,medium)+0.5); zpos = pe_zpos(iZ);
           IF( iZ < 1 | iZ > $MXELEMENT ) [
               $egs_info(*,' Error in egs_shellwise_photo: ');
               $egs_fatal(*,'   Atomic number of element ',k,


mainegra:
@iymadmansour could you confirm this fixes issue #844?

iymadmansour:
@mainegra Thank you so much for looking into this so quickly! Of course, I will see if this fixes https://github.com/nrc-cnrc/EGSnrc/issues/844 and will report back asap.

iymadmansour:
@mainegra @ftessier That fix did the trick! Thank you guys once again for looking into this so quickly.  Rerunning the input files already provided yield the following results:

ToyModelComposition1-MCDF
ToyModelComposition1-XCOM

mainegra:
> @mainegra @ftessier That fix did the trick! Thank you guys once again for looking into this so quickly. Rerunning the input files already provided yield the following results:
> 
> ToyModelComposition1-MCDF ToyModelComposition1-XCOM

Fantastic! Thanks for confirming! And for finding out this bug!

ojalaj:
@mainegra Any idea the effect of this issue in the MV range? Chamber simulations!? DOSXYZnrc patient plan simulations !?

ftessier:
> @ftessier we might want to merge this one as soon as possible!

Yes, marking this for the 2022 release, due soon!

ojalaj:
@mainegra Thanks for the info! For further confirmation, if I have defined the materials either in .density files or directly in .egsinp files, with increasing element Z order, e.g. "  1 0.04196    6 0.62499      8 0.33305" or "elements = H, C", correspondingly, everything should have been fine?

mainegra:
> @mainegra Thanks for the info! For further confirmation, if I have defined the materials either in .density files or directly in .egsinp files, with increasing element Z order, e.g. " 1 0.04196 6 0.62499 8 0.33305" or "elements = H, C", correspondingly, everything should have been fine?

@ojalaj correct! I just tested that the desktop version of ESTAR sorts the elements by Z and that random selected density files in the distribution are also sorted. Nonetheless, this is not a requirement as far as I can see to define a material for use with EGSnrc.

ojalaj:
@mainegra Great! Luckily I have created most of my .density files with the desktop ESTAR, but I also found some .density files, perhaps created with the online version of ESTAR, and their element order is not increasing (but they do work no problem with EGSnrc).

blakewalters (approved):

rtownson (approved):

PR #895, opened 2022-07-08
Fix #894: revert egsphant to egs_brachy encoding

### LINKED ISSUE #894: Discrepancy between EGSnrc ctcreate and egs_glib egsphant media encoding
This issue was reported in https://github.com/clrp-code/egs_brachy/issues/22.

Briefly, #633 expanded the number of media in the EGSnrc `ctcreate` to 95. However, the number of media was _also_ independently expanded in `egs_glib` in the development of `egs_brachy`, but using a different encoding. As a result, `egsphant` files are being created using inconsistent encodings, leading to undefined behaviour and `egs_view` crashes.

The proposed solution is to change the ctcreate encoding to match the one in `egs_glib` (estimated to be used more widely), and perhaps add support for an encoding string inside the egsphant file, for flexibility and perfect clarity.

ftessier:
Revert to the `egs_brachy` encoding for `.egsphant` files, allowing for 63 media encoded with the set of alphanumeric characters (where 0 is reserved for vacuum).

Briefly, #633 expanded the number of media in `.egsphant` files to 95, using all printable ascii characters. However, the number of media was also independently expanded in `egs_glib` for the development of `egs_brachy`, but using a different encoding consisting of only the 63 alphanumeric ascii characters. This leads to inconsistent egsphant files that are no longer interchangeable.

This issue was originally reported and discussed in https://github.com/clrp-code/egs_brachy/issues/22.

### DIFF
--- a/HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
+++ b/HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
@@ -1353,6 +1353,7 @@ REAL z_score_tmp; "dummy variable passed to IAEA phsp file opening macro"
 external combine_results;
 
 character medChar($IMAX);
+character encoding*62;
 
 #ifdef HAVE_C_COMPILER;
 $REAL   part_dose, part2_dose, current_result, current_uncertainty;
@@ -1371,6 +1372,8 @@ $DECLARE_TIMING_VARIABLES;
 "cput1 is cputime at start of simulations"
 "cput2 is cputime when asked for so elapsed cpu is cput2-cput0 or -cput1"
 
+"define the ascii encoding string for media in egsphant file"
+encoding = $ENCODING;
 
 call egs_init; "initialize EGSnrc system"
 
@@ -1524,7 +1527,7 @@ IF(NMED = 0)[ "Input data from CT data file as processed by the code ctcreate"
           read($CTUnitNumber,:medformat:) (medChar(i),i=1,IMAX);
           "Convert from ASCII to a number"
           DO i=1,IMAX[
-            med($IR(i,j,k)) = mod((iachar(medChar(i)) + 47), 95);
+            med($IR(i,j,k)) = index(encoding, medChar(i)) - 1;
           ]
        ]
        read($CTUnitNumber,*);
@@ -4535,6 +4538,8 @@ REAL    xbnd($IMAX+1),             "voxel x boundaries"
         estepe($MXMED);            "linear array of estepe values"
 $REAL rho($MXREG);
 
+character  encoding(62);
+encoding = $ENCODING;
 
 REPLACE {$IRWP(#,#,#)} WITH {(1 + {P1} + ({P2}-1)*iimax + ({P3}-1)*iimax*jjmax)}
 
@@ -4551,8 +4556,8 @@ WRITE(iunit,*) (ybnd(jj),jj=1,jjmax+1);
 WRITE(iunit,*) (zbnd(kk),kk=1,kkmax+1);
 DO kk=1,kkmax[
     DO jj=1,jjmax[
-        WRITE(iunit,1399) (achar(MOD((med($IRWP(ii,jj,kk))+16),95) + 32),
-            ii=1,iimax);
+        WRITE(iunit,1399)
+        (encoding(med($IRWP(ii,jj,kk))+1:med($IRWP(ii,jj,kk))+1),ii=1,iimax);
     ]
     WRITE(iunit,*);
 ]
--- a/HEN_HOUSE/omega/progs/ctcreate/ctcreate.mortran
+++ b/HEN_HOUSE/omega/progs/ctcreate/ctcreate.mortran
@@ -2040,6 +2040,8 @@ REAL    xbnd($IMAX+1),             "voxel x boundaries"
         rho($MXREG),               "linear array of densities"
         estepe($MXMED);            "linear array of estepe values"
 
+character  encoding*62;
+encoding = $ENCODING;
 
 REPLACE {$IR(#,#,#)} WITH {(1 + {P1} + ({P2}-1)*iimax + ({P3}-1)*iimax*jjmax)}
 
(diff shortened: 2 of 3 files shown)


ftessier:
This is a quick solution, where the encoding is hard-coded in `ctcreate` and `dosxyznrc`. @blakewalters can you clean this up to put the variable definition inside a proper macro so that it is defined only once? Ideally we would add an encoding string _inside_ the `.egsphant` file, but that can wait.

rtownson:
All of the material type declarations for each voxel was on a new line in the egsphant, instead of printed in a slice-by-slice grid. My commit just fixed that.

HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
```diff
@@ -1518,13 +1519,14 @@ IF(NMED = 0)[ "Input data from CT data file as processed by the code ctcreate"
...
     "input the media numbers"
+    encoding = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
     DO k=1,KMAX[
        DO j=1,JMAX[
           "Read in the characters that correspond to medium numbers"
           read($CTUnitNumber,:medformat:) (medChar(i),i=1,IMAX);
           "Convert from ASCII to a number"
           DO i=1,IMAX[
-            med($IR(i,j,k)) = mod((iachar(medChar(i)) + 47), 95);
+            med($IR(i,j,k)) = index(encoding, medChar(i)) - 1;
```
rtownson, line 1529: I didn't test this line, @ftessier did you?

rtownson:
> This is a quick solution, where the encoding is hard-coded in `ctcreate` and `dosxyznrc`. @blakewalters can you clean this up to put the variable definition inside a proper macro so that it is defined only once? Ideally we would add an encoding string _inside_ the `.egsphant` file, but that can wait.

@blakewalters I added this macro, so you don't have to :)

HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
ftessier: Not yet!

ftessier:
Can we try 6bb8618 instead, using `advance='no'` to prevent newlines. This is Fortran 90, is anyone still compiling with `f77` you think?

rtownson:
> Can we try [6bb8618](https://github.com/nrc-cnrc/EGSnrc/commit/6bb861831e8b36cf5f737acdeb75b153fa14ce33) instead, using `advance='no'` to prevent newlines? This is Fortran 90, is anyone still compiling with `f77` you think?

Sure! I feel like we should be safe to switch to a standard that came out in 1991...

rtownson:
Just leave the my other commit in the history, in case anyone ever comes to us wanting a version that works on f77.

ftessier:
> Just leave the my other commit in the history, in case anyone ever comes to us wanting a version that works on f77.

I will squash this before merging, but will leave a note in the commit message!

HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
```diff
@@ -3727,11 +3729,6 @@ IF(iarg=5 & i_phsp_out~=0 & ir(np)=irmax+1 & irold<ir(np))[
              w(np)=r_13(k_field)*u_temp + r_23(k_field)*v_temp +
                    r_33(k_field)*w_temp;
 
-           IF(i_iaea_in=1)[
-          "for IAEA phsp source we need to add SAD back to Z coordinate"
-                  z(np) = z(np) + iaea_SAD;
-           ]
-
            "Write the particle to a phsp"
```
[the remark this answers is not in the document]
ftessier: Yes you are right, I will fix it. Good catch.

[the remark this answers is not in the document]
ftessier: Fixed with 2e057e7 force push.

HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
ftessier: @rtownson do you have a ctcreate'd phantom around on which to test this conversion?

ftessier:
In the end I reverted to using the implicit array over `ii=1,iimax` to avoid multiple write calls. In my cursory testing, the implicit array form was significantly faster. For the _record_, the proper syntax of the write statement with `advance='no'` would be:

```fortran
write(15, '(a1)', advance='no') encoding(code:code)
```

ftessier:
Fixed spacing.

ftessier:
**Before merging:** fix encoding string length typo in commit message. The encoding is 62 characters, and `0` is reserved for vacuum; so effectively 61 media can be encoded.

HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
blakewalters: I don't know about your compilers, but I ran into a compilation error complaining that encoding had been defined before all variables had been declared.  I got rid of this by moving the "encoding = $ENCODING;" statement to immediately before "call egs_init;"

ftessier: Strange; it compiled fine here. We should be able to define a variable locally in the write_phantom subroutine, no?

ftessier: Strange, it compiled fine here. We should be able to define it as a local variable in the write_phantom subroutine, no? What is going on?

ftessier: Ah ok, I see, I was looking at the _other_ one in write_phantom. Yes, ok, I see what you mean, will fix right away.

ftessier: Fixed, but leaving this conversation unresolved because this still requires a round of testing.

blakewalters (approved):
I did an "end-to-end" calculation, using ctcreate to generate an .egsphant file with 12 media and then running this through a dosxyznrc simulation and didn't find any issues.

rtownson (approved):

mainegra (approved):

HEN_HOUSE/user_codes/dosxyznrc/dosxyznrc.mortran
ftessier: Resolved with testing by @blakewalters.

ftessier:
Squashed the commits into a single one for merging into `develop`.

mchamberland:
@ftessier @rtownson @MartinMartinov  There is still a discrepancy between this encoding and egs_brachy's: this one starts at 0, but egs_brachy's starts at 1. [See here](https://github.com/clrp-code/EGSnrc_CLRP/commit/2718117b3d6f4fdafb681530e5465ab0898282de).

Or did I miss something in the discussion?

PR #896, opened 2022-07-11
Use position/direction sampling from base source for radionuclide sources

ftessier:
This pull request replaces #672. The original pull request branch was removed from the forked repository, so the branch and pull request is hereby reinstated. The original pull request message was:

> The implementation of `egs_radionuclude_source` is changed so that it takes an input "base source" instead of defining a collimated or isotropic source. This way you can use any of the egs sources or a custom egs source to define the location of the radionuclide in the simulation. The change in particular was made towards fast simulations targeted radionuclide therapy across a whole patient phantom.

### DIFF
--- a/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
+++ b/HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
@@ -23,7 +23,7 @@
 #
 #  Author:          Reid Townson, 2016
 #
-#  Contributors:
+#  Contributors:    Martin Martinov
 #
 ###############################################################################
 */
@@ -74,9 +74,10 @@
 
 \ingroup Sources
 
-A radionuclide source is a source that delivers particles with
-directions uniformly distributed in \f$4 \pi\f$ emitted from
-\link EGS_BaseShape any shape. \endlink
+A radionuclide source is a source that delivers particles according to a
+radionuclide decay scheme. It follows events decay-by-decay, so that internal
+transitions and atomic relaxations from decay events are produced in a
+correlated fashion from the same location.
 
 Note that \ref EGS_RadionuclideSource is an experimental source and only a
 subset of the available radionuclides have been tested against measurement.
@@ -99,13 +100,13 @@ creation due to disintegrations is sampled per-shell rather than
 considering sub-shell probabilities. I.e. probabilities for vacancy creation
 in the K,L,M,N,O shells are considered, but not L1, L2, etc. The actual
 sub-shell in which the vacancy created is sampled uniformly for the given
-shell. This is an approximation, but only relevant when
+shell. This is an approximation, and only relevant when
 '<code>atomic relaxations = eadl</code>' in
 \ref EGS_RadionuclideSpectrum.
 - Alpha particles are not transported.
 - Atomic motion & recoil from emissions is not modeled.
 - For some radionuclides the decay intensities do not add to exactly 100%, due to
-uncertainties on the intensity values. To normalize the decay intensities and
+uncertainties on the intensity values in the data files. To normalize the decay intensities and
 allow for modeling, the discrepancy from 100% is divided over all decays,
 scaled by the uncertainty of each decay.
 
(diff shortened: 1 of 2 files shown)


ftessier:
Rebased on develop, applied astyle.

rtownson:
Yes the missing contributor/documentation looks like a mistake, not sure how it happened.

When you re-instate it, change the indentations to use spaces instead of tabs.

ftessier:
Yes I will run astyle. So are you confirming that the other changes in that commit are all good then? I will attempt to clean it up and wait for review.

rtownson:
> Yes I will run astyle. So are you confirming that the other changes in that commit are all good then? I will attempt to clean it up and wait for review.

Yes, the comments from the first commit are correct. The code added in the second commit is also correct.

This PR breaks backwards compatibility - users with radionuclide sources will have to update their input files. I think that's OK, maintaining backwards compatibility would result in a lot of duplicate code.

rtownson:
I'm going to add one more commit to add a check for the old depreciated inputs, and give a description of the new format.

mainegra (approved):

HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
```diff
@@ -137,20 +137,19 @@ on the number of disintegration events (tracked by <b> \c ishower </b>). This
...
-source on the spectrum, and neglect potential contributions from surrounding
-structures. Restricting the emission angles by collimation generally constitutes
-a physical approximation of your geometry and is an approximate efficiency
-enhancement technique. It is also important to note that the collimated source
-determines the fluence with \f$N/d^2\f$, where \c N is the number of
-disintegrations sampled and \c d is the user-defined minimum distance between the
-source and target shapes.
+A radionuclide source is defined using the following input. It imports a base
+source and uses the base source getNextParticle() invocation to determine decay
+location. This implementation leads to increased random number sampling than
```
blakewalters, line 142: Why don't we just use an EGS_Shape instead of defining a base source if this is only used to define decay location?

HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
```diff
@@ -255,9 +234,8 @@ results for non-disintegration emissions.
...
 :start source definition:
     :start source:
-        name                = my_source
-        library             = egs_radionuclide_source
-        activity            = 28e6
+        name                = source_location
+        library             = egs_isotropic_source
         geometry            = my_envelope
         region selection    = IncludeSelected
         selected regions    = 1 2
```
blakewalters, line 241: Although I now see the base source (source_location), I don't see the egs_radiohead_source (i.e. my_source) defined in the updated example.  Is this missing?

HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
rtownson: Maybe the text is not quite clear - it also uses the direction sampling and getFluence from the base source, so it can uses the more complex source modelling (like a collimated source, which is not just a shape).

blakewalters: Ah okay.  Yes, I see further up in the text it says it makes use of getFluence, but maybe it should also mention that it uses the incident directions as well.

HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
rtownson: You're right! I've fixed this.

HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
rtownson: Done, thanks!

blakewalters (approved):

HEN_HOUSE/egs++/sources/egs_radionuclide_source/egs_radionuclide_source.h
blakewalters: Thanks, @rtownson!

ftessier:
Rebased on develop, squashed and tweaked commit message. Also, remove the year from the contributor line in the header, to match EGSnrc style (only the author has the year indicated, to preserve pre-git history; now contribution date is provided by the git log)

PR #898, opened 2022-07-12
Fix git branch logging and egs_home vs HEN_HOUSE code comparisons

rtownson:
Fix an issue with long branch names where the fortran code would fail to compile because of exceeding the line length. This isn't completely fixed, but now a few more characters are allowed before it crashes. Correct me if I'm wrong, but I think it's not possible to fix completely, because it seems like the DEFINE replacement means that we can't use mortrans usual automatic line wrapping.

Fix the comparisons between egs_home and HEN_HOUSE/user_code applications, so that it doesn't report a difference when the HEN_HOUSE version doesn't exist to start with.

This fixes minor issues introduced in PR #414.

### DIFF
--- a/HEN_HOUSE/gui/egs_configure/egs_install.h
+++ b/HEN_HOUSE/gui/egs_configure/egs_install.h
@@ -655,8 +655,6 @@ static const char machine_macros[]={
 "\n"\
 "REPLACE {$GIT_HASH} WITH \n"\
 "  {'$git_hash'}; \n"\
-"REPLACE {$GIT_BRANCH} WITH \n"\
-"  {'$git_branch'}; \n"\
 "REPLACE {$CONFIG_TIME} WITH \n"\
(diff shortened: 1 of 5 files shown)


blakewalters (approved):

mainegra (approved):

ftessier:
@rtownson perhaps we can remove the git branch name altogether, and only keep the git commit hash? The branch name may cause confusion in fact, since it is a label that can point to any commit. What do you think?

rtownson:
@ftessier sure, I see your point. It should always be paired with a commit to avoid confusion, but the potential issue with long branch names might not be worth keeping it around.

ftessier:
@rtownson something like this, to remove all instances of `GIT_BRANCH`?

ftessier (approved):

ftessier:
Rebased on develop and squashed.

PR #899, opened 2022-07-14
Add checks for gcc version when setting the default c++ standard

rtownson:
Add checks to the configure script, configuration GUI, and egs_view profile before setting `std=c++14` as the default c++ standard. If the gcc version is 4 or less, then `std=c++11` is used as the default. This just helps protect against failed installations for users with older gcc versions using the default config settings.

Very old versions of gcc might still crash, but the user can easily change the `std=` flag as necessary.


blakewalters (approved):

ftessier (approved):

mainegra (approved):

ftessier:
Squashed and reworded commit title to fit.

ftessier:
Added a bit of info in the commit message.

PR #905, opened 2022-07-19
Fix #904: avoid gcc -mcmodel flag on macos systems

### LINKED ISSUE #904: Compilation fails with -mcmodel=medium on macOS
In #725 the compiler option `-mcmodel=medium` was added to the default flags, but this causes compilation failures on macOS (testing on version 11.6.7). On macOS, the `clang` linker is used, which apparently fails to link with code compiled with `-mcmodel=medium`. Specifically, out of the box on branch `develop` with the default options we get the following failures:

```
Compiling statdose ...                            Failed
...
Building the IAEA phase space library ...         Failed
Building the EGSnrc C++ class library ...         Failed
```

(I have no clue why `statdose` fails!). Hence, this compilation flag should **not** be set on macOS. Removing the `-mcmodel=medium` compiler option solves the problem.

ftessier:
Add tests in the configure scripts to avoid setting the gcc memory model flag `-mdmodel=medium` if a darwin (macOS) system is detected. This flag causes compilation failures at the linking stage.

### DIFF
--- a/HEN_HOUSE/scripts/configure
+++ b/HEN_HOUSE/scripts/configure
@@ -123,11 +123,13 @@ _ACEOF
 
 read response
 
+memory_model="-mcmodel=medium"
 printf $format "Checking system type ... " >&2
 canonical_system=$($my_dir/config.guess)
 system_name=$(uname -s | sed ":/:_:g" | tr '[:upper:]' '[:lower:]')
 if test "x$system_name" = xdarwin; then
     system_name="osx"
+    memory_model=""
 fi
 echo "$canonical_system" >&2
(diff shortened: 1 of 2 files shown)


ftessier:
@rtownson @mainegra I don't think we need to change the `egs-configure` gui, since it is never compiled for macOS, right?

rtownson:
> @rtownson @mainegra I don't think we need to change the `egs-configure` gui, since it is never compiled for macOS, right?

Correct, we don't pre-compile any GUIs for macOS currently.

blakewalters (approved):
Yes!!!!

mainegra (approved):

PR #910, opened 2022-08-11
Add the egs_gammaspec app

rtownson:
Add an application for calculating gamma spectrometry related quantities, like detector efficiency and coincidence summing corrections. It outputs spectra both with coincidence summing effects included (e.g. a realistic or "non-perfect" detector), and without summing effects (a "perfect" detector). Total efficiency and full energy peak efficiency are also calculated for each case.

The `egs_radionuclide_source` was changed slightly to allow for the calculations - it now returns the number of decays from `getNextParticle`, instead of the number of source particles. This does not change the normalization when using the source, but apps specially designed to use `current_case` and `last_case` for the radionuclide source may be affected. This will only affect users who have written their own applications, not any of the applications distributed with EGSnrc.

The application specific inputs are:
```
:start scoring options:

    :start output spectrum:
        scoring regions = crystal_no_dead_label # The region numbers or region label(s) denoting the sensitive regions

        minimum spectrum energy  = 0.0 # Optional, MeV, default=0, minimum energy in output spectrum
        maximum spectrum energy = 0.4 # Optional, MeV, defaults=maximum in source, maximum energy in output spectrum
        number of bins  = 2000 # Optional, default=1000, number of bins in output spectrum

        # Example: these are peaks of interest for Ba-133, in MeV
        # Optional, coincidence summing corrections will be calculated for each of these
        gamma analysis energies = .0309 .035 .0531 .0796 .0810 .1606 .2232 .2764 .3028 .356 .3838

        # Optional, yes or no, default=yes, get the analysis energies automatically, using all of the gamma energies from the radionuclide decay scheme. They will be combined with the manually entered 'gamma analysis energies', so make sure they don't overlap. Only works for egs_radionuclide_source.
        automatic analysis energies = no

        # Optional, MeV, default=1e-6, the minimum energy from a single shower that can be detected in the sensitive regions
        minimum detectable energy = 1e-6
    :stop output spectrum:

:stop scoring options:
```


blakewalters (changes requested):
Compiles, runs and seems to produce the desired output, @rtownson!  Just a couple of questions: 1) Would it be useful to have an option to output these efficiencies to a file? and 2) Does this branch need to be rebased on develop before merging?

rtownson:
I think since the list of efficiencies is usually just a few numbers, it's OK to have it just be in the .egslog file, similar to the dose reported for `egs_chamber`.

Wouldn't be a bad idea to do a more recent rebase before merging.

rtownson:
There are quite a few people using this branch - it would be nice to review & merge it into the 2024 release.

mainegra (approved):

blakewalters (approved):

ftessier (approved):

PR #940, opened 2022-11-24
Fix egs_cbct, egs_fac compilation and egs_cbct example

rtownson:
Fix a compiler error that occurred with newer compiler settings due to ambiguous declaration of isinf and isnan, by adding `std::`.  

Also change the z-rotation parameter in the example input file to be 0 so that it calculates the first projection instead of a later projection. This avoids a common user error when trying to produce the blank scan for the first time, as reported here: https://www.reddit.com/r/EGSnrc/comments/z36841/query_fatal_error_message_egs_cbct_blank_scan/

### DIFF
--- a/HEN_HOUSE/user_codes/egs_cbct/egs_cbct.cpp
+++ b/HEN_HOUSE/user_codes/egs_cbct/egs_cbct.cpp
@@ -1473,8 +1473,8 @@ void EGS_CBCT::initOutput() {
                 delete bs;
                 egsFatal(
                 "\n\n***  Wrong blank scan file size = %d bytes\n"
-                "     It should be %d bytes"
-                "     This is a fatal error.\n\n",
+                "     It should be %d bytes\n"
+                "     Check that the blank scan 'z-rotation' is set to 0\n\n",
                 fileSize,Nx*Ny*sizeof(float));
               }
               else{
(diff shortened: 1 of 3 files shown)


mainegra:
@rtownson thanks for looking into this! My only comment is that in the case of the non-zero rotation, if both, the blank scan and the actual scan use the same rotation, then there should be no issue as the projection would be place at the same location of their corresponding scan files!

mainegra:
Setting the rotation to 0 will simply show a different phantom scan. So, it really doesn't matter.

mainegra:
> @rtownson thanks for looking into this! My only comment is that in the case of the non-zero rotation, if both, the blank scan and the actual scan use the same rotation, then there should be no issue as the projection would be place at the same location of their corresponding scan files!

Actually, hold that thought! Something is not right! If I were to implement this now, I would just produce a blank scan whose size has to be Nx*Ny*size(float) since this is the same for any projection. Clearly, the blank scan produced with a non-zero angle makes no sense! So, your recommendation to set it to zero is the correct one! Ideally egs_cbct should not care about projections placement for a blank scan!

mainegra (changes requested):
Perhaps add to the message reporting blank scan size mismatch a hint to set rotation to zero?

ftessier:
I will merge this into `master` now because otherwise the `v2023` pre-release does not compile `egs_cbct` nor `egs_fac`!

rtownson:
Added the egs_fac fixes, now ready to merge.

ftessier (approved):

ftessier:
Squashed the 3 commits.

Issue #804, opened 2021-12-02
The egs_chamber sample input files don't run out-of-the-box


ftessier:
The sample input files provided in the `egs_chamber` application directory are complex, and moreover depend on a BEAMnrc source or materials that are not readily available. Hence, the examples don't run out-of-the-box after a fresh EGSnrc install. Also, a  much simpler sample input file (pencil beam on slab) would prove useful.

Specifically, this was reported by Dave Rogers on 8 March 2014, but was never fixed:

> 1. `example1` uses a beam source as input and since this BEAM model is not
part of the distribution, one cannot run the example.  I would propose that
source be commented out and a simpler point source be used. Leave the beam
source there to show how it would be done, but it is better if an example
input file will actually run. Also, near the top say that it needs
521icru.pegs4dat to run.

> 2. `example2` makes use of materials that are not on the distribution (e.g.
C173521ICRU) so these materials should be changed to similar materials in
521icru and a similar note added to the top of the file re using 521icru.

mxxo:
Just to add to this, `example1_co60_Pfactors.egsinp` fails with this error:

```text
$ egs_chamber -i example1_co60_Pfactors.egsinp -p 521icru
In EGS_Application::initSimulation()
**********************************************
**********************************************
**********************************************

createGeometry: Error: multiple geometries with the same name exist: chamber_in_water

```

ftessier:
Fixed by #952 (merged 2023-01-16, included in the 2025 release): the examples now run out-of-the-box with 521icru media, BEAM source kept as commented illustration.

PR #952, opened 2023-01-09
Fix #945, #804: egs_chamber examples don't run by default

### LINKED ISSUE #945: Typo in egs_chamber exapmle2
The input file of egs_chamber example seems to have tiny typos.

https://github.com/nrc-cnrc/EGSnrc/blob/a6fc389c6465c949645380976f38a013a639759c/HEN_HOUSE/user_codes/egs_chamber/example2_chamber_dd.egsinp#L302

```
x-planes = -30  30
y-planes = -30 30
```

This means  60x60cm^2, not 30x30 cm^2, right?
The same is true for the water region.

I'll send out a PR if you're willing to revise it.

rtownson:
Fix the egs_chamber examples so that they now run without needing to compile any BEAMnrc accelerators or make changes to the input file. Also fix a typo in the example2 water box size.

### DIFF
--- a/HEN_HOUSE/user_codes/egs_chamber/example1_co60_Pfactors.egsinp
+++ b/HEN_HOUSE/user_codes/egs_chamber/example1_co60_Pfactors.egsinp
@@ -23,18 +23,25 @@
 #
 #  Author:          Iwan Kawrakow, 2009
 #
-#  Contributors:
+#  Contributors:    Reid Townson
 #
 ###############################################################################
 #
-#  An example input file for egs_chamber.
+#  An example input file for egs_chamber. This is an advanced example.
 #
 #  This file defines an egs_chamber simulation of a farmer type chamber at
 #  5 cm depth in a 30x30x30 cm^3 water phantom irradiated by a Co-60 beam from
-#  a BEAMnrc simulation source. The XCSE technique is used combined with
+#  a BEAMnrc simulation source. The chamber model is just an example, and does
+#  not match any particular specifications.
+#
+#  The XCSE variance reduction technique is used combined with
 #  correlated sampling for calculation in chamber geometries that do not have
 #  a stem, a central electrode, a wall, etc., to compute the various
 #  correction factors known as Pstell, Pcell, Pwall, and Prepl.
+#
+#  Run this example using 521icru.pegs4dat:
+#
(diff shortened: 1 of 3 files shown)


ftessier (approved):
Nice, thank you!

mainegra (approved):
Great!

blakewalters (changes requested):
Looks good, but one thing: When I run example2_chamber_dd, I get a bunch of:
"initScoring: missing/wrong 'sub geometries'  for geometry ..." messages
followed by:
"initScoring: different medium 0 in region 5 compared to medium 2 in region 1 for geometry phsp_scoring_geometry.
Hope you know what you are doing"
And then my run immediately fails with a segmentation fault.  The latter may be a function of my system (using the Mac clang compiler), but what about those warning/error messages?  Are they also due to my system?

rtownson:
> Looks good, but one thing: When I run example2_chamber_dd, I get a bunch of: "initScoring: missing/wrong 'sub geometries' for geometry ..." messages followed by: "initScoring: different medium 0 in region 5 compared to medium 2 in region 1 for geometry phsp_scoring_geometry. Hope you know what you are doing" And then my run immediately fails with a segmentation fault. The latter may be a function of my system (using the Mac clang compiler), but what about those warning/error messages? Are they also due to my system?

The warnings are normal! The segmentation fault is not.

blakewalters:
@rtownson, do you know if anyone else gets the segmentation fault on a Mac?  I'll go ahead and change this to an approval, because even if this is a more general issue, it's separate from this.

blakewalters (approved):

mchamberland:
> @rtownson, do you know if anyone else gets the segmentation fault on a Mac? I'll go ahead and change this to an approval, because even if this is a Mac-related issue, it would be a separate PR.

I will try reproducing this weekend.

mchamberland:
> @rtownson, do you know if anyone else gets the segmentation fault on a Mac? I'll go ahead and change this to an approval, because even if this is a Mac-related issue, it would be a separate PR.

@blakewalters @rtownson  Could not reproduce the seg fault on my M2 Mac with the latest GCC toolkit from Homebrew.

ftessier:
Just trimmed the commit title to fit.

PR #958, opened 2023-02-06
Fix photonuclear cross section biasing when using cross section enhancement

ftessier:
Add ausgab calls before and after a photonuclear interactions, in order to handle cross section enhancement (xcse) correctly. Without these calls, there is no compensation for the enhanced cross section, hence the photonuclear cross section is effectively increased.

Also, update a couple loops that print out the ausgab flags (IAUSFL) values, to include the photonuclear attenuation input value, and fix a couple typos.

### DIFF
--- a/HEN_HOUSE/omega/beamnrc/beamnrc.mortran
+++ b/HEN_HOUSE/omega/beamnrc/beamnrc.mortran
@@ -4279,10 +4279,13 @@ IF( use_cs_enhance ) [
     IF( iausfl(21) = 0 ) cse_return(21) = .true.;
     IF( iausfl(24) = 0 ) cse_return(24) = .true.;
     IF( iausfl(25) = 0 ) cse_return(25) = .true.;
+    IF( iausfl(30) = 0 ) cse_return(30) = .true.;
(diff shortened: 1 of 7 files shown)


mainegra (approved):
@ftessier this ended up being quite an undertaking!

ftessier:
Fixed the photonuclear `xcse` biasing across EGSnrc. There are other instances where photonuclear asugab should be called, notably when setting `ZLAST`, but I want to keep this commit about `xcse`. Also updated the calls within the Fano test, for consistency, but that is moot because the Fano test will fail anyways when photonuclear is turned on, since photon energy is lost, not deposited.

I have checked that all user codes and BEAM_EX10MeVe compile as expected.

@nrc-cnrc/egsnrc please review as soon as possible so that we can include this fix in release 2023.

rtownson (approved):
All the changes here look correct. I'm still looking at egs_cbct and egs_fac but so far I don't think they need any changes.

blakewalters (approved):
Looks good to me!

ftessier:
Squashed the commits.

PR #972, opened 2023-03-07
Fix compilation on windows for some time zones, add lquadmath

rtownson:
1. Fix a bug where applications would fail to compile on windows in locations where the time zone was a long string. This was due to the COMPILE_TIME flag that included some text about the time zone. To fix the issue, the time zone was removed from the variable and output.

2. Fix an issue where lquadmath was not included with lgfortran, and the compiler would fail on some systems. 

3. Fix a compiler error that occurred when the user had installed git, but then installed EGSnrc using the zip file. Users would see an error about GIT_HASH not being defined in this case.

4. Fix a minor warning that was printed about it not being a git repository.

@ftessier please check that the lquadmath changes don't cause any issues on OS X. I recommend that we cherry pick these fixes into v2023.

### DIFF
--- a/HEN_HOUSE/gui/egs_configure/egs_tools.cpp
+++ b/HEN_HOUSE/gui/egs_configure/egs_tools.cpp
@@ -769,7 +769,7 @@ EGS_DSO::EGS_DSO(const QString &cpp_name){
 #elif defined(Q_OS_LINUX) || defined(Q_OS_UNIX)
    fpic = is_x86_64() ? "-fPIC" : QString();
 #endif
-   flibs = cpp_name.contains("g++") ? "-lgfortran" : QString();
+   flibs = cpp_name.contains("g++") ? "-lgfortran -lquadmath" : QString();
 }
 
 // Initialize default GCC compiler as gfortran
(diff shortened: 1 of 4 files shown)


ftessier:
Testing on macOS Big Sur 11.7.4 `20G1120` on 2017 macbook (Dual-Core Intel Core i5). @mchamberland would you be willing to test compilation on M1 and M2 if you can spare a few cycles?

mchamberland:
@ftesser Will test when I have a few minutes.

ftessier:
> Testing on macOS Big Sur 11.7.4 `20G1120` on 2017 macbook (Dual-Core Intel Core i5).

**Compilation test completed succesfully.**

mchamberland:
@ftessier Compiles fine on M2 MacBook Air, with Ventura 13.2.1 and GCC 12.2 from Homebrew.

rtownson:
On Windows the lquadmath changes won't be applied until we recompile the configuration GUI and post it with the release. In the mean time, users can just add `-lquadmath` to their `specs/egspp_win3264.spec` file, e.g.:

```
fortran_libs = -lgfortran -lquadmath
```

You only need to do this if you can't compile applications and get an error about quadmath.

rtownson:
Please review this soon, it's impacting a number of users trying to install the latest release of EGSnrc.

ftessier (approved):

blakewalters (approved):

mainegra (approved):

ftessier:
Thanks, I will merge this one asap.

rtownson:
Since we hadn't merged this in yet, I added a couple other fixes to windows compilation (original post updated). It seems that on windows, many users were installing git but then still using the zip file to install, which was causing issues.

PR #974, opened 2023-03-11
Add dbs to radiative splitting

blakewalters:
This is a draft PR for the DBS option in the EgsRadiativeSplitting ausgab object and will be migrated to a regular PR once testing is complete, some of the more obvious bugs are ironed out, and it's been rebased on the current develop branch.

EGS developers are encouraged to download and try out the branch and provide comments and corrections as needed.  Note also that posting of test results in the comment feed below is encouraged.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
@@ -1,0 +1,52 @@
+/*
+###############################################################################
+#
+#  EGSnrc tutor7pp application array sizes headers
+#  Copyright (C) 2015 National Research Council Canada
+#
+#  This file is part of EGSnrc.
+#
+#  EGSnrc is free software: you can redistribute it and/or modify it under
+#  the terms of the GNU Affero General Public License as published by the
+#  Free Software Foundation, either version 3 of the License, or (at your
+#  option) any later version.
+#
+#  EGSnrc is distributed in the hope that it will be useful, but WITHOUT ANY
+#  WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+#  FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for
+#  more details.
+#
+#  You should have received a copy of the GNU Affero General Public License
+#  along with EGSnrc. If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+#
+#  Author:          Iwan Kawrakow, 2005
+#
+#  Contributors:
+#
+###############################################################################
+#
+#  Defines he maximum number of media (MXMED) and the maximum number of
+#  particles on the stack (MXSTACK). This file gets included by the egsnrc
+#  fortran subroutines (egsnrc_$my_machine.F), the base application
+#  (egs_simple_application.cpp or egs_advanced_application.cpp in
+#  $HEN_HOUSE/egs++), and possibly the user code, if it uses the particle
+#  stack or one of the structures that depends on the maximum number of media.
+#
+###############################################################################
+*/
+
+
+#ifndef ARRAY_SIZES_
+#define ARRAY_SIZES_
+#define MXMED 10
+#define MXSTACK 1000
+#define MXBRES 100
+#define MXBRXS 50
+#define MXEL 50
+#define MXPWR2I 50
+#define MXGE 2000
+#define MXRAYFF 100
+#define MXSGE 400
+#endif
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
@@ -1,7 +1,7 @@
 /*
 ###############################################################################
 #
-#  EGSnrc tutor7pp application array sizes headers
+#  EGSnrc radiative splitting ausgab object array sizes headers
 #  Copyright (C) 2015 National Research Council Canada
 #
 #  This file is part of EGSnrc.
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
@@ -41,7 +41,7 @@
 #ifndef ARRAY_SIZES_
 #define ARRAY_SIZES_
 #define MXMED 10
-#define MXSTACK 1000
+#define MXSTACK 1000000
 #define MXBRES 100
 #define MXBRXS 50
 #define MXEL 50
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
@@ -248,13 +248,6 @@ void EGS_PhspScoring::storeParticle(EGS_I64 ncase) {
         score_mu=false;
     }
 
-    if (app->top_p.q==0 && sqrt(app->top_p.x.x*app->top_p.x.x+app->top_p.x.y*app->top_p.x.y)<10 &&
-        app->top_p.wt==1.0)
-    {
-        egsInformation("np=%d iq=%d E=%g x=%g y=%g z=%g wt=%g\n",app->Np,app->top_p.q,app->top_p.E,app->top_p.x.x,
-app->top_p.x.y,app->top_p.x.z,app->top_p.wt);
-        exit(1);
-    }
     //counters, min. and max. k.e.
     EGS_Float prm = app->getRM();
     count++;
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
@@ -248,6 +248,13 @@ void EGS_PhspScoring::storeParticle(EGS_I64 ncase) {
         score_mu=false;
     }
 
+    if (app->top_p.q==0 && sqrt(app->top_p.x.x*app->top_p.x.x+app->top_p.x.y*app->top_p.x.y)<10 &&
+        app->top_p.wt>0.001)
+    {
+        egsInformation("iq=%d E=%g x=%g y=%g z=%g wt=%g\n",app->top_p.q,app->top_p.E,app->top_p.x.x,
+app->top_p.x.y,app->top_p.x.z,app->top_p.wt);
+        exit(1);
+    }
     //counters, min. and max. k.e.
     EGS_Float prm = app->getRM();
     count++;
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_phsp_scoring/egs_phsp_scoring.cpp
@@ -249,9 +249,9 @@ void EGS_PhspScoring::storeParticle(EGS_I64 ncase) {
     }
 
     if (app->top_p.q==0 && sqrt(app->top_p.x.x*app->top_p.x.x+app->top_p.x.y*app->top_p.x.y)<10 &&
-        app->top_p.wt>0.001)
+        app->top_p.wt==1.0)
     {
-        egsInformation("iq=%d E=%g x=%g y=%g z=%g wt=%g\n",app->top_p.q,app->top_p.E,app->top_p.x.x,
+        egsInformation("np=%d iq=%d E=%g x=%g y=%g z=%g wt=%g\n",app->Np,app->top_p.q,app->top_p.E,app->top_p.x.x,
 app->top_p.x.y,app->top_p.x.z,app->top_p.wt);
         exit(1);
     }
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
@@ -41,7 +41,15 @@ extra_dep = $(addprefix $(DSOLIBS), $(my_deps))
 
 include $(SPEC_DIR)egspp_libs.spec
 
-$(make_depend)
+$(DSO2)$(library).$(obje): $(library).cpp $(library).h
+
+egsnrc_$(my_machine).$(obje): egsnrc_$(my_machine).F array_sizes.h
+	$(F77) $(FC_FLAGS) $(FDEFS) -c $(FOUT)$@ $<
+
+egsnrc_$(my_machine).F: $(CPP_SOURCES)
+	@echo "Mortran compiling EGSnrc sources ..."
+	@$(MORTRAN_EXE) -s -i -d $(MORTRAN_DATA) -f $(CPP_SOURCES) -o7 $@ \
+	-o8 $(@:.F=.mortlst)
 
 test:
 	@echo "common_h2: $(common_h2)"
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
@@ -43,7 +43,7 @@ ABS_EGSPP = $(HEN_HOUSE)$(EGSPP)$(DSEP)
 CPP_SOURCES = $(EGS_SOURCEDIR)egsnrc.macros $(MACHINE_MACROS) \
             $(EGS_SOURCEDIR)pegs4_macros.mortran \
             $(EGS_INTERFACE)egs_c_interface2.macros \
-            $(EGSPP_USER_MACROS) \
+            egs_radiative_splitting.macros \
             $(MACHINE_MORTRAN) $(EGS_SOURCEDIR)egs_utilities.mortran \
             $(EGS_INTERFACE)egs_c_interface2.mortran \
             $(EGS_SOURCEDIR)transportp.macros \
@@ -52,26 +52,17 @@ CPP_SOURCES = $(EGS_SOURCEDIR)egsnrc.macros $(MACHINE_MACROS) \
             $(EGS_SOURCEDIR)pegs4_routines.mortran \
             $(EGS_SOURCEDIR)egsnrc.mortran
 
-dep_advanced_application = $(ABS_EGSPP)egs_advanced_application.cpp \
-        $(ABS_EGSPP)egs_advanced_application.h array_sizes.h \
-        $(ABS_EGSPP)egs_input.h $(ABS_EGSPP)egs_base_source.h \
-        $(ABS_EGSPP)egs_object_factory.h $(ABS_EGSPP)egs_timer.h \
-        $(ABS_EGSPP)egs_application.h $(ABS_EGSPP)egs_run_control.h
-
 library = egs_radiative_splitting
 lib_files = egs_radiative_splitting
 my_deps = $(common_ausgab_deps)
-extra_dep = $(addprefix $(DSOLIBS), $(my_deps)) egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje) egs_advanced_application_lib_osx3.o
-extra = egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje) egs_advanced_application_lib_osx3.o
+extra_dep = $(addprefix $(DSOLIBS), $(my_deps)) egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje)
+extra = egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje)
 lib_link2 += $(fortran_libs)
 interface_incl = -I. -I$(HEN_HOUSE)lib$(DSEP)$(my_machine) -I$(egs_interface)
 
 include $(SPEC_DIR)egspp_libs.spec
 
 $(DSO2)$(library).$(obje): $(library).cpp $(library).h
-
-egs_advanced_application_lib_$(my_machine).$(obje): $(dep_advanced_application)
-	$(CXX) $(interface_incl) $(DEF1) $(DEF_USER) $(opt) -c $(COUT)$@ $<
 
 egs_interface2_$(my_machine).$(obje): $(EGS_INTERFACE)egs_interface2.c $(common_h_files1) array_sizes.h
 	$(CXX) $(interface_incl) $(DEF1) $(DEF_USER) $(opt) -c $(COUT)$@ $<
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
@@ -23,7 +23,7 @@
 #
 #  Author:          Ernesto Mainegra-Hing, 2018
 #
-#  Contributors:
+#  Contributors:    Blake Walters, 2021
 #
 ###############################################################################
 
@@ -34,56 +34,17 @@ include $(SPEC_DIR)egspp_$(my_machine).conf
 
 DEFS = $(DEF1) -DBUILD_RADIATIVE_SPLITTING_DLL
 
-egs_interface = $(HEN_HOUSE)interface
-EGS_INTERFACE = $(egs_interface)$(DSEP)
-
-ABS_EGSPP = $(HEN_HOUSE)$(EGSPP)$(DSEP)
-
-#required mortran files
-CPP_SOURCES = $(EGS_SOURCEDIR)egsnrc.macros $(MACHINE_MACROS) \
-            $(EGS_SOURCEDIR)pegs4_macros.mortran \
-            $(EGS_INTERFACE)egs_c_interface2.macros \
-            egs_radiative_splitting.macros \
-            $(MACHINE_MORTRAN) $(EGS_SOURCEDIR)egs_utilities.mortran \
-            $(EGS_INTERFACE)egs_c_interface2.mortran \
-            $(EGS_SOURCEDIR)transportp.macros \
-            $(EGS_SOURCEDIR)get_inputs.mortran\
-            $(EGS_SOURCEDIR)get_media_inputs.mortran \
-            $(EGS_SOURCEDIR)pegs4_routines.mortran \
-            $(EGS_SOURCEDIR)egsnrc.mortran
-
 library = egs_radiative_splitting
 lib_files = egs_radiative_splitting
 my_deps = $(common_ausgab_deps)
-extra_dep = $(addprefix $(DSOLIBS), $(my_deps)) egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje)
-extra = egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje)
-lib_link2 += $(fortran_libs)
-interface_incl = -I. -I$(HEN_HOUSE)lib$(DSEP)$(my_machine) -I$(egs_interface)
+extra_dep = $(addprefix $(DSOLIBS), $(my_deps))
 
 include $(SPEC_DIR)egspp_libs.spec
 
-$(DSO2)$(library).$(obje): $(library).cpp $(library).h
+INC2 = -I$(IEGS2) -I..$(DSEP)..
 
-egs_interface2_$(my_machine).$(obje): $(EGS_INTERFACE)egs_interface2.c $(common_h_files1) array_sizes.h
-	$(CXX) $(interface_incl) $(DEF1) $(DEF_USER) $(opt) -c $(COUT)$@ $<
-
-egsnrc_$(my_machine).$(obje): egsnrc_$(my_machine).F array_sizes.h
-	$(F77) $(FC_FLAGS) $(FDEFS) -c $(FOUT)$@ $<
-
-egsnrc_$(my_machine).F: $(CPP_SOURCES)
-	@echo "Mortran compiling EGSnrc sources ..."
-	@$(MORTRAN_EXE) -s -i -d $(MORTRAN_DATA) -f $(CPP_SOURCES) -o7 $@ \
-           -o8 $(@:.F=.mortlst)
+$(make_depend)
 
 test:
 	@echo "common_h2: $(common_h2)"
 	@echo "extra_dep: $(extra_dep)"
-	@echo "CPP_SOURCES: $(CPP_SOURCES)"
-	@echo "interface_incl: $(interface_incl)"
-	@echo "extra: $(extra)"
-	@echo "lib_objects: $(lib_objects)"
-	@echo "lib_link1: $(lib_link1)"
-	@echo "lib_link2: $(lib_link2)"
-	@echo "def1: $(DEF1)"
-	@echo "def_user: $(DEF_USER)"
-	@echo "fc_flags: $(FC_FLAGS)"
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
@@ -93,4 +93,6 @@ test:
 	@echo "lib_objects: $(lib_objects)"
 	@echo "lib_link1: $(lib_link1)"
 	@echo "lib_link2: $(lib_link2)"
-
+	@echo "def1: $(DEF1)"
+	@echo "def_user: $(DEF_USER)"
+	@echo "fc_flags: $(FC_FLAGS)"
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/Makefile
@@ -34,14 +34,47 @@ include $(SPEC_DIR)egspp_$(my_machine).conf
 
 DEFS = $(DEF1) -DBUILD_RADIATIVE_SPLITTING_DLL
 
+egs_interface = $(HEN_HOUSE)interface
+EGS_INTERFACE = $(egs_interface)$(DSEP)
+
+ABS_EGSPP = $(HEN_HOUSE)$(EGSPP)$(DSEP)
+
+#required mortran files
+CPP_SOURCES = $(EGS_SOURCEDIR)egsnrc.macros $(MACHINE_MACROS) \
+            $(EGS_SOURCEDIR)pegs4_macros.mortran \
+            $(EGS_INTERFACE)egs_c_interface2.macros \
+            $(EGSPP_USER_MACROS) \
+            $(MACHINE_MORTRAN) $(EGS_SOURCEDIR)egs_utilities.mortran \
+            $(EGS_INTERFACE)egs_c_interface2.mortran \
+            $(EGS_SOURCEDIR)transportp.macros \
+            $(EGS_SOURCEDIR)get_inputs.mortran\
+            $(EGS_SOURCEDIR)get_media_inputs.mortran \
+            $(EGS_SOURCEDIR)pegs4_routines.mortran \
+            $(EGS_SOURCEDIR)egsnrc.mortran
+
+dep_advanced_application = $(ABS_EGSPP)egs_advanced_application.cpp \
+        $(ABS_EGSPP)egs_advanced_application.h array_sizes.h \
+        $(ABS_EGSPP)egs_input.h $(ABS_EGSPP)egs_base_source.h \
+        $(ABS_EGSPP)egs_object_factory.h $(ABS_EGSPP)egs_timer.h \
+        $(ABS_EGSPP)egs_application.h $(ABS_EGSPP)egs_run_control.h
+
 library = egs_radiative_splitting
 lib_files = egs_radiative_splitting
 my_deps = $(common_ausgab_deps)
-extra_dep = $(addprefix $(DSOLIBS), $(my_deps))
+extra_dep = $(addprefix $(DSOLIBS), $(my_deps)) egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje) egs_advanced_application_lib_osx3.o
+extra = egsnrc_$(my_machine).$(obje) egs_interface2_$(my_machine).$(obje) egs_advanced_application_lib_osx3.o
+lib_link2 += $(fortran_libs)
+interface_incl = -I. -I$(HEN_HOUSE)lib$(DSEP)$(my_machine) -I$(egs_interface)
 
 include $(SPEC_DIR)egspp_libs.spec
 
 $(DSO2)$(library).$(obje): $(library).cpp $(library).h
+
+egs_advanced_application_lib_$(my_machine).$(obje): $(dep_advanced_application)
+	$(CXX) $(interface_incl) $(DEF1) $(DEF_USER) $(opt) -c $(COUT)$@ $<
+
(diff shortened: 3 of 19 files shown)


blakewalters:
Plots showing results of some initial tests in which I compared spectra using DBS in egs++, BEAMnrc and beampp in 2 different applications:

1.  EX16MVp

EX16MVp_DBS_comparison_spectra

2. A 450 kVp X-ray tube (spectra scored in vacuum)

i) With bound Compton (i.e. not using SmartCompton in DBS):

450kV_rad_splitting_test_in_vac

ii) Without bound Compton (i.e. using SmartCompton in DBS):

450kV_rad_splitting_test_in_vac_smart_compt

From 2.ii, it's clear that using SmartCompton in the egs++ DBS algorithm has introduced some bias.  Currently, the algorithm uses the same SmartCompton implementation as beampp.  I plan to replace this implementation with the one found in BEAMnrc and see if the difference persists.

ftessier:
This is awesome @blakewalters. I move to merge this in develop soon, so that we have a lead testing time in the field before the next release! Let's review this on the short term!

blakewalters:
Update 13/05/2023: After implementing BEAMnrc's SmartCompton algorithm (translated into C++) in the egs_radiative_splitting's DBS routine, agreement between BEAMnrc and egs++ in the 450 kVp tube case shown above is inching ever closer:

450kV_rad_splitting_test_in_vac_smart_compt_12052023

In fact, at this point we might say there's no significant difference.  

Continued testing is needed, though, given that, during one of these 450 kVp validation runs, I noted a phat photon crossing the scoring plane with r < DBS splitting radius.

blakewalters:
Update 07/07/2023:

After finding/fixing the bug in the SmartCompton method which was causing the odd high weight photon to be scored inside the splitting field (see update from 13/05/2023 above), 450 kVp X-ray photon spectra calculated using the DBS routine implemented in egs_radiative_splitting both with and without SmartCompton look to be in good agreement as shown in the figure below.

450kV_rad_splitting_test_in_vac_smart_compt_2

This is borne out in a plot of the fractional difference between egs++ and BEAMnrc spectra which also shows a significant difference (>5% at the extreme ends of the spectrum) between BEAMnrc and egs++ results.

450kV_rad_splitting_test_in_vac_smart_compt_2_diff

At this point, I can't say much about the source of this difference except that it is most likely not coming from egs++ DBS's handling of Compton events!  

I think it might be worth repeating the egs++ simulations with UBS just to confirm that the difference isn't due to any slight differences in geometry between BEAMnrc and egs++.

ftessier:
Wonderful @blakewalters!

blakewalters:
Update 09/15/2023:

Implementation of beampp's SmartCompton algorithm in egs++ (as opposed to the BEAMnrc SmartCompton algorithm from the previous update) results in similar differences with BEAMnrc:

450kV_rad_splitting_test_in_vac_smart_compt_6_diff_bpp_sc

But when standard Compton is used in egs++ these differences disappear:

450kV_rad_splitting_test_in_vac_smart_compt_5_diff_nosc

This implies to me that there's a general issue with the way I'm implementing SmartCompton in egs++.  If anyone has a chance to look at this portion of the code, please tell me if you see something glaring!

In the meantime, I'm planning to test elements of the SmartCompton routine that beampp and BEAMnrc have in common (e.g., creation of a single phat photon directed away from the field and creation of the single high-weight compton electron) in the hopes of isolating the source of this difference.

mainegra:
@blakewalters great job! Is the above the current status?

blakewalters:
That's where it's at, @mainegra.  My next debugging step will be to see if there's something going on with the energies of the split brem photons.  These energies are determined in a separate function after the interaction has been split and the photons generated.

blakewalters:
**Update 31/05/2024:**

There've been a couple of major fixes in the DBS splitting routine over the past couple of months.  Most significant of these stems from the realization that nonphat photons about to undergo Compton, photoelectric, pair or Rayleigh events should be subject to Russian Roulette--and made phat if they survive--prior to interacting.  This is a key element of DBS, and I can't believe I didn't notice this omission in my egs++ implementation before!  In addition, however, there was a bug in the function, doUphi21, determining angles for Rayleigh scattered particles as well as a C++-->Fortran indexing issue in the getBremsEnergies() function calculating energies for brem photons generated using doSmartBrems, among other bugs.

I've also since moved to doing BEAMnrc/egs++ comparisons using transmission photon spectra instead of reflection spectra to reduce the number of potential discrepancies between BEAMnrc-simulated and egs++-simulated source/target geometries.  

Here's the result of a comparison of 450 kVp W spectra with relaxation effects turned on.  In this case, bound Compton effects are also simulated and, thus, smartCompton is not used:

**Spectra in 10 cm vaccum (within r=10 cm splitting field):**

450kVp_split_test_relax_on_try2

**Fractional difference between BEAMnrc and egs++:**

450kVp_split_test_relax_on_diff_try2

In this case, I'm pretty confident of the match between BEAMnrc and egs++ spectra, although this could be verified to < 0.5% by running more histories.

Below is the result for the same case with relation effects off.  Here, bound Compton is turned off and, by default, smartCompton routines are used in both BEAMnrc and egs++:

**Spectra in 10 cm vacuum (within r=10 cm splitting field):** 

450kVp_split_test_relax_off_try2

**Fractional Difference between BEAMnrc and egs++**

450kVp_split_test_relax_off_diff_try2

Although we might say the two spectra agree within uncertainty, there is a drift in the egs++ spectrum from lower to higher values with increasing energy (see fractional differences) that's a little troubling.  This needs to be verified by running more histories, but if it persists, then it is almost certainly due to issues in the egs++ smartCompton method.  Currently, this method uses the beampp implementation of smartCompton, which differs slightly from the BEAMnrc implementation.

HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/array_sizes.h
```diff
@@ -0,0 +1,52 @@
+/*
+###############################################################################
+#
+#  EGSnrc tutor7pp application array sizes headers
```
ftessier, line 4: Update the file header to change `tutor7pp` to `radiative splitting`.

blakewalters:
### Update 06/07/2024

Continuing with the tests from the last update, here's the fractional difference between BEAMnrc and egs++ photon spectra with relaxation effects off when smartCompton is not used in egs++ DBS:

450kVp_split_test_relax_off_diff_nosc_try2

which shows good agreement and seems to confirm that the bias seen in the previous comparison with relaxation off is due to the current smartCompton implementation in egs++ DBS.

As mentioned above, the smartCompton implementation in egs++ is currently based on the one in beampp, and so there are a couple of possible next steps:
1. Comb through the egs++ smartCompton implementation for bugs
2. Implement smartCompton following BEAMnrc's logic  (I've tried this before with previous iterations of egs++ DBS.  However, with recent fixes to the algorithm, it bears repeating...again!)

blakewalters:
### Update 07/06/2024 later pm

After (re)implementing the BEAMnrc smartCompton algorithm in egs++ DBS (and noting some errors I'd made in my original implementation of this along the way), we get good agreement between BEAMnrc and egs++ spectra with relaxation effects off:

450kVp_split_test_relax_off_diff_beamsc_try2

We have to decide, then, if we're going to go with our original vision of having more than one flavour of egs++ DBS (one modeled after BEAMnrc, one modeled after beampp, another modeled after?).  If so, then we'll have to revisit and debug the beampp smartCompton algorithm.  For now, though, I think egs++ DBS is ready for some more intense testing!

blakewalters:
Update 07/04/2024:

Results of recent tests comparing BEAMnrc/DBS and egs++/DBS in the case of a 225 kVp e- beam incident on a reflection anode (1 mm W, Cu backing, anode angle = 4.92 deg.).  Note that I've used a pencil beam of e- (parallel beam incident at a point) to try to mitigate any differences due to source geometry.  For the BEAMnrc simulation, this meant using source 13 (e- beam pointed in the -X direction at the face of the anode) with rectangular dimensions (0.0002 x 0.0002 cm, i.e. really small).

Here are the photon spectra at SSD=10 (splitting SSD) and r=10 cm (splitting radius).  The DBS splitting number in all cases is 10,000.  I've also included results for egs++ with UBS (splitting no. = 100) and egs++/DBS with smartBrems turned off:

225kVp_refl_split_test_pspect

With the exception of the peak at ~60 kV, where egs++/DBS looks a little low, a visual comparison indicates good agreement.  The fractional difference between BEAMnrc and egs++ results, however, tells a slightly different story:

225kVp_refl_split_test_pspect_diff

Here we see that egs++/DBS is up to 2.5% greater than BEAMnrc at low energies, egs++/DBS with smartBrems is up to ~1% high at energies > 50 kV.  egs++/UBS and egs++/DBS without smartBrems are up to ~0.5% high for energies > 50 kV, although better stats are required to determine the significance of this difference.  In the case of egs++/DBS with smartBrems, though, it seems clear that some bias is introduced.

Next steps:
1.  Simulate to better uncertainty to confirm differences
2. Comb through the smartBrems routine once again to check for possible differences between egs++ and BEAMnrc implementation and, possibly, reimplement it in egs++ to match that in BEAMnrc

blakewalters:
Update 08/02/2024:

After running the 225 kVp reflection anode results down to ~0.1% uncertainty, a plot of the difference between BEAMnrc/DBS and egs++/DBS photon fluence spectra confirms the differences of ~1% seen in the last update.  The "good" news is that, after running more histories in the egs++/UBS (100x splitting) case, the difference between BEAMnrc and this case is confirmed as well:

225kVp_refl_split_test_pspect_ubs_dbs

Moreover, egs++/DBS and egs++/UBS results are within uncertainty, leading one to suspect that differences between egs++ and BEAMnrc are due to differences in modeling the X-ray tube geometry (I'm assuming, of course, that my double checking that all MC transport parameters are identical doesn't require triple checking!).

A plot of the difference in photon fluence along the anode-cathode axis (incident e- in the -X direction) might indicate a slightly smaller anode angle in the BEAMnrc simulation:

225kVp_refl_split_test_pxflu
225kVp_refl_split_test_pxflu_diff

On the other hand, this may also indicate slight differences in the effective thickness of the W target layer.

More investigation is required to track potential geometric differences down.  However, it now seems they're adding an unnecessary layer of complication in comparisons between BEAMnrc/DBS and egs++/DBS when a reflection anode is used.  At this point, then, I suggest going ahead with a comparison at 225 kVp using a transmission target.

blakewalters:
**Update Oct. 2, 2024**

Further simulations of the 225 kVp reflection anode (W thickness=0.1 mm, angle=4.92 degrees) from the previous update have confirmed consistent differences on the order of 1% between BEAMnrc and egs++ photon spectra, regardless of whether egs++ is used with UBS or DBS.  This implies some slight difference in the way the anode geometry is defined (maybe traceable to differences in boundary tolerance between the two codes?) and indicates, at least for now, a way forward is to compare results with egs++_DBS to those with egs++_UBS.  

The plots below show the fractional difference between photon spectra for the reflection anode generated using egs++_DBS and egs++_UBS.  Note that, instead of analyzing phase space data as I did previously, I've scored spectra using the egs_fluence ausgab object.  This facilitates running more histories to improve the statistics of the comparison and eliminates the requirement for storing and analyzing phase space data.  The first plot shows the fractional difference with atomic relaxations turned off (i.e., smartCompton used), while the second includes atomic relaxations (smartCompton not used) and electron impact ionization:

225kVp_refl_split_test_pspect_diff_egspponly

225kVp_refl_split_test_pspect_relaxon_diff_egspponly

With relaxation off, spectra agree to within ~0.05%, while with atomic relaxation on there is a significant difference of 0.2% at the photon peak at ~62 keV.  

For reference, the plot below shows photon spectra with relaxation on with and without electron impact ionization (eii).  Note that any differences between egs++_UBS and egs++_DBS are too small to see at this scale:

225kVp_refl_split_test_pspect_relaxon_eiionoff_egspponly

I then turned eii off and, once again, egs++_DBS agrees with egs++_UBS to within ~0.05%:

225kVp_refl_split_test_pspect_relaxon_eiioff_diff_egspponly

It seems, then, that eii results in additional fluorescent photons that aren't handled correctly in egs++_DBS.  Alternatively, it may magnify the effects of a bug in the handling of these photons whether eii is on or not.

The adventure continues!  Any and all ideas about what might be going on with eii are welcome.

rtownson:
@blakewalters In the EII sampling routine it copies the particle properties from the parent electron like this (line 6174 in egsnrc.mortran):
```
$TRANSFER PROPERTIES TO (np) FROM (np_save);
```
And this is copied to relaxation secondaries. For the fluorescent photon should the weight be adjusted? I'm not quite sure the solution but certainly this is the spot to look. Probably at the latch and what it should be for fluorescence v.s. auger electrons.

blakewalters:
Thanks, @rtownson!  I'll start there.

blakewalters:
**Update Oct 9, 2024**

Using the egs_fluence ausgab object together with egs_beam_source, I've been able to use egs++ to calculate BEAMnrc photon spectra down to ~0.05% uncertainty.  As a result, we can more precisely compare BEAMnrc and egs++ results.  

The plot below shows the fractional difference between 225 kVp reflection spectra with relaxation effects on and eii on where now egs++ spectra with DBS (egs++_DBS) and UBS (egs++_UBS) are compared against the BEAMnrc spectrum with DBS (ssd=10 cm, r=10 cm, nsplit=10000).  

225kVp_refl_split_test_pspect_diff_2

The plot clearly shows the ~1% difference between egs++ and BEAMnrc consistent with the previous update and which we surmise is due to slight differences in anode geometry.  

Note, however, that egs++_UBS is ~0.2% higher than BEAMnrc at the peak at ~62 keV (this shows up as a ~0.2% dip in fractional difference relative to BEAMnrc).  This is similar to the fractional difference between egs++_UBS and egs++_DBS at this energy, and we may cautiously advance that the treatment of fluorescent photons resulting from eii events in egs++_DBS is consistent with that in BEAMnrc with DBS (or BEAMnrc_DBS, if that's easier).  This makes sense, given that egs++_DBS is heavily based on the DBS implementation in BEAMnrc.

The plot below shows a comparison of 225 kVp spectra using a transmission target, where no differences between BEAMnrc and egs++ geometries are expected:

225kVp_trans_split_test_pspect_diff_2

In this case, egs++_DBS and BEAMnrc_DBS show agreement to within ~0.05% with the exception of the lowest energy bins, where uncertainties are > 1%, while the spectrum calculated using egs++_UBS is again ~0.2% higher at the ~62 keV peak (showing up as a ~0.2% dip in the above plot).  We also note a ~0.1% difference between egs++_UBS and BEAMnrc at ~190 keV which is likely within uncertainty.  (Note: I'm going to regenerate this plot with more precise spectra)

We can hesitatingly conclude, then, that treatment of fluorescent photons resulting from eii is the same in egs++_DBS and BEAMnrc_DBS.  The question now is: Why are they both ~0.2% low at the K-shell peak(s)?

blakewalters:
@mainegra, @ftessier, @rtownson:

Continuing on with testing of the DBS implementation in EGSRadiativeSplitting, I've been mulling over a couple of questions:

1. How do we implement DBS so that it will work with transformed sources?  Right now, the implementation assumes the source is at Z=0, with the SSD and splitting radius defined relative to this.  It would be easy to add more inputs to the AO to allow the user to translate/rotate the splitting field, but would it be more convenient and useful to somehow be able to "attach" a DBS splitting field to a particular source?  That way, when a source is transformed, the relevant DBS splitting parameters are automatically transformed along with it.  I guess an option would be to implement an "egs++_simulation_source" (something we might want to do anyway), in which case the DBS AO would just be part of the source simulation that can then be transformed in any way the user wishes in the second-stage simulation.  Come to think of it, that might be the best way to accomplish this.  Thoughts?

2. What about BCSE?  I'm assuming we ultimately want to implement BCSE along with DBS, but this will require hooking into MORTRAN macros inside of EGSnrc.  It's currently fairly straightforward to access subroutines inside EGSnrc, but I don't know if macros can be modified, short of making changes to interface coding (e.g., egs_c_interface2.macros).  Let me know if any of you know another way!

mainegra:
- Adding enough info during input to allow for arbitrary splitting field definition would be my choice 
- BCSE is implemented in egs_brachy,  perhaps a good place to start?

rtownson:
The DBS should not be tied to the source or the way it is transformed, it should be completely independent IMO. 

Ideally, it should let you set filter where it takes place by `geometry`, `include regions` and `include media` like in egs_isotropic_source. Then use a generic shape (not just a circle?) with a transform to denote the field, like the target shape for `egs_collimated_source`.

ftessier:
I agree with @mainegra and @rtownson, except that I feel there should be an option to transform the target shape with the source. That is, when dealing with an `egs_transformed_source`, one could opt to apply the transformation to the target shape as well. Seems simple enough to implement, a flag and corresponding conditional to multiply by `T`, the transformation. Perhaps this ought to be a generic source option, useful for `egs_collimated_source` as well, for example.

rtownson:
> I agree with @mainegra and @rtownson, except that I feel there should be an option to transform the target shape with the source. That is, when dealing with an `egs_transformed_source`, one could opt to apply the transformation to the target shape as well. Seems simple enough to implement, a flag and corresponding conditional to multiply by `T`, the transformation. Perhaps this ought to be a generic source option, useful for `egs_collimated_source` as well, for example.

I thought this was already essentially the case. Since `egs_transformed_source` transforms whatever particle you get out of the original source, which already had its direction towards the target. Thus its direction toward the target gets transformed to point toward a new target, as per the rotation in the transformation. Or are you imagining something else?

ftessier:
Yes, of course you are right! What I had in mind is that the DBS target shape would be defined within the source, not within the ausgab object. Or perhaps one could define a completely decoupled DBS field in the ausgab object? I can think of scenarios where either makes sense.

blakewalters:
@ftessier and @rtownson, this is why I was thinking that the shared library source might be the best way to take care of this: DBS is defined as an ausgab object in the source simulation and then that source can be duplicated/transformed in the 2nd stage simulation without having to worry about defining multiple DBS AO's.  Kind of like what already happens when we use BEAMnrc to simulate an X-ray tube and then use this as a source in egs++.  Of course, there may be issues with optimizing DBS parameters across multiple transformed sources.

ftessier:
Now that I think about it, would it makes sense to implement DBS as a `source` feature with a target shape, instead of an ausgab object? Perhaps I am going astray here: DBS is implemented as a region-based VRT, so to speak, it does not fit in the source. But perhaps there is a way to adjust our thinking and make it a source feature? 🤔

blakewalters:
### Update 12/03/2024:

### Testing with IBRDST=1

Continuing testing of the egs++ DBS implementation, I was interested in seeing its accuracy when the full Koch-Motz (KM) distribution is used for bremsstrahlung angular sampling (IBRDST=1).  This is done by setting "brems angular sampling=KM" in the transport parameter input block.  Note that comparisons above only the leading term of the KM distribution (IBRDST=0), or "brems angular sampling=simple" in the transport parameter inputs.

The plot below shows the fractional difference in the 225 kVp transmission photon spectra between the baseline case using the "simple" brems angular distribution and when the full KM expression is used.  BEAMnrc was used to generate both spectra.

225kVp_trans_split_test_relaxon_simple_km_diff

Use of the full KM distribution results in differences in the energy spectrum ranging from -0.2% at 120 keV to 0.8% at 225 keV.  Given that egs++ with DBS agrees with BEAMnrc with DBS to within <~0.05%, then, potential issues arising when IBRDST=1 is used with the egs++ DBS routine should be visible in a comparison with BEAMnrc (also using IBRDS=1).

The plot below shows the fractional difference in transmission photon spectra generated using egs++ with DBS and BEAMnrc with DBS where the full KM distribution is used for brems angular sampling in both simulations (black).  The plot also shows the fractional difference when egs++ with UBS is used (red line).

225kVp_trans_split_test_pspect_diff_ibrdst1

This shows that, even using the full KM distribution, egs++ with DBS agrees with BEAMnrc with DBS to within ~0.05%.  We also note that the ~0.2% difference at the K-shell peak (~60 keV) between BEAMnrc and egs++ with UBS when electron impact ionization (EII) is on persists with IBRDST=1.

### A side issue: Comparison of egs++ with UBS and BEAMnrc with UBS against BEAMnrc_DBS:

In making these IBRDST=1 comparisons and attempting to locate the source of the discrepancy between egs++_UBS and BEAMnrc_DBS (and egs++_DBS) at ~60 keV, I also introduced into the comparison runs with BEAMnrc using UBS (nsplit=100) and found some "interesting" results overall.

Both figures below show the same plot of fractional difference between the photon spectra generated with egs++_UBS and BEAMnrc_UBS and our baseline BEAMnrc_DBS case at two different scales.  Simulations with UBS were run with charged particle Russian Roulette on and off.  In all cases, I reverted to the simple brems angular distribution (IBRDST=0).

225kVp_trans_split_test_pspect_diff_ubs_beamnrcdbs_fullscale

225kVp_trans_split_test_pspect_diff_ubs_beamnrcdbs

The full-scale plot (upper figure) shows significant differences between BEAMnrc with UBS and Russian Roulette on (green line) and BEAMnrc with DBS at energies < 50 keV, with differences of ~40% at energies <= 20 MeV.  These large differences disappear when Russian Roulette is turned off (magenta line).  The close-up (lower figure) shows that the ~0.2% difference between egs++ with UBS and BEAMnrc with DBS at the K-shell persists when Russian Roulette is turned off (red and black lines).  Moreover, we don't see this difference with BEAMnrc_UBS when Russian Roulette is off (magenta line), a further indication of an issue with the UBS implementation in egs++ (with EII on).

Therefore, while not distracting ourselves too much from the original egs++ DBS implementation, these tests have given rise to a couple of UBS-related questions that need to be addressed:

1. What is the source of the large discrepancy at low energies with BEAMnrc UBS when Russian Roulette is on?
2. What is causing the discrepancy at ~60 keV in the egs++ implementation of UBS (with EII on)?

ftessier:
> 1. Within the AO, implement the option to transform (rotate/translate) the DBS splitting cone.
>
> 2. Also implement an option to read in the (active) source transform so that the the DBS splitting cone can also either: a) move with a dynamic source or b) be applied to the current source in a source collection.  This transform would be applied on top of any transform input in the AO itself (i.e., step 1).

Why are we considering a cone, instead of a general target shape for DBS?

blakewalters:
@ftessier, only because the smartBrems and smartCompton routines in DBS pre-calculate the number of split particles to generate based on the photon angular distributions for the respective interactions and a splitting field with a circular base (i.e. a cone).  I guess we could use a rejection technique to make the shape of the field more general, and ultimately we may want to try this, but it doesn't seem like a priority at this point in the implementation.

ikawrakow:
>  I guess we could use a rejection technique to make the shape of the field more general, and ultimately we may want to try this, but it doesn't seem like a priority at this point in the implementation.

It is obvious, right? And so elegant - just do UBS and Russian Roulette away the photons not going towards the FOI. What was Kawrakow thinking?

It was 20 years ago, but if IIRC, the speed gain from estimating the probability to emit bremsstrahlung or Compton-scatter into the FOI was very significant. The probability estimate used in the BEAMnrc implementation is just that - an estimate. It overestimates, and the overestimation can be quite significant in some cases. When I was implementing the treatment head simulation code for ViewRay I was able to derive a better estimate (compared to what is in BEAMnrc) and that sped it up by a factor of 2, IIRC. If you want to give the user the ability to define arbitrary FOIs, that will come at very significant efficiency cost.

blakewalters:
@kawrakow, you really are at large (but hopefully not at loose ends)!  Yes, I suspected it would be a nontrivial matter to give the user the option to define an arbitrarily-shaped splitting field.  Thanks for confirming this!  And that the flexibility to define a sawtooth splitting field or one that says "Bernie 2028" is really not worth the hit to efficiency.

blakewalters:
**Update 01/08/2025**

Recently, the ability to apply an affine transform to the DBS splitting cone (defined by the splitting ssd and radius, fs, of the splitting field) has been implemented.  This allows the user to user to orient the splitting geometry to match that of the e- beam and bremsstrahlung target.

To put this functionality through its paces, I evaluated the difference between the 225 kVp photon transmission spectra generated with the transforms illustrated in the figure below (T1-T3) and that with no transform applied to the splitting geometry (T0).  In each transformed case, the transform applied to the DBS splitting cone matches an equivalent transform of the incident e- beam/target geometry.  In all cases, DBS splitting parameters are: nsplit=1000, ssd=10 cm, fs=10 cm.

egspp_dbs_transform_fig

In T1, the target geometry/splitting cone has been translated +10 cm in both X- and Y-directions; in T2, the geometry has been rotated 45-degrees counterclockwise about the Y-axis; and in T3, the geometry has first been rotated 180-degrees about the Y-axis then translated -10cm in the Z-direction.

Fractional differences between the spectrum with no transform and spectra with transforms T1-T3 are shown in the figure below:

225kVp_trans_split_test_pspect_diff _transforms

Note that I've exaggerated the -ve portion of the vertical axis to show that differences in low energy (<20 keV) components of the spectra are within uncertainty, although the photon fluence in this energy range is a factor of 2-3 lower than elsewhere in the spectra.

With the exception of low energy photons, this plot shows agreement to within ~0.05% between results with transforms applied to the splitting cone and the untransformed case, thus giving us confidence that transforming the splitting cone does not alter the results.  Even so, some questions remain:
1. In order to obtain approximately the same uncertainty with T2 (45-degree rotation of the geometry about the Y-axis) as in the untransformed case, I had to increase the no. of incident histories by a factor of 4.  Why?
2. Despite the fact that spectra < 20 keV agree within uncertainty, why are the differences so large, and why are they consistently -ve (i.e. higher than in the untransformed case)?

I make no claim that these cases represent an exhaustive test of the effects of transforming the splitting geometry, and suggests for further testing are welcome!

blakewalters:
**Update 02/26/2025**

In the 3 most recent commits to this PR, I've implemented a charged particle splitting option in egs++ DBS.  Inputs for this option existed prior to this, but were not enabled.

Charged particle splitting can be used when one is interested in recovering some of the contaminant electrons which are otherwise subject to Russian Roulette when DBS is used in simulating photon beams.  This then allows the user to examine quantities such as the spectrum and fluence of these electrons, contaminant dose, etc.

In BEAMnrc DBS, e-/e+ splitting is done at a horizontal region boundary, the "splitting plane", within a component module (CM--always FLATFILT).  Charged particles crossing this region boundary are split nsplit (i.e. the bremsstrahlung splitting no.) times with their weight reduced by a factor of 1/nsplit, with an option to redistribute split particles evenly about a circle of radius sqrt(x^2+y^2), where x and y are the X- and Y-positions of the particle at the splitting plane prior to splitting.  This latter option is known as "radial redistribution" of split particles, and can improve spatial sampling of charged particles for beams with radial symmetry.  In addition, the user is also asked to input the Z-position of a Russian Roulette plane (`zrr`) below which Russian Roulette is no longer played with charged particles and Russian Roulette IS played with photons about to undergo photoelectric, pair production or Compton events.  These events are then split to generate nsplit (or nsplit*2 in the case of pair production) low-weight charged particles.  In general, `zrr` is < the Z-position of the splitting plane.  For more information on e-/e+ splitting, see the BEAMnrc Manual.

The implementation of charged particle splitting in egs++ DBS is similar to that in BEAMnrc DBS.  Inputs are also similar, with the difference that the splitting plane is not restricted to a region boundary within a CM but can be any region or group of regions in the geometry.  On entering this(these) region(s), phat charged particles are immediately split and possibly radially redistributed about the Z-axis.  This offers greater flexibility, but may also prove a pitfall, in that it is difficult to imagine a situation where the splitting plane should not be perpendicular to the beam axis and encompass the entire beam field (at that plane).  

Also, in egs++ the charged particle splitting plane is subject to any transform applied to the DBS splitting cone (see above).  Thus, radial redistribution of split particles about the transformed plane would actually place particles outside their original (untransformed) region.  To avoid this possibility, if the radial redistribution option is used, we always redistribute split particles evenly about the Z-axis in the original (untransformed) geometry.  This mens that split particles may not end up being evenly distributed about the beam axis (which, presumably, is the same as the transformed axis of the splitting cone), and so we advise against radial redistribution of split charged particles in cases where you are also applying a transform to the DBS splitting cone.

Onward, then, to testing of the e-/e+ splitting option, in which I repeat many of the tests above for the 225 kVp transmission beam but now examine both photon and electron spectra at the splitting ssd.  In the case of electron spectra, I've increased the field radius considered to 20 cm (splitting radius=10 cm) in an attempt to improve the statistics of the e- spectrum (and recognizing that the fluence of contaminant e- is not restricted to the DBS splitting field).

More to come, stay tuned!

blakewalters:
**Update 02/27/2025**

...continued from previous update.

As mentioned above, tests of e-/e+ splitting use the same 225 kVp photon transmission setup used in earlier tests, with a point source of e- incident on an infinite W slab (target) of thickness 0.01 cm.  Below the target is vacuum. 

The first comparisons between BEAMnrc and egs++ DBS with the e-/e+ splitting option involved placing the splitting plane at the bottom of the W target (at Z=0.01 cm) and the Russian Roulette plane (zrr) at 1 cm (i.e. in vacuum below the target).  Since interactions resulting in charged particles (pair production, photoelectric, Compton) are only split below zrr, this ensures that all charged particles reaching the splitting plane are phat and, thus, serves as a test of splitting (and radial redistribution) at this plane.

The figure below shows the e- spectra 10 cm below the target (i.e. at the DBS ssd) in a circle of radius 20 cm:

225kVp_trans_split_test_esplit_zrr1cm_electron

In BEAMnrc and egs++, charged particle splitting at the splitting plane has resulted in improved statistics in the e- spectra, and superficial agreement between BEAMnrc and egs++ e- spectra can be seen.  A plot of particle weights (not shown) indicates that all e- reaching the scoring plane have weight=1/1000, as expected. 

The plot below shows the fractional difference between the BEAMnrc and egs++ e- spectra in this case:

225kVp_trans_split_test_esplit_zrr1cm_electron_diff

This shows agreement between BEAMnrc and egs++ e- spectra within an uncertainty of 2-10%.

The figure below shows the fractional difference between BEAMnrc and egs++ photon spectra (scored in a field with r=10 cm at the ssd) for this e-/e+ splitting case:

225kVp_trans_split_test_esplit_zrr1cm_photon_diff

Similar to earlier comparisons without e-/e+ splitting, BEAMnrc and egs++ photon spectra agree to within ~0.05%

In order to make a more definitive statement about the agreement between BEAMnrc and egs++ e-spectra, I wanted to generate e- spectra with uncertainties < 0.1%.  I found that with zrr=1 cm, however, this required > 100x10^9 histories, which presented a challenge given the computing resources available.  With this in mind, I varied the value of zrr in an attempt to optimize e- statistics, ultimately arriving at zrr=0.005 cm (i.e., halfway through the target), which allowed me to generate spectra having < 0.1% uncertainty at their peak (kinetic energy=40 keV) using 40x10^9 histories.

The plot below shows the fractional difference between BEAMnrc and egs++ e- spectra with this new value of zrr:


225kVp_trans_split_test_esplit_zrr005cm_electron_diff
 
From this plot, we can infer that BEAMnrc and egs++ e- spectra agree to within < 0.5% over most of the energy range, certainly a stronger statement regarding agreement than we were able to make above.  

Comparing the photon spectra with this value of zrr (and no. of histories), we obtain the following:

225kVp_trans_split_test_esplit_zrr005cm_photon_diff

Which shows agreement within 0.02% but also shows the BEAMnrc photon spectrum to be consistently lower than the egs++ spectrum by on average 0.01%.  Observing this, I tumbled down a bit of a rabbit hole and went back to our original comparison of these photon spectra without e-/e+ splitting, noted that this 0.01% difference would not be observable given the uncertainties in those calculations, and so reran them with 40x10^9 histories (vs 13x10^9 histories used in the original comparison).  The resulting fractional difference between BEAMnrc and egs++ photon spectra is shown below:

225kVp_trans_split_test_pspect_diff_3

From this plot we see that, while the spectra agree within ~0.02%, the BEAMnrc photon spectrum is ~0.01% lower than the egs++ spectrum.  While this is consistent with the systematic difference between the photon spectra observed above with e-/e+ splitting on, we note that overall this systematic difference is smaller (and varies less) with e-/e+ splitting off.

Pinpointing possible sources of the difference in behaviour (i.e., the "difference in the fractional difference" between BEAMnrc and egs++ photon spectra) with e-/e+ splitting on vs e-/e+ splitting off will likely prove difficult.  One contributing factor may be the different target geometries used: In simulations with e-/e+ splitting, I modeled the W target using a circular slab with r=100 cm (i.e. a shape that could be modeled with a FLATFILT CM in BEAMnrc), while a 200x200 cm square slab was used in simulations without e-/e+ splitting.  Differences in photon transport near (and past) the outer boundaries of these two different shapes may have higher-order effects on the fractional difference between BEAMnrc and egs++ spectra.

Pursuing the possible effects of differing outer boundaries of the target slab a bit further, I plotted the fractional difference in photon spectra with and without e-/e+ splitting for both BEAMnrc and egs++.

Stay tuned...

blakewalters:
**Update 03/27/2025**

By a process of systematically omitting photon interactions that may be split when e-/e+ splitting is on--pair production, photoelectric, Compton, and Rayleigh scattering--from the simulations (luckily, this is fairly straightforward in the egs++ implementation of DBS!), it appeared that Rayleigh interactions could be the source of the difference between photon spectra with e-/e+ splitting on and off.  Indeed, the fractional differences between the egs++ photon spectra with e-/e+ splitting on and off both with and without Rayleigh scattering included in the simulations seem to bear this out.  Below is a plot showing these fractional differences, where the resolution of the photon spectra has been halved (i.e. 10 energy bins over the range 0-225 keV) in an attempt to reduce the simulation time required to obtain the uncertainties required to resolve these differences:

225kVp_trans_split_test_esplitonoff_raylonoff_diff

Here we see that, when Rayleigh splitting is on, the photon spectrum with e-/e+ splitting is systematically ~0.02% higher than that without e-/e+ splitting (consistent with our result from 02/27/2025).  With Rayleigh splitting off, however, the systematic aspect of the difference largely disappears, perhaps with an exception at higher energies, where significant differences are more difficult to resolve because of the higher uncertainties.

After my initial observation that the Rayleigh scattering implementation in egs++ DBS--which is essentially a copy of that in BEAMnrc DBS--may be the cause of the discrepancy between photon spectra with and without e-/e+ splitting, I did a fairly extensive debugging operation on this Rayleigh scattering implementation.  Detailed output of the various angular sampling routines (Rayleigh angular sampling followed by "adjustment" of scattering angles by a random azimuthal angles using the routine UPHI(2,1)), however, revealed no obvious reasons why the photon spectrum with e-/e+ splitting on and when Rayleigh scattering is simulated would systematically include more photons than that with e-/e+ splitting on.  I was, thus, left with the possibility that Rayleigh scattering is simply magnifying some other discrepancy resulting from e-/e+ splitting.

Still pursuing the idea of a bug in the egs++ DBS implementation of Rayleigh scattering, however, I plotted the fractional differences between BEAMnrc and egs++ photon spectra with no e+/e- splitting, comparing the differences with Rayleigh scattering on and off.  Results are shown in the plot below:

225kVp_trans_split_test_raylonoff_diff

Within uncertainty, this shows no significant systematic "difference between the differences" whether Rayleigh is included in the simulation or not, lending further credence to the possibility that Rayleigh is simply amplifying a discrepancy introduced by e-/e+ splitting.  To see if this also might be the case in BEAMnrc DBS, I plotted the fractional difference between BEAMnrc photon spectra with and without e-/e+ splitting when Rayleigh scattering is off and on.  Results are shown in the graph below:

225kVp_trans_split_test_esplitonoff_raylonoff_beamnrc_diff

This appears to show that the photon spectrum is systematically greater (i.e. more particles across all energies) with e-/e+ splitting when Rayleigh scattering is off, which is the opposite of the behaviour with egs++ DBS!  Uncertainties in these BEAMnrc results, however, are larger than those with egs++, making it difficult to draw definitive conclusions.

At this point, we cautiously surmise that e-/e+ splitting may be introducing a very small bias in the photon spectrum, the presence of which is rendered noticeable to within 0.01-0.02% when Rayleigh scattering is on.  More accurate simulations, however, will be required to: 1) confirm with certainty that this bias exists and 2) determine whether or not it also exists in BEAMnrc DBS.

I have already started rerunning the above egs++ simulations with 4x the original number of histories (40x10^9 histories in all) to answer questions 1) and 2) above.  These results will be posted here when available.  

Adventurous souls are also invited to look through the DBS e-/e+ splitting implementation in egs_radiative_splitting to see where there may be a bug that would introduce a 0.01-0.02% discrepancy in photon spectra.

In the meantime, testing of the e-/e+ splitting implementation on other fronts--such as confirming that transforming the DBS splitting cone does not change the e- and photon spectra collected in a plane perpendicular to the beam--continues...

blakewalters:
**Update 03/27/2025 pm**

Leaving behind the observed 0.01-0.02 % discrepancy between photon spectra with and without e-/e+ splitting (when Rayleigh scattering is on) for the moment, let us continue with testing of the e-/e+ splitting implementation under conditions of a transformed DBS splitting cone.  The 3 transforms tested were the same as those used to ensure agreement between photon spectra with transformed splitting cones without e-/e+ splitting and are shown in the update of 01/08/2025.  

Note that in all transformed cases, we have opted to not radially redistribute split charged particles about the Z-axis.  This is because, for reasons given above, radial redistribution is always done about the untransformed Z-axis which will violate radial symmetry about the beam axis (which the transformed DBS cone presumably tracks) in most cases.

Also note that spectra are always collected at a plane perpendicular to the axis of the transformed splitting cone (i.e., the scoring plane is transformed as well).

The figure below shows the fractional differences in e- spectra between the untransformed (T0) and transformed cases for the 3 transforms considered.

225kVp_trans_split_test_esplit_zrr005cm_electron_transformations_diff

The plot shows agreement to within 0.2-0.5% over most of the energy range.  Greater precision, though desirable, will be time consuming owing to the few contaminant e- generated per incident particle in this setup.

Tne plot below shows fractional differences between photon spectra in the untransformed (T0) and transformed cases:

225kVp_trans_split_test_esplit_zrr005cm_photon_transform_diff

which shows agreement to within ~0.01% for transforms T1 and T2 and agreement to within ~0.02% for T3.  Note that the reason for the slightly greater difference in the case of T3 is not known, however, it is still within uncertainty of the T0 case.

blakewalters:
**Update 03/28/2025**

This update revisits yesterday's comparison between photon spectra obtained using egs++ DBS with and without e-/e+ splitting with Rayleigh scattering on and off.  Recall that, with Rayleigh scattering on, e-/e+ appeared to introduce a ~0.01-0.02% increase in the photon spectrum.  A detailed look at the egs++ DBS Rayleigh scattering routine did not reveal any obvious bugs, leading me to conclude that Rayleigh scattering may, in fact, "amplify" a systematic difference introduced by e-/e+ splitting. 

Redoing the comparison using 4x the number of histories (i.e. uncertainties on photon spectra halved), we obtain the following plot of fractional difference:

225kVp_trans_split_test_esplitonoff_raylonoff_highres_diff

from which it is possible to conclude that any systematic difference introduced by e-/e+ with Rayleigh scattering on is at most ~0.01% and, moreover, is within uncertainty of any differences with Rayleigh scattering off.

My recommendations at this point are:
1) Continue/expand egs++/DBS testing to run comparisons with BEAMnrc/DBS over a number of energies in the keV - MeV range and try higher splitting numbers.  The current tests at 225 kVp with nsplit=1000 have proven valuable for debugging the DBS algorithm's handling of brems, photelectric, fluorescent, and, to a certain extent Compton and Rayleigh events, but have not tested annihilation or pair production.  Comparisons against BEAMnrc/DBS at different energies will change the subset of interactions tested and, indeed, at least one comparison in the MeV range is essential.  Also, simulations of kVp beams are optimized with splitting numbers on the order of 10,000-100,000.  These higher splitting numbers may amplify any bugs in the code that I've missed in the current tests.
2) Clean up the code.  There are many instances in the current egs++/DBS implementation of repeating code that could be put into functions, code that could be rationalized, variable names that could be made more explicit, etc.  Cleaning up the code in this way would make it easier to find potential bugs.
3) Test egs++/DBS in actual applications.  I'm currently using egs++/DBS to simulate a prototype kV irradiator and obtaining agreement with prior simulations where BEAMnrc/DBS was used to simulate the X-ray source.  Admittedly, I'm only running simulations to ~2% uncertainty, so the test isn't very exacting, and the more conditions under which this egs++/DBS implementation can be benchmarked the better.
4) Documentation.  I've left fairly extensive comments in the code, but comprehensive documentation that can then be extracted by doxygen is required!

PR #1005, opened 2023-06-07
Increase $MXBRSPLIT to 5000.

blakewalters:
Necessary for Lab 4.5, Question 4.5.1, but it's probably a good idea to increase this anyway.


rtownson (approved):

ftessier:
I perhaps agree with this change for EGSnrc, but that won't help this time around for the EGSnrc course, which is based off the `master` branch.

blakewalters:
@ftessier, didn't I issue the PR against the master branch?

ftessier:
@blakewalters yes, but we are not going to release a new commit on `master` right now. Those have to be handled with more attention than I have now, because of citations. I think we will release an update soon though, on account of #996. Don't know if that requires a `2023a` release, I don't think so since it is compilation related. Still, I ought to double-check.

rtownson (approved):

ftessier:
Tweaked the commit title.

Issue #1001, opened 2023-06-05
Split photons from RR-surviving e-/e+ in Radiation Splitting AO


mainegra:
Fluorescent photons created by charged particles surviving RR when radiative splitting ON should be split to avoid having heavy photons. This should happen in EGSnrc, but it is not implemented yet, so do it here! Note that when this is implemented in EGSnrc, the weight check will make sure photons aren't split again!

A fix is already available, but needs to be merged via a PR. This is a reminder to self!

PR #1011, opened 2023-06-09
Fix radiative splitting ao

mainegra:
Initial radiative splitting implementation naively just turned ON EGSnrc internal flag `nbr_split` to split brems and annihilation events uniformly. However no charged particle RR was played after pair production (`i_play_RR`) nor were secondary photons, produced by charged particles surviving RR after bremsstrahlung, split.

This PR fixes issue #1001

Implement radiative splitting + RR
--------------------------------------
- Now turned ON before a radiative event for a primary or a fat charged particle. 
  Turned OFF after the event to avoid higher order splitting. Fluorescent fat photons, 
  generated by fat charged particles (survived RR) are also split.

- RR is turned ON using EGSnrc internal flag `i_play_RR` so that RR is played after pair events. 
  RR can be turned OFF by entering a negative splitting number. 

**NOTE2:** Brems events play RR intrinsically hence it can't be turned OFF.
           Setting `i_play_RR` affects pair production ONLY.

Other minor changes:
------------------------
- Last commit to `egs_dose_scoring.cpp` removed the statement printing media names during dose output to the screen for dose to medium calculations. I have re-added the line.
- First commit to `egs_dose_scoring.cpp` changed the header formatting when outputting media doses which misaligned it.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/egs_radiative_splitting.h
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/egs_radiative_splitting.h
@@ -106,21 +106,71 @@ public:
     ~EGS_RadiativeSplitting();
 
     void setApplication(EGS_Application *App);
-
+    /*! Switch for splitting + RR. Negative nsplit value switches OFF RR. */
     void setSplitting(const int &n_s) {
         nsplit = n_s;
+       if ( nsplit < 0 ){
+          nsplit *= -1;
+          i_play_RR = false;
+       }
+       else if ( nsplit > 1 ){
+          i_play_RR = true;
+       }
+       /* Avoid zero division. A zero value turns off brems */
+       wthin = nsplit ? 1./nsplit : 1.0;
+    };
+
+    bool needsCall(EGS_Application::AusgabCall iarg) const {
+        if (
+             iarg == EGS_Application::BeforeBrems       ||
+             iarg == EGS_Application::BeforeAnnihFlight ||
+             iarg == EGS_Application::BeforeAnnihRest   ||
+             iarg == EGS_Application::AfterBrems        ||
+             iarg == EGS_Application::AfterAnnihFlight  ||
+             iarg == EGS_Application::AfterAnnihRest    ||
+             iarg == EGS_Application::FluorescentEvent  ){
+            return true;
+        }
+        else {
+            return false;
+        }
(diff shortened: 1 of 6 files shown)


HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/egs_radiative_splitting.h
```diff
@@ -106,21 +106,77 @@ class EGS_RADIATIVE_SPLITTING_EXPORT EGS_RadiativeSplitting : public EGS_AusgabO
...
+            return false;
+        }
     };
-    int processEvent(EGS_Application::AusgabCall iarg, int ir) {
+
+    int processEvent(EGS_Application::AusgabCall iarg) {
+
+        /* A fat particle's weight is larger than a thin particle's max weight */
+        bool is_phat = (app->top_p.wt - wthin) > epsilon;
+        bool is_primary = app->top_p.latch == 0 ? true : false;
```
rtownson, line 143: Is this compatible with a phase-space from BEAMnrc where the latch is already set? E.g. it's a primary particle but the latch is non-zero due to one of the bits being set by passage.

ftessier (approved):

ftessier:
@blakewalters @mainegra this pull request is in conflict with #974 (add dbs to radiative splitting). Since the latter is still a draft, I suggest that we merge and then resolve conflicts in the `blake/add-dbs_to_radiative_splitting` branch.

blakewalters:
I agree, @ftessier, let's get this merged in before worrying about DBS splitting, which shouldn't be affected by this change in any case.

HEN_HOUSE/egs++/ausgab_objects/egs_radiative_splitting/egs_radiative_splitting.h
ftessier: @rtownson I will let you "resolve this conversation if you trust that it is ok...

rtownson: Yes, we can resolve this. Currently egs++ actually zeros out the latch when it's read from a phase-space.

rtownson (approved):

blakewalters (approved):
Yeessss!

rtownson:
Formatted with astyle.

PR #1057, opened 2023-08-25
Add triangular mesh support and octree acceleration

a-l-demelo:
Adds support for STL triangular meshes, thanks to @mxxo. Also adds octree acceleration to the implementation, similar to egs_mesh, thanks to @a-l-demelo.

There is still an outstanding bug in the mesh code, so do not expect correct results yet. This is a draft pull request.


ftessier:
Rebased on `develop` to resolve conflicts.

ftessier:
Cleaned up commit titles.

ftessier:
This is a lot of code, so I asked Claude AI (Sonnet 4.6) to chime in about this pr. Here is her take:

- **The main feature** — new egs_triangle_mesh geometry with STL parser and octree
- **Two legitimate bug fixes** in egs_dynamic_geometry (memory leaks) and egs_mesh/Makefile (hardcoded linux path)
- **Debug scaffolding** left throughout the codebase
- **Cosmetic noise** in egs_spheres and egs_genvelope
- **Modifications** to EGS_BaseGeometry that shouldn't be there

ftessier:
Here is a todo list ahead of the merge, created by Claude. I need to double-check if they are all warranted, but some good ideas here:

## Recommended path to merge

### Blockers

- [x] ~~Fix octree `hownear` — leaf must not return bbox wall distance as the answer; parent must explore neighbouring octants when `distance_to_bbox < current_min_t`, not just the single containing octant~~

  **This is not a bug:** we are purposefully capping hownear at octant bbox boundaries.
  
- [x] ~~Fix octree `howfar` — remove early return on first triangle hit; a closer intersection may exist in a later octant; only short-circuit if `min_dist < distance_to_next_octant_bbox`~~ 
  
  **This is not a bug:** `min_dist` is passed by reference and updated in place, so the running minimum is correctly maintained across recursive calls; the early return is safe.
  
- [x] Remove all `cout` calls from production code paths and replace with `egsInformation()`/`egsWarning()` per EGSnrc convention

- [x] Remove `virtual void printTriCheck()` and `virtual void debugtool()` from `EGS_BaseGeometry` — debug hooks do not belong in the base class interface

- [x] ~~Remove all `inctricheck_*()` methods and `tri_check_*` counters from `EGS_TriangleMesh` — these are unprotected debug counters causing data races under parallel transport; make `trimesh` arguments `const &` throughout octree traversal~~

  We will leave those in for now, since they are not in the way and may help in a debugger context. But we should consider removal as the code mature.

### Should be addressed

- [x] Fix ODR violation in `stl_parser.h` — move non-inline function bodies(`parse_ascii_stl_file`, `parse_binary_stl_file`, `parse_stl_file`, etc.) to a `.cpp` file or mark them `inline`

  **Marked the functions `inline`,** no functional change but guarding about ODR violation.

- [x] ~~Harden `isWhere` against degenerate meshes — the fixed arbitrary ray direction `(1/√3, 1/√3, 1/√3)` will produce wrong results for meshes with edges/vertices aligned to that diagonal; add a fallback with a second ray direction or a multi-ray majority vote~~

  **No need to harden further,** simply tweak comment where this was already acknowledged, 

- [x] Validate STL normals on load — warn (and ideally recompute from edge cross product) when `|n|` is not approximately 1, consistent with how `egs_mesh` handles normals

  **Added validation of STL normals,** and added a "normals" option: calculate or stored.

- [x] Switch `octree accelerate` default to `yes` ~~once `hownear` is fixed~~ — defaulting to the O(N) naive path defeats the purpose of the PR for users who don't read the docs

  **Octree acceleration is now on by default.**

- [x] Remove unused variable `xo_dist` and unused `n_hist` counter from `isWhere`

### Housekeeping

- [x] Add `a-l-demelo` to the contributor list in the file headers of new files

- [x] ~~Separate the unrelated fixes (`egs_dynamic_geometry` memory leaks, `egs_mesh/Makefile` portability) into their own PRs or call them out explicitly in the PR description~~

- [x] Remove the unused `int e` parameter in `BoundingBox::intersects_triangle` (only referenced in commented-out `cout` statements)

- [x] Add at least one geometry correctness test — e.g. a sphere approximated as a mesh compared against `egs_sphere` with a known analytic answer

HEN_HOUSE/egs++/egs_application.cpp
```diff
@@ -1051,6 +1051,10 @@ int EGS_Application::finishSimulation() {
     int err = 0;
     if (run) {
         err = run->finishSimulation();
+
```
ftessier: Remove ad hoc debugging code ro release.

ftessier:
I removed the commented out debug statements. If you think these are still required for further development, I can reintroduce them with a proper `verbosity` input parameter. What do you think?

ftessier:
Added a todo comment about a potentially more efficient octree traversal inspired by the binary location codes implemented in egs_octree.

ftessier:
Refactored findOtherIntersectedOctants to avoid heap allocation and vector copying on every call to howfar and isWhere.

ftessier:
Turned on octree acceleration by default, and improve egsInformation for message origin in the code base.

ftessier:
@a-l-demelo and @rtownson, please review changes since d52427c4180b4a24a7141063633b2ba142d7ff7c

rtownson:
@ftessier The changes seem good. I tested with a sphere stl mesh and got agreement with 0.1% uncertainty.

ftessier:
Just condensed a commit title to fit 50 characters 😄

ftessier:
@rtownson perhaps we should add the sphere `.stl` you mentions, and insert a call in the test suite to perform that test?

That would complete the checklist above by taking care of the last recommendation:

> - [x] Add at least one geometry correctness test — e.g. a sphere approximated as a mesh compared against `egs_sphere` with a known analytic answer

rtownson:
Ok, I included the sphere .stl file and an example input file with expected results in the file. I didn't put this in the test suite since it currently doesn't have a clean way to launch actual simulations and test results from within. This might be better suited to egs_quality, if we ever go that route.

ftessier:
Is it worth throwing a pure egs_sphere in there for comparison?

ftessier:
Ah nice! I had forgotten that .stl renders on github directly! 🤩 

image

ftessier:
@rtownson, perhaps add a link in the input file to https://3dviewer.net/ to view the stl. A pretty nice viewer!

ftessier:
Awesome! So the mesh version is about 5 times slower, not bad considering the sphere is simplest of the _functional_ geometries.

ftessier:
Just rebased on the updated `develop` branch.

PR #1069, opened 2023-10-20
Fix egs xyzrepeater bug

mainegra:
Fix bug in EGS_XYZRepeater hownear
    
    - Typo in hownear when particle in bounding box determining which of
      the repeated geometries to check:
    
      Using nz rather than nxy when computing index of repeated geometry array
      results in addressing array index out of bounds with subsequent segfault error.
    
      Showed when eletron transport turned ON up during dose calculations around
      a CivaDot source using egs_dose_scoring and egs_app and 2 microns regions.
    
      It did not affect photon-only nor egs_kerma calculations since there is no
      call to hownear. ;-)


rtownson (approved):

blakewalters (approved):

ftessier (approved):

rtownson:
Tweaked commit message.

PR #1077, opened 2023-11-15
Fix egs++ iaea phsp scoring electron energy

rtownson:
Fix a bug in egs_phsp_scoring where the electron (and positron) energy for IAEA phase-spaces was written as the total energy instead of the kinetic energy. This is a critical bug and impacted all IAEA format phase-spaces generated from egs++ that contained charged particles.

Thanks to Erik Nilsson for reporting this issue!


blakewalters (approved):
Yes, this is consistent with how we're using IAEA format in the other codes...and, of course, with the definition of the IAEA format itself.  Thanks, Reid!

mainegra (approved):

PR #1082, opened 2023-11-30
Fix a c++14 mingw compiler error and some warnings

rtownson:
Fix a compiler error and some warnings found when compiling with mingw on windows, gcc 13.2.0. Note that currently users need to add ` -std=c++14` manually by editing the g++ settings in the installer, when installing with modern compilers that default to a newer standard. This should get resolved automatically at the next release when the installer is updated to include `c++14` by default.


mchamberland:
I've been struggling with installing on Windows these past two weeks. Will give that a shot!

mchamberland:
It looks like inputting the -std=c++14 option in the 'Standard options' text field in the configure Windows application gets ignored. I had to put it under the 'Optimization options' for it to get picked up. Is that a bug?

rtownson:
> It looks like inputting the -std=c++14 option in the 'Standard options' text field of the g++ settings in the configure Windows application gets ignored. I had to put it under the 'Optimization options' for it to get picked up. Is that a bug?

Sounds like a bug!

mchamberland:
I don't really understand all of the egs_configure code, but I don't see where the standard g++ options get used. [I can see the optimization options getting used in the create_egspp_config(), though.](https://github.com/nrc-cnrc/EGSnrc/blob/fe83057e6db99768a281e343498e3eedf185206d/HEN_HOUSE/gui/egs_configure/egs_install_conf.cpp#L1123)

mchamberland:
Also, this PR appears to work as advertised: I was able to compile on Windows with mingw, gcc13.2.

blakewalters (approved):

ftessier:
@rtownson, did "the installer get updated to include c++14 by default"?

rtownson:
> @rtownson, did "the installer get updated to include c++14 by default"?

Yes, in https://github.com/nrc-cnrc/EGSnrc/pull/899

ftessier (approved):

mainegra (approved):

Issue #1117, opened 2024-04-12
Biased sampling of phase space source (C++) in parallel runs with ncase > no. of particles in phase space file


blakewalters:
**Describe the bug**
Currently in parallel runs where ncase (total no. of histories) > Nphsp (no. of particles in phase space source), any "chunk" of the run for which the final history of the chunk > Nphsp generates a warning indicating that a suitable chunk of the phase space source cannot be found (confusing labelled an "illegal attempt to set a simulation chunk") then proceeds to use the last chunk of the phase space source that was successfully accessed (Note: usually close to, but not necessarily including, the end of the phase space data).  Thus, there is the potential for sampling this last successfully accessed chunk of the phase space source over and over again. 

**To Reproduce**
1. Generate a small phase space file containing Nparticle particles
2. Use it as a source in a simple egs++ simulation with ncase = 3*Nparticle
3. Run a single parallel job using command line arguments -P 4 -f 1 -j 1
4. Look in the _w1.egslog file and see the appearance of many "illegal attempt to set simulation chunk" warnings each followed by a message indicating that the same chunk of the phase space file (same first and last particle no.'s) is about to be reused.

**Expected behavior**
Just like ncase is divided into (npar \* nchunk) chunks, we need to divide Nparticle into (npar \* nchunk) segments.  Each chunk of the parallel run will then have its own designated segment of the phase space source outside of which it will not sample.  This will ensure that we don't attempt to sample past the end of the phase space data and that as much of the phase space data as possible is sampled.  This is similar to how we deal with parallel runs with phase space sources in the .mortran codes.

**Operating system**
All applicable

**EGSnrc version**
Develop branch

**Additional context**

PR #1126, opened 2024-05-09
Fix parallel phsp source partition

blakewalters:
This fix divides phase space sources used in parallel runs into npar\*nchunk equal segments, where npar is the no. of parallel runs and nchunk is the no. of chunks (defaults to 10).  Each chunk of the parallel run is then assigned its own segment of the phase space source outside of which it will not sample.  This scheme, identical to that used in the Mortran codes, should ensure relatively even sampling of the phase space source over the entire parallel run.  Note that any remaining histories after completion of npar*nchunk chunks are assigned to the last segment. 

Fixes issue #1117 and #1156

### DIFF
--- a/HEN_HOUSE/egs++/egs_application.cpp
+++ b/HEN_HOUSE/egs++/egs_application.cpp
@@ -798,9 +798,9 @@ int EGS_Application::initSimulation() {
     return 0;
 }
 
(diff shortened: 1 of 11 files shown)


rtownson (approved):

ftessier:
Squashed some commits, fixed commit messages. No content diff with tip of original pull request branch.

ftessier (approved):

mainegra (approved):

PR #1132, opened 2024-05-23
Fix #1131: clarify comment about ms conditional

### LINKED ISSUE #1131: Incomplete multiple-scattering comment in egsnrc.mortran
Lines [1247-1250](https://github.com/nrc-cnrc/EGSnrc/blob/fe83057e6db99768a281e343498e3eedf185206d/HEN_HOUSE/src/egsnrc.mortran#L1247-L1250) in egsnrc.mortran state

```
IF((tustep <= tperp) & ((~exact_bca) | (tustep > skindepth)))
                [
                    "We are further way from a boundary than a skindepth, so
                    "perform a normal condensed-history step
```

The comment ignores the fact that only if the step is greater than skindepth does the CH step get taken. Otherwise the CH step is replaced by single scattering.   As far as I can tell, the coding is correct, it's just the comment that I find misleading. It should be something like

```
"We are further away from a boundary than a skindepth or the requested CH step is
"greater than skindepth so perform a normal condensed-history step
```

ftessier:
Clarify that multiple-scattering transport is triggered only if the intended electron step is longer than a skin depth, even in bulk media away from any region boundaries. This improves efficiency, because a multiple-scattering step costs about the same as 3 single-scattering steps (the default skin depth is 3 mean free paths). Below a skin depth, it is more efficient (and more accurate) to use single-scattering.

### DIFF
--- a/HEN_HOUSE/src/egsnrc.mortran
+++ b/HEN_HOUSE/src/egsnrc.mortran
@@ -1246,8 +1246,10 @@ $start_new_particle;
                 is_ch_step = .false.;
                 IF((tustep <= tperp) & ((~exact_bca) | (tustep > skindepth)))
                 [
-                    "We are further way from a boundary than a skindepth, so
-                    "perform a normal condensed-history step
+                    "We are further away from a boundary than a skindepth and
+                    "either: we are not using exact boundary-crossing, or the
+                    "requested condensed-history step is greater than the
+                    "skindepth. So perform a normal condensed-history step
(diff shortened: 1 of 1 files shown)


mainegra (approved):

rtownson (approved):

blakewalters (approved):
Okay, I get it!  It took me a couple of readings and a bit of puzzling through this logic.  And that's not the first time I've found this logic puzzling.

PR #1144, opened 2024-06-05
Zero energy deposited by track ends in vacuum.

### LINKED ISSUE #1061: Dose deposited in vacuum in DOSRZnrc
Using DOSRZnrc with a vacuum region on the surface, I occasionally get dose deposited in the vacuum region which is impossible. The attached file takes some time to run but shorter runs don't see the dose deposited.
[vac_bug.txt](https://github.com/nrc-cnrc/EGSnrc/files/12507256/vac_bug.txt)
Blake looked into this in Oct 2022 and in a series of emails isolated the problem but the solution wasn't obvious. I can forward emails if not found at NRC.

blakewalters:
Handle "rare" cases of energy deposited (edep) in vacuum due to tracks ending once particles cross into the region by immediately zeroing edep.  Requires a fix to dosxyznrc to handle the case of 0/0=NAN causing a crash.  May also require fixes to other user codes to handle similar NAN cases.

Fixes #1061

### DIFF
--- a/HEN_HOUSE/src/egsnrc.mortran
+++ b/HEN_HOUSE/src/egsnrc.mortran
@@ -1770,8 +1770,7 @@ IF( medium > 0 ) [
         IF(lelec < 0) [edep = e(np) - prm;] ELSE[$POSITRON-ECUT-DISCARD;]
     ]
     ELSE [ idr = $PEGSCUTAUS; edep = e(np) - prm; ]
-] ELSE [idr = $EGSCUTAUS; edep = e(np) - prm; ]
-
+] ELSE [idr = $EGSCUTAUS; edep = 0.0; ] "zero edep in vacuum"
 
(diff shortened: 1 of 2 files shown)


rtownson:
There has been a long-standing bug in DOSXYZnrc where the code will tend to segmentation fault if you use the "surrounding material" as VACUUM (the default). I wonder if this fixes that!

blakewalters:
That would be nice, @rtownson!  But I don't know.  Do these segmentation faults occur during or at the end of the runs?

rtownson:
I think it was during, but I'll have to find a case that reproduces this.

blakewalters:
@rtownson, @ftessier, @mainegra:  Is there any reason this can't be included in the 2024 release?

rtownson (approved):

mainegra (approved):

PR #1152, opened 2024-06-08
Check for endianness definitions in IAEA utilities

mchamberland:
This gets rid of compilation warnings that LITTLE_ENDIAN, BIG_ENDIAN, and PDP_ENDIAN were already defined on macOS.

This is the same as the previous pull request I had opened; I just renamed my branch to better keep track of things.


ftessier:
I think that this is ok, since the _value_ of these constants is never used per se in the IAEA phsp code, it is just a label, so to speak.

ftessier (approved):

mchamberland:
For the record, I did double-check that the macOS versions have the same definitions, not that it matters.

rtownson (approved):

blakewalters (approved):

mainegra (approved):

rtownson:
Tweaked commit message.

PR #1163, opened 2024-07-26
Update the astyle options file to be compatible with newer versions of astyle

rtownson:
Tested on Artistic Style Version 3.4.11. This was required because the option names changed slightly between versions. A comment was made at the top of the options file describing what changed. 

Also ran astyle on egs++. There were a handful of minor changes.


ftessier (approved):

ftessier:
@mainegra @blakewalters Please double-check as soon as possible, so that we can fold this format-only commit in `develop` before more changes accumulate ahead of release.

mainegra (approved):

Issue #1145, opened 2024-06-05
Allow generation of .3ddose files for transformed EGS_XYZGeometry


blakewalters:
**Is your feature request related to a problem? Please describe.**
Currently, egs++ supports reading a CT .egsphant file into a rectilinear EGS_XYZGeometry.  Using EGS_DoseScoring, a .3ddose file can potentially be generated for a simulation using this EGS_XYZGeometry.  However, if, as is almost always the case, the EGS_XYZGeometry created using CT data must be rotated/translated from "CT space" into the simulation coordinate system, a .3ddose file for the transformed geometry can no longer be generated.

**Describe the solution you'd like**
It would be very useful to be able to generate a .3ddose file for CT data.  By default, this data would be output in "CT space" and could be plotted atop the untransformed CT image.  There may be a way to also transform the dose data into simulation space.

**Alternatives**
I suppose one could build their simulation geometry around the original CT data, but this seems unnecessarily complicated and inflexible.

mchamberland:
[egs_brachy did it first.](https://github.com/clrp-code/egs_brachy/commit/7ddc4bccdc4959ef64b6584997b873ce3a77daae)

😉

Though, this situation does not come up as often in brachytherapy simulations. Also, I have not handled converting the 3ddose file to "simulation space". It seems too daunting to me at the time, but I admit I only thought about the problem for about 5 minutes before coming to that conclusion!

blakewalters:
@mchamberland, I've been using a fix that undoes the transformations at the beginning of the run just to determine: 1) no. of voxels, 2) voxel boundaries, and 3) mapping of global dose region -> region no. in EGS_XYZGeometry.  This seems to work okay...but it causes a crash in egs_view if the egs_dose ausgab object with the dose geometry isn't commented out before attempting to display.

blakewalters:
@rtownson and @ftessier: This issue is addressed by https://github.com/nrc-cnrc/EGSnrc/pull/1165, which might be an easy "add" to the 2024 release.

PR #1165, opened 2024-08-09
Enable dose file with transformed geometry

blakewalters:
This modification allows output of .3ddose files for a transformed EGS_XYZGeometry.  Transformed dose distributions can also be read by and viewed with egs_view.  Note that the coordinate system of the output .3ddose file corresponds with that of the untransformed base EGS_XYZGeometry.  Bear this in mind when using other applications to view dose distributions.  This addition is very useful for generating .3ddose files for CT phantoms, which almost always require some transformation(s) to position them correctly in the geometry.

Addresses issue #1145


rtownson:
This issue was also fixed in https://github.com/nrc-cnrc/EGSnrc/pull/1276

So we should avoid merging this PR, and instead double check that everything in here was covered by 1276.

PR #1170, opened 2024-08-19
Fix #1004: beamdp axis labels are confusing

### LINKED ISSUE #1004: error in axis labels in beamdp.mortran plot routines
There are many outputs for plots like

```
YTITLE='planar fluence/incident particle /cm\\S-2\\N';
```

This should be

```
YTITLE='planar fluence/incident particle /cm\S-2\N';
```

i.e., `\\S` --> `\S` and `\\N` --> `\N`. These are superscript and back to normal instructions for xmgrace. 

The error occurs at  8 locations.

blakewalters:
Addresses #1004.

Note: I find our method for specifying axis units in beamdp using a backslash in front of the unit (i.e., '/unit') confusing, especially for quantities which may also have a unit in the denominator (e.g., '/MeV cm^-2').  I wonder if we should move to just using brackets instead.

### DIFF
--- a/HEN_HOUSE/doc/src/pirs3100-g-app/g_template.f
+++ b/HEN_HOUSE/doc/src/pirs3100-g-app/g_template.f
@@ -5832,7 +5832,7 @@ C> @cond
             IPLOTE=0
             MEDIUM=i
             XAXIS = 'kinetic energy / MeV'
-            YAXISE = 'dE/drhoX MeV/g/cm\\S2\\N'
+            YAXISE = 'dE/drhoX MeV/g/cm\S2\N'
             YAXISEmfp = 'mean free path / cm'
             YAXISPmfp = 'mean free path / cm'
             write(GRAPHTITLE,'(24a1)')(media(j,i),j=1,24)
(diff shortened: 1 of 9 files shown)


mainegra (approved):

ftessier:
@blakewalters I agree with you about slash notation for units, for it becomes quite confusing, as you said, when the units themselves involve a quotient. I find that $v\ (\mathrm{m/s})$ clearer than $v / (\mathrm{m/s})$, for example.

However, the [SI Brochure](https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf) Section 5.4 is quite definitive in its recommendation for _Rules and style conventions for expressing values of quantities:_

> Symbols for units are treated as mathematical entities. In expressing the value of a quantity as the product of a numerical value and a unit, both the numerical value and the unit may be treated by the ordinary rules of algebra. This procedure is described as the use of quantity calculus, or the algebra of quantities. For example, the equation p = 48 kPa may equally be written as p/kPa = 48. It is common practice to write the quotient of a quantity and a unit in this way for a column heading in a table, so that the entries in the table are simply numbers. (...) The axes of a graph may also be labelled in this way, so that the tick marks are labelled only with numbers, as in the graph below.

So you and I will probably just have to get used to it?

ftessier:
I fixed all the other occurrences of erroneous flags `\\S` (superscript) and `\\N` (normal) in Grace plot axes labels throughout the EGSnrc code base.

ftessier (approved):

ftessier:
Commits should be squashed before merging.

rtownson (approved):

Issue #512, opened 2019-02-02
Single element materials without density correction file not working with pegsless mode


ojalaj:
When using a single element material without a density correction file, following is output:

  density_file polymethylmethacrylate__lucite___perspex___plexiglas_                                                                                                                         
  Error: Material not defined: W16p9_700

***************** Error:
 Material used in the geometry was not defined in the material data.

***************** Quiting now.

and to .mederr:
  For medium: W16p9_700
  Rho specified in material data file
  IAPRIM specified in material data file
  Error: Medium W16p9_700                not correctly defined.

Tested e.g. with EX16MVp on a 'develop' branch from early October 2018. Works fine e.g. with WRE_19p4_700 (multiple elements, no density correction file) or SN700ICRU (single element, density correction file) or STEEL700ICRU (multiple elements, density correction file).

ftessier:
@ojalaj is this still an issue?

ojalaj:
@ftessier Haven't tested on newest release, but tested on 2020 release and the answer is yes - I was still able to reproduce this issue.

ftessier:
Then it is probably still active then (I don't recall any changes to pegsless in Release 2021. I will check it here, thank you! 🙏🏻

ftessier:
I am not able to reproduce this problem in egs++ (using the `egs_app` example on a fresh EGSnrc clone). Here is the content of my `mat.dat` file:

```
medium   = test
rho      = 1
elements = AG
```

Upon running with this, the simulation runs as expected and the relevant part of the output shows:

```
  Material data file: mat.dat                                                                                                                                                                                                                                                         

   Medium: air                     
 Sterncid: air                     
     rho:    1.20479E-03 g/cm^3
 Elements: C   N   O   AR  
    rhoz:    1.24000E-04   7.55267E-01   2.31781E-01   1.28270E-02
      pz:    1.03237E-05   5.39218E-02   1.44869E-02   3.21092E-04
  iunrst:     0
  iaprim:     0
    gasp:    0.00000E+00 atm.
  density correction file: /home/ftessier/repos/tmp/HEN_HOUSE/pegs4/density_corrections/compounds/air_dry_nearsealevel.density

   Medium: test                    
 Sterncid: test                    
     rho:    1.00000E+00 g/cm^3
 Elements: AG  
    rhoz:    1.07870E+02
      pz:    1.00000E+00
  iunrst:     0
  iaprim:     0
    gasp:    0.00000E+00 atm.
```

[...]

PR #1182, opened 2024-09-06
Fix pegsless error reading media consisting of a single element.

blakewalters:
Addresses #512

### DIFF
--- a/HEN_HOUSE/src/get_media_inputs.mortran
+++ b/HEN_HOUSE/src/get_media_inputs.mortran
@@ -257,18 +257,8 @@ DO I = NMIN, NMAX  [ "for each value_sought"
    ]
 
    "===============read the value(i)======================="
-   IF ( (lnblnk1(TEXT)=0) | (lnblnk1(TEXT)=1) )
+   IF ( (lnblnk1(TEXT)=0) )
    [
-        IF (vname(:ivname)='TITLE')
-        [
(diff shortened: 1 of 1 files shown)


blakewalters:
This issue was caused by erroneously throwing an error when any input was 1 character long and can be traced back to the fact that the subroutine GET_INPUT_PLUS, used to read media inputs, is heavily based on (i.e. some blocks directly copied from) GET_INPUT, the routine used to read general input parameters.  GET_INPUT allows for the definition of a "TITLE" string for which the check on single character input is relevant (I think), which is not relevant for media inputs.  Thus, I've also stripped all references to this input field.

mainegra (approved):

rtownson (approved):

ftessier:
Rebased on current `develop` to consider only the one actual commit in this pr.

HEN_HOUSE/src/get_media_inputs.mortran
```diff
@@ -278,7 +268,6 @@ DO I = NMIN, NMAX  [ "for each value_sought"
         ERROR_FLAGS(I)=1;
         RETURN;
    ]
-:TITLE-THERE:
    CONTINUE;
```
ftessier: I think this `CONTINUE;` statement is dangling now that `:TITLE-THERE:` label has been removed.

PR #1194, opened 2024-10-01
Fix bug in UBS/BCSE EII photon splitting block in BEAMnrc

mainegra:
BUG: Commit 8023cd01 labels fluorescent photons after EII as secondary.
         The pertinent block for UBS and/or BCSE requires knowledge of the local
         region number IRL, which is only set later in AUSGAB.

FIX: Set IRL above the UBS/BCSE block in the AUSGAB routine.

This PR also adds the libclean target to the standard BEAMnrc Makefile.

### DIFF
--- a/HEN_HOUSE/makefiles/beam_makefile
+++ b/HEN_HOUSE/makefiles/beam_makefile
@@ -162,6 +162,12 @@ clean:
 	@$(MAKE) -s WHAT=_noopt execlean
 	@$(MAKE) -s WHAT= execlean
 
+libclean:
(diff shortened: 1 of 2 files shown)


ftessier:
Rebased on `develop` and formatted commit titles and messages.

rtownson (approved):
Is `libclean` called automatically when you `make clean`? If not, I think it should be.

ftessier (approved):

blakewalters (approved):

mainegra:
> Is `libclean` called automatically when you `make clean`? If not, I think it should be.

I thought of it. But it is seldom that one is working with both BEAM or BEAM as a source.
Was trying to avoid carpet cleaning everything.

PR #1215, opened 2024-11-03
Remove ambiguous terminology with "restart" across EGSnrc

### LINKED ISSUE #1210: The term "restart" is ambiguous across EGSnrc
Setting the run control to restart a job is a misnomer — really we are resuming from the last batch. Restarting sounds like we are running it from the beginning. This has also been confused by new users with the idea of restarting a phase-space.

ftessier:
Fix #1210 and more! Replace all instances and variants of `restart` in EGSnrc with more precise terminology:

- **Resume** — continue a simulation after an interruption (as @rtownson suggested above)
- **Rewind** — read a phase space file from the beginning, aligning with the programming concept of rewinding a file
- **Relaunch** — for processes, consistent with the idea of launching a program or command

The three commits should not be squashed, to keep the distinct steps visible in the history, corresponding to updating each semantic meaning of `restart`.


ftessier:
Although a bit tedious, I would invite a careful review here. I have checked all changes twice, and running `rg -i restart` in `EGSnrc/` yields no instances on this branch. You can use `grep` also, `rg` is [ripgrep](https://github.com/BurntSushi/ripgrep)), which is faster.

rtownson:
I think it wouldn't be too hard to make this change backwards compatible. I agree with changing the documentation and everything to no longer reference the old terms, but we could still accept the old inputs so people don't have to change their input files & scripts.

mainegra (approved):
Not sure how to check this one! I guess the best is to merge it and take it for a spin! :-)

ftessier:
For backward compatibility, we would have to isolate the instances where "restart" is read as a string in the input files, and support both "restart" and the new keyword (resume, rewind or relaunch).

rtownson (approved):

PR #1222, opened 2024-11-20
Change configure script to overwrite existing .conf

rtownson:
Change the configure.expect script so that existing configurations of the same name are overwritten. This avoids user confusing when they are following the installation instructions to re-configure, but the expect script refuses to continue due to a configuration of the same name already existing.

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_fluence_scoring/egs_fluence_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_fluence_scoring/egs_fluence_scoring.cpp
@@ -855,7 +855,7 @@ void EGS_PlanarFluence::ouputResults() {
         spe_output << "@    xaxis  label char size 1.560000\n";
         spe_output << "@    xaxis  label font 4\n";
         spe_output << "@    xaxis  ticklabel font 4\n";
-        spe_output << "@    yaxis  label \"fluence / MeV\S-1\Ncm\S-2\"\n";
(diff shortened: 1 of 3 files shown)


mainegra (approved):

blakewalters (approved):

ftessier (changes requested):
I don't like this change because it will silently override an existing configuration. I will revert back to the previous exit behaviour, but add a graceful message to indicate what to do.

ftessier:
I need to check the existing file paths, and that it works as expected.

rtownson:
> I don't like this change because it will silently override an existing configuration. I will revert back to the previous exit behaviour, but add a graceful message to indicate what to do.

Why is it dangerous to overwrite an existing configuration? I think if you're making compiler flag changes or configuring for different machines then you're not using the expect script, and you're not naming your configuration 'expect.conf'.

ftessier:
Yes, but you can pass an argument to specify the name of the configuration file. It is good practice to warn about file overwrites, and let the user remove the files themselves (providing the exact paths). If only for my benefit 😉. The real issue here is that if the EGSnrc environment variables are set, and you are trying to install elsewhere, it would silently override the configuration for your original installation. I do this all the time. 😄

ftessier:
Still need to test this version of the script.

ftessier:
Fixed the information message about which files to remove, to enable expect to proceed without overwriting files. Seems to work as expected.

ftessier:
@rtownson I prefer this version, please rebase and merge if you agree.

PR #1228, opened 2024-11-28
Fix g calc type 1

mainegra:
## First change
    
`Calculation type` 1 did not incorporate the changes implemented in 2017 to remove kerma calculation dependency to transport and production thresholds `ECUT` and `AE`, as well as the inclusion of electron impact ionization in the $\mu_\mathrm{tr}$ calculation step.
    
As a result, the $\mu_\mathrm{en}$ and $\mu_\mathrm{tr}$ values calculated using either calculation type do not agree. The mismatch arises from the lack of distinction of the energy deposited by subthreshold events between Auger electrons, which contribute to kerma, and fluorescent photons, which do not.
 
This issue is addressed by introducing additional AUSGAB calls before and after the routines handling photoelectric effect, Compton scattering, and pair production for type 1 calculations. For type 0 calculations a check is performed in ausgab to apply or not the classification schema or not, based on the value of the logical variable `classify_sub_threshold_relaxations`.
    
Notably, reproducing current NIST $\mu_\mathrm{en}$ values is only possible when this fix is excluded, indicating that the NIST database may need an update to reflect these corrections.

    
## Second change:
    
Add option to reproduce NIST values by setting input key below to `no`
```    
    CLASSIFY SUB-THRESHOLD RELAXATIONS = NO
```
    
Default set to `yes`.


HEN_HOUSE/user_codes/g/g.mortran
```diff
@@ -540,15 +539,33 @@ IF( calc_type = 1 ) [  "will seek a given precision -- default does not do this"
...
         $EVALUATE gbr2 USING gbr2(my_gle);
         np = 1; e(np) = ein; u(np)=0; v(np)=0; w(np)=1; iq(np) = 0;
-        $RANDOMSET rnno; edep = 0;
+        $RANDOMSET rnno; edep = 0; etot_tmp = 0.0;
         IF( rnno < gbr1 & ein > rmt2 )[
-                 call pair;
+           IF (classify_sub_threshold_relaxations)[
+              $AUSCALL($PAIRAUSB);
+           ]
+              call pair;
```
ftessier, line 547: Reduce indentation for `call pair;`

HEN_HOUSE/user_codes/g/g.mortran
```diff
@@ -887,7 +911,7 @@ DO icase=1,ncase [
   erad_tmp = 0.0;            "loses just to annihilation"
   "initialize etot_tmp differently for photons and electrons"
   IF( iqin .ne. 0 ) [ etot_tmp = (ein - rm); ]
-  ELSE [ etot_tmp = 0; ]
+  ELSE              [ etot_tmp = 0; ]
```
ftessier, line 914: It is best to avoid if-else one-liner statements.

HEN_HOUSE/user_codes/g/g.mortran
```diff
@@ -1203,6 +1228,31 @@ $GET_INPUT(IVAL);
...
+" Ignore sub-threshold relaxations?               "
+"-------------------------------------------------"
+ival                = ival + 1;
+values_sought(ival) = 'CLASSIFY SUB-THRESHOLD RELAXATIONS';
+nvalue(ival)        = 1;
+type(ival)          = 3;
+allowed_inputs(ival,0) = 'NO';
+allowed_inputs(ival,1) = 'YES';
+$GET_INPUT(IVAL);
+IF( error_flag = 0 ) [
```
ftessier, line 1241: Ideally, we would harmonize the `IF` bracketing style. My perspective is that we ought to settle on conditional branch style with spaces around ` (...) ` (no spaces inside the brackets, and one space before the opening ` [`

```fortran
IF (...) [
  ...
]
ELSE (...) [
  ...
]
```

ftessier (approved):

rtownson (approved):

ftessier:
Needs a squash with title `Harmonize calculation types in g application`.

PR #1235, opened 2024-12-13
Add support for region labels to egs_kerma

rtownson:
Also improve the label replacement algorithm to allow for inputs that have a mix of parameters that are not regions. For example, "start stop delta", where the delta is the increment value.

Use `git diff --ignore-space-space` to best view the differences.


mainegra (approved):

rtownson:
Closing this to supersede it with #1343.

PR #1236, opened 2024-12-13
Fix egs kerma diff flu ouput for log-scale

mainegra:
## Bugs fixed in this PR

- Fixes diff fluence normalization to proper bin width when using log-scale.
- Fixes bug printing out average source energies and number of particles to egsdat file causing combining parallel jobs to fail.
- Fixes bug introduced to preserve latch and only multiply it by -1 when entering an exclusion region. Now photon is killed rather than relying on latch. Speeds up example calculation `example_1m_40keV_FD.egsinp` by 40 %
- Veryfies that scoring regions are not define as exclusion regions.

## Keep source latch

The latch of particles was reset to 0 and now is accepted as it comes from the source. This can be useful for VRT AOs such as radiative splitting.

## Verifying approach killing photons entering exclusion regions

| Case      | Kerma / Gy/particle | wall contribution| histories / h |
| ----------- | ----------- |---|---|
| latch = -1      |   3.530745e-18 +/- 0.0074  %     | 1.0134794   +/- 0.003971  % | 2.06299e+09 |
| kill photon   | 3.530758e-18 +/- 0.0077  %       | 1.0135483   +/- 0.004152  % | 2.81407e+09 |

### DIFF
--- a/HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
+++ b/HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
@@ -237,19 +237,20 @@ public:
 
                     }
                 }
-                /* Mark photon inside cavity */
-                latch = 1;
+                /* Mark photon inside cavity <= why did I need this?*/
+                //latch = 1;
             }
             /* Mark photons in exclusion regions */
-            else if (is_excluded[ig][ir]) {
-                latch = -1;
+            else if (is_excluded[ig][ir] && latch >= 0 ){
+                latch *= -1;
+                the_stack->latch[np] = latch;
             }
-            /* photon outside cavity and has not been in any exclusion zone */
+            /* photon outside cavity and has not been in any exclusion zone <= why?
             else if (latch >= 0) {
                 latch = 0;
-            }
+            }*/
 
-            the_stack->latch[np] = latch;
+            //the_stack->latch[np] = latch; <= if commented blocks above used
         }
         return 0;
     }
@@ -270,7 +271,7 @@ public:
             Nel     += p.wt;
             p.E += the_useful->rm;// source provides K.E.
         }
-        p.latch = 0; // Reset to 0
+        //p.latch = 0; // Reset to 0
         //if ( p.q != 0 )
         //   egsInformation("q = %d E = %g RM = %g wt = %g \n",p.q,p.E,the_useful->rm,p.wt);
 
--- a/HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
+++ b/HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
@@ -199,10 +199,9 @@ public:
         int np = the_stack->np-1, ir = the_stack->ir[np]-2, iq = the_stack->iq[np];
         /* Photon about to be transported in geometry */
         if (iarg == BeforeTransport &&  !iq && ir >= 0) {
-            int latch = the_stack->latch[np];
             /* Track-Length Kerma scoring (classic) */
             if (is_sensitive[ig][ir]) {
-                if (!fd_geom && latch >= 0) {
+                if (!fd_geom) {
                     EGS_Float E = the_stack->E[np], gle = log(E),
                               emuen   = E_Muen_Rho->interpolateFast(gle)*rho_cv[ig],
                               wtstep  = the_stack->wt[np]*the_epcont->tvstep;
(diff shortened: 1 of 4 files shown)


rtownson (approved):

HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
```diff
@@ -239,19 +239,20 @@ class APP_EXPORT EGS_KermaApplication : public EGS_AdvancedApplication {
...
-                latch = 1;
+                /* Mark photon inside cavity <= why did I need this?*/
+                //latch = 1;
             }
             /* Mark photons in exclusion regions */
-            else if (is_excluded[ig][ir]) {
-                latch = -1;
+            else if (is_excluded[ig][ir] && latch >= 0 ) {
+                latch *= -1;
+                the_stack->latch[np] = latch;
```
blakewalters, line 248: @mainegra, I'm not totally sure how latch is being used.  From the logic, it looks to me like if latch=0--true unless egs++ DBS is being used or latch has been set in a phase space or BEAMnrc source--then we can't mark the photon as having been in an exclusion region (i.e. latch *=-1 = 0).  Does this amount to a necessary restriction on when exclusion regions can be used or am I missing something?

mainegra: The current assumption is that if a particle enters an exclusion zone, it is marked with a negative latch, and it does not contribute to kerma. One could save time and simply kill it, I guess. According to this, there wouldn't be a conflict.

blakewalters: But what if latch=0 going into the simulation?  Can the photon be marked on entering the exclusion zone then?

ftessier (approved):

HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
mainegra: Ah, yes, I see your point now! I used to simply set it to -1 and in an attempt to preserve the incoming latch scheme I changed it to multiplying by -1 which as you correctly stated won't work! At this point it is best to revert it to simply setting it to -1 and clarifying that `egs_kerma` ignores any external latch-based classification. For `egs_kerma`, if the particle is marked as -1 and so are its secondary particles, then one might as well kill the particle and save the CPU time.

blakewalters: Thanks for confirming my shaky notions of logic, @mainegra!

rtownson:
Tweaked commit messages.

HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
blakewalters: Okay, thanks @mainegra.

blakewalters: Makes sense to me, @mainegra.  Thanks!  

Just a couple more questions: 
1. example_40keV_SDD_1m_FD.egsinp needs $EGS_HOME/egs_kerma/emuen_icru90_1.5MeV.data but I don't see it included with the PR.  Should this be changed to point to one of the .data files that are included?
2. Unmodified, example_20MeV_e-_Pb-shield_FD.egsinp stops with a "too many ustep errors" message on my machine.  Does this happen for you, or is this a property of my OS?

rtownson: I also get the ustep errors for `example_20MeV_e-_Pb-shield_FD.egsinp`.

mainegra: @blakewalters great catch!

1.  The E*muen file in that example is not up-to-date. It should be rather `emuen_rho_air_1keV-20MeV.data`. I will update the example file!

2. This one is weird and I am investigating it. If one increases UE from 20.511 MeV to 20.711 MeV in the material definition section, the example runs without error messages. Morever the negative usted errors report a `kinetic` electron energy of 20.511 MeV as if there is an extra rest mass (RM) being added to the e- energy.

blakewalters: Hmm, I see that total energy in the negative ustep messages.  Hopefully, it's just an output issue.  I'll take a look.

rtownson: @blakewalters Ernesto has already solved it, he'll post a fix shortly.

mainegra: Done! @blakewalters take anbother look and if it all looks good! This has been very useful Blakey! Gracias!

blakewalters: De nada!  Might take me a bit to get to it, but I'll check it out this aft.

blakewalters (approved):

PR #1243, opened 2025-01-16
Add ensdf parsing to ignore NN neutron record

rtownson:
The NNDC ensdf file for Cf-252 had decays to `NN-6` representing neutron nuclei that as far as I can tell is not supported by the ensdf standard. A check for this case has been added, so that these records are ignored. Previously, you would just encounter a crash since NN is not a valid element.


ftessier:
Added a commit message to detail the three different problems addressed in this pr.

PR #1259, opened 2025-02-21
Fix fluence rectangle transform

blakewalters:
Fix input of transform applied to fluence scoring rectangle in the egs_fluence_scoring AO.

Previously, we looked for this input between the ":start/stop ausgab object:" delimiters instead of the ":start/stop planar scoring:" delimiters.  As a result, it wasn't read in/applied.


mainegra (commented):
Why was this changed? Seems wrong to me!

ftessier:
Yes, these commits will be squashed before merging.

PR #1267, opened 2025-03-14
Bug fix and minor fluence calculations issues

mainegra:
Bug fix
=====
During output, `egs_kerma` requires the medium for region `ir` to be able to extract its volume from the user-provided region mass to score fluence. This was implemented by using the `EGS_AdvancedApplication` method

```C++
  getMedium(ir);
```

which gets the medium index for region `ir` in the geometry defined as `simulation geometry`. 

This is a bug since one can have more _calculation geometries_ with different media in their scoring regions. The solution is to use the `medium` method of each calculation geometry `g[j]`:

```C++
g[j]->medium(ir);
```

# Decoupling differential and integral fluence calculations

The existing implementation requires a `fluence scoring` block inside the `scoring options` input block. It triggers an _integral_ fluence scoring and a _differential_ fluence scoring _ONLY IF_ at least `minimum` and `maximum energies`, as well as `number of bins` are provided (see documentation for all input options). To only score _integral_ fluence, one leaves the `fluence scoring` input block empty. This approach can be confusing and a more explicit approach is now used:

  a) Integral fluence: Within the `scoring options` input block use the following input key
```
:start scoring options:
     ...
     score integral fluence = [yes|no] # defaults to no
    ...
:stop scoring options:
```

  b) Differential fluence:  Insert a `differential fluence scoring` input block providing at least `minimum`, `maximum` energies as well as `number of bins`.
```
:start scoring options:
     ...
    :start differential fluence scoring:
        minimum energy = Emin
        maximum energy = Emax
        number of bins = nbins
        scale          = linear # linear (default) or logarithmic
    :stop differential fluence scoring:
    ...
:stop scoring options:
```

The `fluence scoring` block is still recognized for backwards compatibility. But, in this case, only differential fluence is calculated.


rtownson:
Please review this PR sooner so that I can merge in it before the egs_view editor PR. Then I can update the input parameter support in the egs_view editor with these new inputs.

HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
```diff
@@ -895,7 +922,7 @@ class APP_EXPORT EGS_KermaApplication : public EGS_AdvancedApplication {
                                        irmax_digits, ir, m,fe*normT, dfe);
                     }
                 }
-            }
+            }*/
```
ftessier: Remove the dead code that is commented out.

ftessier:
This needs an `astyle` pass before merging

HEN_HOUSE/user_codes/egs_kerma/egs_kerma.cpp
```diff
@@ -1608,7 +1607,34 @@ int EGS_KermaApplication::initScoring() {
...
+            flugT = new EGS_ScoringArray* [ngeom];
+            for (int j=0; j<ngeom; j++) {
+                flugT[j] = new EGS_ScoringArray(geoms[j]->regions());
+            }
+        }
+
+        vector<string> diff_f_key = {"fluence scoring",
+                                     "differential fluence scoring"
+                                    };
+        for (short int i = 0; i < diff_f_key.size(); i++) {
```
ftessier: Type mismatch between `short int` and `size_t`. Just use `size_t` for index `i` here.

ftessier: Done. ✅

PR #1273, opened 2025-03-21
Fix phsp scoring outside the geometry on entry

rtownson:
Fix a bug where particles were scored "on entry" in the phase-space even though they had left the overall geometry. This was occurring when the particles were only meant to be scored on entry to a particular geometry, if the region of the phase-space scoring geometry happened to be defined outside the actual geometry used in the simulation (like an infinite plane).

Bug was reported here:
https://www.reddit.com/r/EGSnrc/comments/1jga62g/framelike_structure_in_beamdp_xy_scatter_plot/

@blakewalters please check this!


rtownson:
> So, one need to include an envelope around the actual geometry if one wanted to score outside that geometry right?

If you want to score on a plane outside your geometry, you'll have to envelope it all, then have the plane cutting through your envelope at some distance.

PR #1276, opened 2025-03-31
Add mesh support to egs_dose_scoring

rtownson:
Automatically determines the mass for calculating dose for both egs_mesh and XYZ geometries, when the egs_dose_scoring ausgab object is set to output a dose file. This also allows for dose to each medium to be reported correctly for XYZ and mesh geometries without the user needing to specify the volume of each voxel/tetrahedron. 

Note that the masses are only automatically calculated when a `output dose file` is used, because that's the only time we can be sure (i.e. the user has assured us) that the geometry isn't cut up by a composite geometry.

Dose for egs_mesh geometries can now be output in `.vtk` or `.csv` formats without using `mevegs`.

Also fixes a bug where dose to 3ddose files would be wrong if the XYZ geometry was transformed.

E.g.
```
:start ausgab object:
  library           = egs_dose_scoring
  name              = some_name
  medium dose = yes
  region dose = no
  :start output dose file:
     geometry name = must name an EGS_XYZGeometry (for 3ddose output) or EGS_Mesh (for vtk or csv)
     file type = 3ddose, vtk, or csv
  :stop output dose file:
:stop ausgab object:
```

### DIFF
--- a/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
+++ b/HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
@@ -230,32 +230,6 @@ void EGS_DoseScoring::setApplication(EGS_Application *App) {
             int ny=dose_geom->getNRegDir(1);
             int nz=dose_geom->getNRegDir(2);
 
-            // Let's keep this commented out but around, because it was really
-            // useful for debugging that the below method worked correctly
-            // Note that the following did *not* work for transformed geometries, because it looks at the original positioning of the voxels before they were transformed!
-            /*
-            EGS_Vector tp;
-            EGS_Float minx,maxx,miny,maxy,minz,maxz;
-            for (int k=0; k<nz; k++) {
-                for (int j=0; j<ny; j++) {
-                    for (int i=0; i<nx; i++) {
-                        minx=dose_geom->getBound(0,i);
-                        maxx=dose_geom->getBound(0,i+1);
-                        miny=dose_geom->getBound(1,j);
-                        maxy=dose_geom->getBound(1,j+1);
-                        minz=dose_geom->getBound(2,k);
-                        maxz=dose_geom->getBound(2,k+1);
-                        tp.x=(minx+maxx)/2.;
-                        tp.y=(miny+maxy)/2.;
-                        tp.z=(minz+maxz)/2.;
-                        int g_reg = app->isWhere(tp);
-                        df_reg[g_reg]=i+j*nx+k*nx*ny;
-                        //egsInformation("%d %d\n", df_reg[g_reg], g_reg);
-                    }
-                }
-            }
-            */
-
             int globalOffset = app->getGlobalRegionOffset(dose_geom->getName());
             if(globalOffset < 0) {
                 globalOffset = 0;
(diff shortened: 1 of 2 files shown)


HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
```diff
@@ -217,32 +222,73 @@ void EGS_DoseScoring::setApplication(EGS_Application *App) {
...
-                    df_reg[g_reg]=i+j*nx+k*nx*ny;
+
+        // Determine the region no. in the specified geometry and the corresponding global reg. no.
+        if(file_type==0) {
+
+            int nx=dose_geom->getNRegDir(0);
+            int ny=dose_geom->getNRegDir(1);
+            int nz=dose_geom->getNRegDir(2);
+
+            // Let's keep this commented out but around, because it was really
```
ftessier: Remove dead code. It can be recovered from the history.

HEN_HOUSE/egs++/geometry/egs_genvelope/egs_envelope_geometry.cpp
```diff
@@ -495,6 +495,72 @@ extern "C" {
...
+                }
+
+                return shift;
+            }
+        }
+
+        // If it's not found above, search through the inscribed geometries in case they are composite geometries
+        for (int i=0; i<n_in; i++) {
+            int shift = geometries[i]->getGlobalRegionOffset(geomName);
+            if(shift > 0) {
```
ftessier: This is different from `if(shift >= 0)` in the regular envelope (above). Is this intentional? I am not familiar with the fast envelope logic!

HEN_HOUSE/egs++/geometry/egs_genvelope/egs_envelope_geometry.cpp
```diff
@@ -495,6 +495,72 @@ extern "C" {
...
+                if (new_indexing) {
+                    shift += local_start[i];
+                }
+                else {
+                    shift = nmax;
```
ftessier: Again, this is different from `shift += nbase+i*nmax;` for the regular envelope case above. Worth a double-check whether this is intentional. Compare with howfar logic: https://github.com/nrc-cnrc/EGSnrc/blob/e09fd6942996bba660a8cdb2baaf50d2ee800549/HEN_HOUSE/egs%2B%2B/geometry/egs_genvelope/egs_envelope_geometry.h#L248-L255)

Need to check getLabelRegions() as well for fast envelope, which I think has the same bug!

HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
```diff
@@ -472,8 +559,131 @@ void EGS_DoseScoring::outputDoseFile(const EGS_Float &normD) {
...
+            df_out << "10\n";
+        }
+
+        // doses
+        df_out << "CELL_DATA " << mesh->num_elements() << "\n";
+        // adjust number here vvvv if the number of fields written out changes
+        df_out << "FIELD FieldData 2\n";
+        // %20 url-encoded space, Paraview errors on space character
+        df_out << "dose%20[Gy] 1 " << mesh->num_elements() << " double\n";
+        const double JOULES_PER_MEV = 1.602e-13;
```
ftessier: I would define this to full precision, since it is now [defined exactly](https://physics.nist.gov/cgi-bin/cuu/Value?e).

```
const double JOULES_PER_MEV = 1.602176634e-13;
```

HEN_HOUSE/egs++/ausgab_objects/egs_dose_scoring/egs_dose_scoring.cpp
```diff
@@ -472,8 +559,131 @@ void EGS_DoseScoring::outputDoseFile(const EGS_Float &normD) {
...
+        for (int i = 0; i < mesh->num_elements(); i++) {
+            double e_dep, uncert;
+            doseF->currentResult(i,e_dep,uncert);
```
ftessier, line 592: The normalization factor `normD` is not applied for the VTK output, is this intentional?

rtownson: It was intentional in mevegs, which this code was copied from. Though I do agree that we should change it and make it normalized by the fluence like other outputs.

ftessier:
Rebased on updated develop, null diff with previous branch tip.

PR #1307, opened 2025-06-26
Fix #1300: egs_view segmentation fault that started when we added dynamic geometries

### LINKED ISSUE #1300: Geometry viewer egs_view crashes when loading CD geometry
**Describe the bug**
Attempt to open an input file with `egs_view` for a simple geometry, results in a `segfault` error. Using older versions of `egs_view` (Aug and Nov 2024) still works. I will try to narrow down since when `egs_view` has been broken!

BUG seems to also appear when scoring tracks!

**To Reproduce**
Example which fails with recent versions of `egs_view` while it works with versions I have from 2024. Geometry is a lead box filled with air and two openings in the front and back sides. Using an `XYZ geometry` as the base of a `CD geometry` and inserting a cylindrical section into the regions where the openings would be.

```Yaml
:start geometry definition:

    :start geometry: # ===> lead box filled with air
        library = egs_ndgeometry
        type = EGS_XYZGeometry
        name    = a_box
        x-planes = -10 -8 8 10
        y-planes = -10 -8 8 10
        z-planes = -10 -8 8 10
        :start media input:
         media = lead air
         set medium = 1 1 1 1 1 1 1
        :stop media input:
    :stop geometry:

    :start geometry:         # ===> opening
        library = egs_cylinders
        type = EGS_ZCylinders
        radii = 2.0 100.0
        name    = opening
        :start media input:
          media = lead air
          set medium = 0 1
        :stop media input:
    :stop geometry:

    :start geometry: #===> led box with openings
        library = egs_cdgeometry
        name = a_case
        base geometry = a_box
        set geometry  =  4  opening
        set geometry  = 22  opening
    :stop geometry:

    simulation geometry = a_case

:stop geometry definition:
```
**Expected behavior**

Visualize lead box with openings in fron and back faces

**Screenshots**

```bash
Segmentation fault (core dumped)
```

**Operating system**

- [X] Linux
- [ ] Windows
- [ ] macOS
- [ ] Other: 

**EGSnrc version**
v2025, commit `79890aaa`

rtownson:
Fix a segmentation fault that was occurring in egs_view due to accessing memory that was not initialized in the geometry. This occurred during a call to `containsDynamic()`, which in the composite geometries was missing a check on whether a geometry existed before propagating the `containsDynamic()` call. The result was a segmentation fault in egs_view for specific use cases of a CD geometry. I added a protective check for all composite geometries, but only detected the issue for CD geometries so far.

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
+++ b/HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
@@ -823,25 +823,33 @@ do_checks:
 
     void getNextGeom(EGS_RandomGenerator *rndm) { //calls getNextGeom on its component geometries to update dynamic geometries in the simulation
         for (int j=0; j<bg->regions(); ++j) {
-            g[j]->getNextGeom(rndm);
+            if (g[j]) {
+                g[j]->getNextGeom(rndm);
+            }
         }
-        bg->getNextGeom(rndm);
+        if (bg) {
+            bg->getNextGeom(rndm);
+        }
     };
 
     void updatePosition(EGS_Float time) {//calls updatePosition on its component geometries to update dynamic geometries in the simulation
         for (int j=0; j<bg->regions(); j++) {
-            g[j]->updatePosition(time);
+            if (g[j]) {
+                g[j]->updatePosition(time);
+            }
         }
-        bg->updatePosition(time);
+        if (bg) {
+            bg->updatePosition(time);
+        }
     };
 
     void containsDynamic(bool &hasdynamic) {//calls containsDynamic on its component geometries (only calls if hasDynamic is false, as if it is true we already found one)
         for (int j=0; j<bg->regions(); j++) {
-            if (!hasdynamic) {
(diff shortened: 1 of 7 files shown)


HEN_HOUSE/egs++/geometry/egs_cd_geometry/egs_cd_geometry.h
```diff
@@ -823,25 +823,33 @@ class EGS_CDGEOMETRY_EXPORT EGS_CDGeometry : public EGS_BaseGeometry {
...
     };
 
     void updatePosition(EGS_Float time) {//calls updatePosition on its component geometries to update dynamic geometries in the simulation
         for (int j=0; j<bg->regions(); j++) {
-            g[j]->updatePosition(time);
+            if (g[j]) {
+                g[j]->updatePosition(time);
+            }
+        }
+        if (bg) {
```
ftessier, line 836: This reads a bit funny because we guard against a null `bg`, but `bg->regions()` appears unguarded in the loop condition. I think it is ok because a cd geometry must have a base geometry, but perhaps we ought to guard the loop as well?

PR #1324, opened 2025-08-13
Enable region ranges and default-all for labels

ftessier:
Add the ability to apply a label to all regions in the local geometry by default when no region number is specified in the label input, e.g., `set label = mylabel`.

Add parsing of region ranges specified with a hyphen, as in `set label = mylabel 5-12`, to label all regions in the range. Both multiple region numbers and ranges can be specified as separate tokens in the same `set label` input.

This becomes handy for geometries containing a large number of regions. Previously, all region numbers had to be specified individually.

Also, improve tokenization and parsing robustness, as well as error messages.


rtownson:
Closing this to supersede it with #1343

PR #1327, opened 2025-09-11
Add catch2 testing framework 

MattInglisWhalen:
egs++ is a large library with no testing framework. This pull request
(PR) proposes the use of the header-only catch2 framework, available at
https://github.com/catchorg/Catch2/releases/download/v2.13.10
in order to write unit tests for each of the egs++ classes. This is meant
for developers only, and therefore is written to not conflict with 
user-only builds of egs++.

This PR provides a pair of examples (testing for EGS_Vector and 
EGS_RandomGenerator, with test_egs_vector.cpp and test_egs_rndm.cpp)
in the new `tests` folder of egs++. Some convenience functions for these 
two classes have also been added, and all EGS_Vector functions are now
properly commented.

As part of a good testing framework, coverage reports are invaluable.
This PR leverages built-in c++ compiler flags and the `gcov` utility
(packed with gcc) to produce html summaries of the tested files. 
Due to idiosyncrasies of `gcov`, a second pass is made using the new 
`fix_coverage` utility (compiled from the new `fix_gcov.cpp`  source). 
This allows uninstrumented functions (not compiled and therefore 
unused) to also be shown in reports. This requires building egs++ with
compiler coverage flags.

These test/coverage additions to egs++ required a significant extension
to the egs++ Makefile. It makes for a less concise reading of the Makefile,
but the tradeoff is that it allows for an automated execution of the 
test + coverage cycle. The main `make` keywords that have been added are
`make run-tests` and `make coverage`. A few helper keywords accompany
these additions.

Finally, the `make style` keyword has been added in order to maintain 
a consistent and convenient styling of all tested classes. This relies on the
dual utilities `astyle` and `cpplint.py`. The former should be available
to linux and osx developers through apt or brew, while the latter only 
requires access to python3 on your system and the cpplint.py file itself.
The use of these utilities has forced style changes to egs_rndm.h, 
egs_rndm.cpp, and egs_vector.

image

Caveats: I have tested this workflow only on ubuntu. 
`catch2`, `astyle`, and `cpplint.py` are not the property of EGSnrc and therefore
should not be included in the public repo. I have probably not included the 
proper licensing boilerplate in new files.


rtownson:
Ah I could see your case working if you have added something like `export LD_LIBRARY_PATH=$HEN_HOUSE/egs++/dso/env` to your `~/.bashrc`. But this isn't something a user normally does, so it would be nice to not expect it.

MattInglisWhalen:
Issue reproduced and understood. I was focused on the coverage so much that I wasn't testing `make tests` by itself from a fresh pull. I had no additional LD_LIBRARY_PATH variable set.

The fix is to cut the the `compile_test_dso` dependency from the `coverage:` rule (line 325) and paste it as the first dependency for the `tests:` rule (line 303). So line 325 should read 

> coverage: $(TEST_DIR)fix_coverage_report run-tests 

and line 303 should read

> tests: compile_test_dso $(TEST_DIR)run_tests

I think reverting your changes and implementing my suggestions would be the cleanest way forward.

rtownson (approved):

PR #1329, opened 2025-09-23
Add a check in egs_isotropic_source that max_phi > min_phi

rtownson:
Adds a check in `egs_isotropic_source` to avoid a possible easy mistake that users can make, when providing collimation ranges in theta or phi. The check now requires that `max_phi > min_phi` and `max_theta > min_theta`. This avoids a user providing `min_phi=350` and `max_phi=10`, for example, and expecting the range to pass over 0. Instead, the user is advised to add 360 to `max_phi` to achieve this.


ftessier:
I removed the "branch point" term, which is obscure here. Also, it is not possible to go above $\theta = \pi$ in spherical coordinates. Right now we don't guard against it.

PR #1341, opened 2025-10-28
Fix #1340: qt5 egs_view compiler error for QMainWindow

### LINKED ISSUE #1340: Fail to compile egs_view in linux
**Describe the bug**
Hi there,

I’m trying to compile egs_view after a recent fresh installation/configuration using the shell script, as our department’s cluster now has Qt5 installed. However the compilation for egs_view fails.

I’ve properly set the QTDIR and QT_SELECT environment variables in my .bashrc file as recommended:

export QTDIR=/usr/lib64/qt5
export QT_SELECT=qt5

All other GUIs were successfully built using the same process within HEN_HOUSE/gui, so the issue seems specific to egs_view. I’d really appreciate any help identifying what might be causing this problem.

Thanks,

**To Reproduce**
A precise description of the steps to reproduce the behavior:

1. Within HEN_HOUSE/egs++/view type make clean and then make
2. An error message appears and compilation fails: 

viewcontrol.cpp: In constructor ‘GeometryViewControl::GeometryViewControl(QWidget*, const char*)’:
viewcontrol.cpp:92:25: error: array used as initializer
     : QMainWindow(parent) {

**Screenshots**

Image

**Operating system**

- [x] Linux

**EGSnrc version**
v2025a

rtownson:
Fix a compiler error for egs_view in qt5 reported here: https://github.com/nrc-cnrc/EGSnrc/issues/1340

Since it didn't occur for me, it may only happen when stricter compiler flags are used. Regardless, it will help with the eventual move to qt6.

### DIFF
--- a/HEN_HOUSE/egs++/view/viewcontrol.h
+++ b/HEN_HOUSE/egs++/view/viewcontrol.h
@@ -211,7 +211,7 @@ private:
     bool showElectronTracks;
     bool showPositronTracks;
     bool hasDynamic; //boolean to track whether or not to make time objects visible or hidden
-    char head_inctime[20] = "include time index="; // must match the same in egs_particle_track.h
+    char head_inctime[20] = {'i','n','c','l','u','d','e',' ','t','i','m','e',' ','i','n','d','e','x','='}; // must match the same in egs_particle_track.h
     bool hasTrackTimeIndex;
     bool isPlaying;
     vector<bool> show_regions;


rtownson:
> Are we moving to qt6? That will require a major rehaul of our GUIs! If I remember correctly qt6 breaks backwards compatibility albeit in the name of progress (major cleanup). 😜

In the egs_view PR I added Qt6 support to egs_view, while maintaining backwards compatibility! Qt6 has been cleaned up a lot in recent years so it's easier to support now.

HEN_HOUSE/egs++/view/viewcontrol.h
```diff
@@ -157,6 +157,7 @@ public slots:
 
 private:
 
+    Ui::GeometryViewControl *ui;
```
ftessier: I don't understand this new `ui` pointer. We still [inherit `ui`](https://github.com/nrc-cnrc/EGSnrc/pull/1341/changes#diff-7afb67768479dfa124e3418673093ff235e78bf388dcc0efc168923ea866bea6R92), so this new pointer seems to do nothing? I don't see any `ui->` accesses inside viewcontrol.cpp, so I infer this new `ui` is never actually used.

HEN_HOUSE/egs++/view/viewcontrol.cpp
```diff
@@ -88,10 +88,10 @@ static unsigned char standard_blue[] = {
...
 };
 
-GeometryViewControl::GeometryViewControl(QWidget *parent, const char *name)
-    : QMainWindow(parent) {
-    setObjectName(name);
-    setupUi(this);
+GeometryViewControl::GeometryViewControl(QWidget *parent)
+    : QMainWindow(parent), ui(new Ui::GeometryViewControl) {
+    setObjectName("GeometryViewControl");
+    ui->setupUi(this);
```
ftessier: Related to my comment on [line R160](https://github.com/nrc-cnrc/EGSnrc/pull/1341/changes#diff-52eed31750b3cd01a15152eb893b00417027abd1363b131d360194963ac3a7b0R160): Why not continue to use `setupUi(this)` on the inherited `ui`?

ftessier: It seems that the inherited `ui` is not "setup" anymore, so how does this even work? 😄

ftessier: Ah, I see in the pr comments: this would be to help the move to Qt6? But Qt6 documents that both [single-inheritance](https://doc.qt.io/qt-6/designer-using-a-ui-file.html#the-single-inheritance-approach) (pointer) and [multiple-inheritance](https://doc.qt.io/qt-6/designer-using-a-ui-file.html#the-multiple-inheritance-approach) (inherit) patterns are supported. The move to the pointer can be made, sure, but the current change is misleading: the inherited `ui` is not setup. This does not seem to matter here as I find no direct access to the inherited `ui.` in the source code.

If we moved to the pointer model, we should remove the inherited `ui` and convert any `ui.` accesses to a pointer `ui->` dereference. And ensure there is a `delete(ui)` in the class [destructor.](https://github.com/nrc-cnrc/EGSnrc/blob/8ccb9fc4fb7208b922f94d6bb27c280a05b21f5e/HEN_HOUSE/egs%2B%2B/view/viewcontrol.cpp#L194)

rtownson: You're right, these changes were my first attempt to fix the compiler error from the issue and shouldn't have still been there in this PR. I've removed them and just left the actual fix.

ftessier (approved):

Issue #1338, opened 2025-10-24
EGSnrc_osx.F error


bobbiplob:
Running egs_app using slab.egsinp producing EGSnrc_osx.F error:

sahak@sahakVLFW egs_app % egs_app -i slab.egsinp 
In EGS_Application::initSimulation()
**********************************************
EGS_RandomGenerator::createRNG: no 'RNG definition' input
EGS_Application::initRNG(): using default RNG
================================================================================
EGSnrc version 4 for arm-apple-darwin24.5.0            Fri Oct 24 18:03:36 2025
================================================================================
configuration............................................osx
configuration time.......................................2025-10-10 00:39:41 UTC
app compile time.........................................2025-10-10 00:47:02 UTC
git commit hash..........................................79890aa
application..............................................egs_app
pegs file................................................
using host...............................................sahakVLFW
input file...............................................slab
output file(s)...........................................slab
================================================================================
In initCrossSections(): spin effects = 1
 ===> Photonuclear flag:   0
 RAYLEIGH OPTION REQUESTED FOR MEDIUM NUMBER  1

 RAYLEIGH OPTION REQUESTED FOR MEDIUM NUMBER  2

  PEGSLESS INPUT.  CALCULATING ELECTRON CROSS-SECTIONS.
  density_file air_dry_nearsealevel                                                                                                                                                                                                                                            
  density_file water_liquid                                                                                                                                                                                                                                                    

  Medium data: 

 AE =    5.21000E-01 MeV,  UE =    5.05110E+01 MeV
 AP =    1.00000E-02 MeV,  UP =    5.00000E+01 MeV

  No material data file supplied.  Material data obtained from
  .egsinp file or density correction file.

   Medium: air                     
 Sterncid: air                     
     rho:    1.20479E-03 g/cm^3
 Elements: C   N   O   AR  
    rhoz:    1.24000E-04   7.55267E-01   2.31781E-01   1.28270E-02
      pz:    1.03237E-05   5.39218E-02   1.44869E-02   3.21092E-04
  iunrst:     0
  iaprim:     1
    gasp:    0.00000E+00 atm.
  density correction file: /Users/sahak/EGSnrc/EGSnrc/HEN_HOUSE/pegs4/density_corrections/compounds/air_dry_nearsealevel.density

   Medium: water                   
 Sterncid: water                   
     rho:    1.00000E+00 g/cm^3
 Elements: H   O   
    rhoz:    1.11894E-01   8.88106E-01
[...]

Issue #1321, opened 2025-08-07
egs_app Fortran runtime errors when running on Lubuntu 25.04


ElliotLucaMedCo:
**Describe the bug**
Multiple Fortran runtime errors on encountered on egsnrc_linux.F when trying to run the sample simulation:
egs_app -i slab.egsinp

Seems to only occur in the latest lubuntu 25.04. I had it running in 24.04.2 LTS with no issue.

**To Reproduce**
Using guide at https://github.com/nrc-cnrc/EGSnrc/wiki/install-EGSnrc-on-VirtualBox on a Lubuntu 25.04 image. At step 6 got multiple Fortran errors when running the egs_app -i slab.egsinp

Fortran runtime errors found in egsnrc_linux.F at the following lines:
line 21316 
line 22111 
line 13669 
line 19047
line 19069
line 19097

All giving:
Fortran runtime error: Missing comma between descriptors 
(/a$)
    ^

**Expected behavior**
egs_app -i slab.egsinp should run without issue

**Operating system**

- [X] Linux
- [ ] Windows
- [ ] macOS
- [ ] Other: 

**EGSnrc version**
develope

**Additional context**
Worked fine once I deleted all the $.

crcrewso:
I just ran across this trying to recreate some of my old work. I think this is going to become a bigger problem as more distros start using the gfortran >= 15

This warning could be solved by adding the -std=legacy flag to the gfortran arguments, or by adding all of the missing commas in the source code. A change was made to elevate legacy string formatting to a warning. 

[GCC Change](https://www.mail-archive.com/fortran@gcc.gnu.org/msg06156.html)

PR #1342, opened 2025-10-28
Fix gcc14 compiler support

rtownson:
Fix a number of compiler errors and warnings related to gcc14 (possibly some earlier versions as well). Note that this adds `-std=legacy` to fortran compilation by default. Ideally, we should fix the mortran/fortran to comply with modern standards eventually.

This fixes Discussion https://github.com/nrc-cnrc/EGSnrc/discussions/1335, Issues https://github.com/nrc-cnrc/EGSnrc/issues/1338 and https://github.com/nrc-cnrc/EGSnrc/issues/1321.

Currently, gcc15 compiles but the egs++ applications segmentation fault while running (even with -std=legacy). I feel like this implies there's an issue in the interface. This PR doesn't solve that issue.

### DIFF
--- a/HEN_HOUSE/egs++/egs_application.h
+++ b/HEN_HOUSE/egs++/egs_application.h
@@ -1142,12 +1142,15 @@ public:
         return geometry->getMediumName(ind);
     };
     virtual EGS_Float getMediumRho(int ind) {
+        (void)ind;
         return -1.0;
     };
     virtual EGS_Float getEdep() {
         return 0.0;
     };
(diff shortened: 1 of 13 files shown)


ftessier:
> Currently, gcc15 compiles but the egs++ applications segmentation fault while running (even with -std=legacy). I feel like this implies there's an issue in the interface. This PR doesn't solve that issue.

I suggest this be:

- Called out explicitly in the release notes ("gcc15 not yet supported"), or
- Tracked as a blocking issue for the next release?

PR #1343, opened 2025-10-31
Improve label parsing and add labels to egs_kerma

rtownson:
This is a merge of PRs #1324 and #1325, intended to supersede them both. The conflicts between the two have been resolved.

Add the ability to apply a label to all regions in the local geometry by default when no region number is specified in the label input, e.g., set label = mylabel.

Add parsing of region ranges specified with a hyphen, as in set label = mylabel 5-12, to label all regions in the range. Both multiple region numbers and ranges can be specified as separate tokens in the same set label input. This becomes handy for geometries containing a large number of regions. Previously, all region numbers had to be specified individually.

Add parsing of inputs that have a mix of parameters that are not regions. For example, "start stop delta", where the delta is the increment value - now the "start" and "stop" can be labels.

Add label support to `egs_kerma`.

Also, improve tokenization and parsing robustness, as well as error messages.

### DIFF
--- a/HEN_HOUSE/egs++/egs_base_geometry.cpp
+++ b/HEN_HOUSE/egs++/egs_base_geometry.cpp
@@ -1049,20 +1049,11 @@ void EGS_BaseGeometry::getNumberRegions(const string &str, vector<int> &regs) {
     if (!str.empty()) {
 
         // Tokenize the input string
-        vector<string> tokens;
-        const char *ptr = str.c_str();
-        do {
-            const char *begin = ptr;
-            while (*ptr != ' ' && *ptr) {
-                ptr++;
-            }
-            tokens.push_back(string(begin, ptr));
-        }
-        while (*ptr++ != '\0');
+        vector<string> tokens = egsTokenize(str);
 
+        // Search for tokens that are numbers, not strings
+        // Push the region numbers onto the regions array
         for (int i=0; i<tokens.size(); i++) {
-            // Search for tokens that are numbers, not strings
-            // Push the region numbers onto the regions array
             if (tokens[i].find_first_not_of("-0123456789") == std::string::npos) {
                 regs.push_back(atoi(tokens[i].c_str()));
             }
@@ -1072,17 +1063,8 @@ void EGS_BaseGeometry::getNumberRegions(const string &str, vector<int> &regs) {
 
 void EGS_BaseGeometry::getLabelRegions(const string &str, vector<int> &regs, bool sanitize) {
 
-    // Tokenize the input string - this allows for multiple labels
-    vector<string> tokens;
-    const char *ptr = str.c_str();
-    do {
-        const char *begin = ptr;
-        while (*ptr != ' ' && *ptr) {
-            ptr++;
-        }
-        tokens.push_back(string(begin, ptr));
-    }
-    while (*ptr++ != '\0');
+    // Tokenize the input string
+    vector<string> tokens = egsTokenize(str);
 
     // Start insertion at the beginning of regs
     size_t insert_pos = 0;


mainegra:
@rtownson would it be a good idea to update the documentation for this extensive PR? This is really useful

HEN_HOUSE/egs++/egs_base_geometry.cpp
```diff
@@ -1080,18 +1080,40 @@ void EGS_BaseGeometry::getLabelRegions(const string &str, vector<int> &regs) {
     }
     while (*ptr++ != '\0');
```
ftessier: Is this not left over from the hand-rolled pointer walk tokenizer?

HEN_HOUSE/egs++/egs_base_geometry.cpp
```diff
@@ -1080,18 +1080,40 @@ void EGS_BaseGeometry::getLabelRegions(const string &str, vector<int> &regs) {
...
 
+    // Start insertion at the beginning of regs
+    size_t insert_pos = 0;
+    bool foundLabel;
+
     // Get all regions lists for this named label
-    for (int j=0; j<tokens.size(); j++) {
-        for (int i=0; i<labels.size(); i++) {
+    for (int j = 0; j < tokens.size(); j++) {
```
ftessier, line 1074: I need to check if this is correct when `regs` is initially empty.

ftessier:
This pr might need some commit cleaning before merging.

HEN_HOUSE/egs++/egs_base_geometry.cpp
ftessier: Oh no, I see, I have not replaced the hand-rolled tokenization in getLabelRegions() and get getNumberRegions() with the new tokenizer function.

HEN_HOUSE/egs++/egs_base_geometry.cpp
ftessier: Yeah, if regs is empty, i.e. not first populated with non-label regions with getNumberRegions(), then the not found branch below will walk past the end of `regs`. I added an egsFatal() guard for now so this does not fail silently. We ought to review the logic of this label replacement, feels a bit convoluted now.

```c++
if(!foundLabel) {
    insert_pos += 1;
}
```

HEN_HOUSE/egs++/egs_base_geometry.cpp
```diff
@@ -1059,14 +1059,14 @@ void EGS_BaseGeometry::getNumberRegions(const string &str, vector<int> &regs) {
         for (int i=0; i<tokens.size(); i++) {
             // Search for tokens that are numbers, not strings
             // Push the region numbers onto the regions array
-            if (tokens[i].find_first_not_of(" -0123456789") == std::string::npos) {
+            if (tokens[i].find_first_not_of("-0123456789") == std::string::npos) {
```
ftessier, line 1057: Removed dead space character from the token parsing set, because the tokenizer loop already filters out all spaces.

HEN_HOUSE/egs++/egs_base_geometry.cpp
```diff
@@ -1045,64 +1045,76 @@ void EGS_BaseGeometry::getNumberRegions(const string &str, vector<int> &regs) {
...
-        const char *ptr = str.c_str();
-        do {
-            const char *begin = ptr;
-            while (*ptr != ' ' && *ptr) {
-                ptr++;
-            }
-            tokens.push_back(string(begin, ptr));
-        }
-        while (*ptr++ != '\0');
+        vector<string> tokens = egsTokenize(str);
```
ftessier, line 1052: Use the simpler and safer string stream tokenizer instead of the hand-rolled pointer walk.

HEN_HOUSE/egs++/egs_base_geometry.cpp
```diff
@@ -1045,64 +1045,76 @@ void EGS_BaseGeometry::getNumberRegions(const string &str, vector<int> &regs) {
...
-    do {
-        const char *begin = ptr;
-        while (*ptr != ' ' && *ptr) {
-            ptr++;
-        }
-        tokens.push_back(string(begin, ptr));
-    }
-    while (*ptr++ != '\0');
+    // Tokenize the input string
+    vector<string> tokens = egsTokenize(str);
```
ftessier, line 1067: Use the simpler and safer string stream tokenizer instead of the hand-rolled pointer walk.

ftessier:
@rtownson, please look at the conversations to resolve or comment further.

rtownson:
> @rtownson, please look at the conversations to resolve or comment further.

I'm happy with your improvements if you are, looks good

ftessier:
Need to:

- [x] Fix the guard on insert position
- [x] Resolve conflicts with `develop`
- [x] Condense commit titles to <= 50 characters

ftessier:
Rebased on `develop` branch and resolved one conflict:

```diff
diff --git a/HEN_HOUSE/egs++/egs_base_geometry.h b/HEN_HOUSE/egs++/egs_base_geometry.h
index 9ad1d59c..a6d6a5fd 100644
--- a/HEN_HOUSE/egs++/egs_base_geometry.h
+++ b/HEN_HOUSE/egs++/egs_base_geometry.h
@@ -755,6 +755,10 @@ public:
     /*! \brief Set the labels from an input string */
     int setLabels(const string &inp);

+    virtual void updatePosition(EGS_Float time) {
+        (void)time;
+    };
+
     /*!
     * \brief Validate that all region numbers are within valid bounds
     *
@@ -781,8 +785,6 @@ public:
     */
     bool validateRegions(const std::vector<int> &regions);

-    virtual void updatePosition(EGS_Float time) { };
-
     /* This method is essentially used to determine whether the simulation
      * geometry contains a dynamic geometry. Like getNextGeom(), the only
      * non-empty implementations of this function are in composite geometries
```

PR #888, opened 2022-07-05
Integrate ESTAR density effect corrections into EGSnrc

rtownson:
The ESTAR fortran code from NIST has been translated into C++ and integrated into EGSnrc, so that density effect corrections are now calculated on-the-fly during the initialization of a simulation. Materials can be defined directly in the input file, so .pegs4dat and .density (density correction files) are no longer needed.

Example:
```
:start media definition:
  :start water:
    elements = H, O
    number of atoms = 2, 1
    bulk density = 1.0 # g/cm^3 used for transport
    local density = 1.0 # g/cm^3 used for density corrections
    ivalue = 0.000075 # MeV, optional
    output density file = /mypath/water.density # full path to output density corrections to file, optional
  :stop water:
:stop media definition:
```
For mixtures of compounds:
```
    :start saltWater:
        mixture compounds = H2O, NaCl
        mass fractions = 0.95, 0.05
        bulk density = 1.0
    :stop saltWater:
```

For comparison, there's an option to output a density file when the simulation is run.

This adds the options for bulk and local densities to be specified separately. The `bulk density` is the density used for transport, and the `local density` is used for the density effect corrections (when a density file is not provided). Density correction files, if provided, will always override these inputs. If the `local density` is not provided, `bulk density` (or `rho`) is used for everything. The old `rho` input is still accepted, and if provided is used both as the bulk and local density. 

Thanks to our co-op student @Sehmimul for this significant contribution!

Issue #706, opened 2021-04-23
Review all density effect correction files distributed with EGSnrc


ftessier:
We ought to verify all the `*.density` density effect correction files distributed with EGSnrc, under `HEN_HOUSE/pegs4/density_corrections/`, in light of the update to the estar.f code for ICRU Report 90. Also, in recent years we found that the NIST web interface to generate ESTAR data suffered some issues, specifically in regards to the user input for the mean excitation energy (I-value). It is unclear how the EGSnrc density correction files were generated originally, so at any rate a full review is warranted. Updated data should be consistent with ICRU Report 37, ICRU Report 90, and otherwise generated with the `estar.f` Fortran program, icru90 version.

See related #573 and #705 as well.

ftessier:
@bourgouina reminded me that one broad change in the icru90 version of `estar.f` done by @mainegra was the removal of the 13% boost in the I-value for mixtures (it only applies to molecular compounds). This was confirmed through discussion with the ESTAR code authors. So indeed most of the mixtures in the database will be affected. I also think we should really add the estar code to the repository to allow anyone to generate their own density files (I just have to sort out any licence or copyright issue!).

ftessier:
@mainegra if we can add `estar` to the repo, then I would suggest integrating the commit history for estar, and then just develop it further inside the EGSnrc repo. Another one for the chicken `coop/`! 🤓 . In time of course, ideally the density effect corrections would be calculated on the fly, like for _pegsless_ runs!

mainegra:
As a first step is great. But I think it should be integrated to EGSnrc's data generation mechanism.

PR #1396, opened 2026-04-20
Overhaul the estar code design

rtownson:
Convert the estar implementation to be integrated into egs++ and generally improve the design. This does change the density correction values of all materials, due to a bug fix in the bisection root solver.

Important changes:

1. The material density now actually reaches the I-value selection. The old code hardcoded double rho = 1; // Material density not actually needed, so use a dummy (old estarPreProcess.cpp:197); the new code passes the real density through compoundstoelements_ and tests it against const double rhocut = 0.1; (estar_formulaCalcs.cpp:291). Compound mixtures with true density ≤ 0.1 g/cm³ now get gaseous instead of condensed I-values — a genuine bug fix, with correspondingly different I and δ for such media.

2. The bisection now truly converges. The old loop exited on while (del > tolerance) where del was the signed objective value — i.e. on the first midpoint that overshot the root, regardless of magnitude. The new loop sets del = std::abs(fMid); (estar.cpp:1170) and iterates to |f| ≤ tolerance. Expect small shifts in δ at essentially every energy for estar-generated media. This is not expected to agree with the original fortran code, which uses an approximation.

3. calciumcarbonate.density reverts six values at the 4th significant digit (e.g. 4.00E-01,7.165E-02 becomes 7.164E-02), back to their pre-https://github.com/nrc-cnrc/EGSnrc/pull/888 state — presumably connected to https://github.com/nrc-cnrc/EGSnrc/issues/706, worth one line saying so.

## Example of impact
```
    :start air2:
        mixture compounds = H2O, O
        mass fractions = 0.9, 0.1
        bulk density = 0.00120479
        gas pressure = 2
        output density file = test-gas2.density
    :stop air2:

    :start carbon:
        elements = C
        number of atoms = 1
        bulk density = 2.0 # g/cm^3 used for transport
        local density = 2.0 # g/cm^3 used for density corrections
        output density file = test-carbon2.density
    :stop carbon:
```

image

image

## Testing against existing density files

All compound density files in EGSnrc were tested against the new algorithm, using 20 MeV electrons on a 1 cm slab in air. All density files agreed in energy deposited within uncertainties, except the following. The I-values from the original density files were used. Regions 0, 2 are air, and region 1 is the 1 cm slab of material. Status is "OK" if |Z score| < 2, otherwise "CHECK".

```
Medium                          Reg      E_new(MeV)  unc_new(%)  E_old(MeV)  unc_old(%)    AbsDiff(MeV)     %Diff   Z-score  Status

aluminumalloy_type2014            0    1.060100e-05     0.848    1.088400e-05     0.843   -2.830000e-07    -2.600    -2.203  CHECK
aluminumalloy_type2014            1    2.111300e-01     0.077    2.105200e-01     0.077    6.100000e-04     0.290     2.657  CHECK
aluminumalloy_type6061            0    9.705500e-06     0.885    1.006600e-05     0.873   -3.605000e-07    -3.581    -2.934  CHECK
aluminumalloy_type6061            1    1.967500e-01     0.079    1.956000e-01     0.079    1.150000e-03     0.588     5.247  CHECK
boroncarbide                      1    1.081900e-01     0.093    1.069500e-01     0.093    1.240000e-03     1.159     8.764  CHECK
boroncarbide                      2    7.704400e-04     0.094    7.746800e-04     0.094   -4.240000e-06    -0.547    -4.128  CHECK
copperalloy_freecuttingbrass      0    4.728400e-05     0.413    4.848300e-05     0.410   -1.199000e-06    -2.473    -4.303  CHECK
copperalloy_lowbrass              1    1.543300e+00     0.032    1.541700e+00     0.032    1.600000e-03     0.104     2.292  CHECK
copperalloy_lowbrass              2    1.114100e-03     0.079    1.118500e-03     0.079   -4.400000e-06    -0.393    -3.528  CHECK
copperalloy_manganesebronze       0    4.526300e-05     0.425    4.600500e-05     0.421   -7.420000e-07    -1.613    -2.718  CHECK
copperalloy_manganesebronze       2    1.118900e-03     0.079    1.122000e-03     0.079   -3.100000e-06    -0.276    -2.476  CHECK
copperalloy_redbrass              2    1.112900e-03     0.079    1.116700e-03     0.079   -3.800000e-06    -0.340    -3.051  CHECK
ferroboride                       0    3.187200e-05     0.467    2.938100e-05     0.530    2.491000e-06     8.478    11.564  CHECK
ferroboride                       1    1.174300e+00     0.034    1.096400e+00     0.039    7.790000e-02     7.105   133.158  CHECK
ferroboride                       2    1.123900e-03     0.071    1.132300e-03     0.078   -8.400000e-06    -0.742    -7.057  CHECK
leadoxide                         1    2.655100e+00     0.021    2.634300e+00     0.023    2.080000e-02     0.790    25.261  CHECK
nickelalloy__hastelloyb_          0    7.000400e-05     0.302    5.119500e-05     0.401    1.880900e-05    36.740    63.827  CHECK
nickelalloy__hastelloyb_          1    1.924100e+00     0.025    1.752800e+00     0.030    1.713000e-01     9.773   240.365  CHECK
nickelalloy__hastelloyb_          2    1.091000e-03     0.072    1.098100e-03     0.080   -7.100000e-06    -0.647    -6.025  CHECK
steel_stainless_type302           2    1.117400e-03     0.079    1.120000e-03     0.079   -2.600000e-06    -0.232    -2.080  CHECK
uraniumdicarbide                  1    3.205000e+00     0.020    3.202300e+00     0.020    2.700000e-03     0.084     2.980  CHECK
```


mainegra (approved):

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    bool getElemArray(char *formulaStr, int NEP, GetElementsStruct &GElem) {
+        const char delim = ' ';
+        std::vector<std::string> components;
+        tokenize(string(formulaStr), delim, components);
+
+        if (components.size() < (size_t)NEP) {
+            cout << "\n***************\n";
+            cout << "Error: List of elements is inconsistent with the number expected.\n";
+            cout << "Expected " << NEP << " elements but only found " << components.size() << ".\n";
+            cout << "\n***************\n";
```
ftessier: Convert stray `cout` to `egsInformation()`?

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    if (rho <= 0) {
+        egsFatal("estar::estarCalculation: Density must be greater than 0, got %g "
+                 "for medium %d.\n", rho, mediaNum);
+        return 9;
+    };
+
+    // Here we pass the processed data from the .egsinp file
+    // in getDataFromFormulae which computes the ivalue and other relevant quantities
+    fc = getDataFromFormulae(knmat, rho, elementArray, massFraction, numOfAtoms, NEP, mediaNum);
+    cout << "\n";
```
ftessier: Stray `cout`?

HEN_HOUSE/egs++/estar/estar_formulaCalcs.cpp
```diff
@@ -0,0 +1,466 @@
...
+            numberOfAtoms[i] = static_cast<int>(numOfAtoms[i]); // convert float to int
+            numberOfAtomsStr[i] = to_string(numberOfAtoms[i]); // convert int to string
+        }
+
+        // now we produce the final string
+        for (int i=0; i < NEP; ++i) {
+            compoundFormula = compoundFormula + elementArray[i] + numberOfAtomsStr[i];
+        }
+        cout << "\n";
+        cout << "The medium " << mediaNum << " is a compound of " << NEP << " elements with formula: " << compoundFormula << "\n";
```
ftessier: Stray `cout`?

HEN_HOUSE/interface/egs_interface2.h
```diff
@@ -770,6 +770,39 @@ extern __extc__ void egsHatch(void);
...
 
+/*! Shorthand notation for the \c estar subroutine (which is renamed to
+    \c egs_estar for the C/C++ interface using mortran's replacemant
+    capabilities) */
+#define estar F77_OBJ_(estar,ESTAR)
+/*! \brief Calculates estar density corrections
+
+
+ */
+extern __extc__ int estar_(char *formulaStr,
```
ftessier: ### No-op estar macro, hardcoded name mangling

I am confused here: this doesn't follow the pattern of the other definitions in this header (e.g. `egsHatch` above), where the macro defines a distinct C-side alias and the declaration is written *through* it, so that `F77_OBJ_` decides the name mangling. I would have expected something like:

```c++
#define egsEstar F77_OBJ_(egs_estar,EGS_ESTAR)
extern __extc__ int egsEstar(char *formulaStr, ...
```

As it stands, `#define estar F77_OBJ_(estar,ESTAR)` is a no-op: the macro can only replace the whole token `estar`, but the declaration below uses the different token `estar_`, and no code anywhere uses the bare token `estar`. The doc comment also says estar "is renamed to `egs_estar`", but no such
rename exists.

The `compoundstoelements` define below has the same issue.

Note the definitions in `estar.cpp` **don't include this header** and hardcode `extern "C" int estar_(...)`, so everything links only because gfortran mangles to `name_`.

I suggest to adopt the alias pattern: declare and define via `egsEstar`, include this header in `estar.cpp`, and (optionally, for the `egs_` prefix convention) rename the mortran call sites to `CALL egs_estar`;

But maybe I am missing a subtlety — it would not be the first time! 😆

### AI attribution

[AIA HAb Ce Se Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-HAb-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review; analysis verified and comment drafted by me, with AI copy-editing.

HEN_HOUSE/interface/egs_interface2.h
```diff
@@ -770,6 +770,39 @@ extern __extc__ void egsHatch(void);
...
+                      double *densityCorr,
+                      double *enGrid,
+                      int *NEP,
+                      int *ISCOMP,
+                      float *meanIval,
+                      float *ipotval,
+                      int *mediaID,
+                      char *outputFilename);
+
+#define compoundstoelements F77_OBJ_(compoundstoelements,COMPOUNDSTOELEMENTS)
```
ftessier: Same issue as for `estar_` above.

ftessier (changes requested):

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    // Validate the values pointed to
+    if (*ncomp <= 0 || *ncomp > 100) {
+        egsFatal("estar::compoundstoelements_: ncomp=%d is out of valid range "
+                 "[1, 100].\n", *ncomp);
+    }
+    if (*mediaDensity <= 0) {
+        egsFatal("estar::compoundstoelements_: mediaDensity=%f must be > 0.\n", *mediaDensity);
+    }
+
+    // In egsnrc.macros, the max number of elements per medium is $MXELE=50
```
ftessier: Typo: `$MXEL`

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    // And these arrays are 50 characters long, so we can restrict to that size
+    const size_t MAX_ELEMENT_STR_SIZE = 50 * 50; // 50 elements * 50 chars each
```
ftessier, line 334: ### Make the 50-element limit explicit

This constant quietly does double duty: the `charPos` check below (line 379) fires at element 51 only because each element occupies exactly 50 characters in the buffer — and that is currently the *only* thing stopping the loop at line 373 from writing past the Fortran-side arrays (`RHOZ`, `ZELEM`, `ASYM` are dimensioned `$MXEL = 50`). `fc.mmax` counts distinct elements across up to 100 compounds, so it can legitimately exceed 50.

The guard works today, so no big deal — but since this PR added explicit checks everywhere else, one more seems in scope, e.g. right after `mixtureCalculation()` (line 370):

```c++
// Fortran-side arrays (RHOZ, ZELEM, ASYM) hold at most $MXEL = 50 elements
if (fc.mmax > 50) {
    egsFatal("estar::compoundstoelements_: medium has %d distinct elements, "
             "but EGSnrc supports at most 50 ($MXEL).\n", fc.mmax);
}
```

### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

HEN_HOUSE/egs++/estar/estar_dataParser.h
```diff
@@ -0,0 +1,137 @@
...
+};
+
+/*! \brief Parses a chemical formula string into a parseformula object.
+ *
+ *  Accepts a single chemical formula string (e.g. "Na", "MgCl2", "H2O")
+ *  and splits it into its constituent element symbols and atom counts.
+ *
+ *  Prints an error message and asserts if the formula is incorrectly
+ *  formatted (e.g. "nA" instead of "Na" — element symbols must begin
+ *  with an uppercase letter).
```
ftessier: ### Dead assert includes, stale "asserts" doc

`<assert.h>` is included at `estar.cpp:45` and `estar_dataParser.cpp:41`, but nothing in the estar sources calls `assert()` — a leftover from an earlier iteration of the error handling? Also a leftover mention of assert at `estar.cpp:250`.

Relatedly, the doc for `parse()` (line 129 here) says it "prints an error message and asserts" on a malformed formula, but it actually aborts via `egsFatal()` (`estar_dataParser.cpp:141`) — the better behavior; the doc just needs to catch up.

```suggestion
 *  Prints an error message and aborts via egsFatal() if the formula is
 *  incorrectly formatted (e.g. "nA" instead of "Na" — element symbols must
 *  begin with an uppercase letter).
```

### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    bp.density_corr = g;
+    return bp;
+
+}
+
+scof fscof(int nmax, const vector<double> &x, const vector<double> &f) {
+
+    if (nmax < 2) {
+        egsFatal("estar::fscof: fscof requires at least 2 points, got nmax=%d\n", nmax);
+        return scof(0);
```
ftessier: ### Return after egsFatal

`return scof(0);` after `egsFatal()` is **unreachable** — `egsFatal` does not return. Harmless, but it muddies whether this path is survivable (it isn't).

### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issues surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    formula_calc fc = mixtureCalculation(rho,
+                                         estarFormulaArrayInput.data(),
+                                         estarWeightArrayInput.data(),
+                                         numCompounds);
+
+    // Set NEP to actually be the number of elements now, instead of the number of compounds
+    NEP[0] = fc.mmax;
+
+    size_t charPos = 0;
+    for (size_t i=0; i != fc.mmax; ++i) {
```
ftessier: ### Signed/unsigned comparison

`size_t i` is compared against the `int` member `fc.mmax`, which warns under `-Wsign-compare`; the sibling loop at line 358 already casts. Same fix here:

```suggestion
    for (size_t i=0; i != (size_t)fc.mmax; ++i) {
```

### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

HEN_HOUSE/egs++/Makefile
```diff
@@ -173,6 +179,9 @@ $(DSO1)egs_fortran_geometry.$(obje): egs_fortran_geometry.cpp egs_fortran_geomet
 $(egspp_objects):
 	$(obj_rule1)
 
+$(estar_objects):
```
ftessier, line 187: ### estar objects never rebuild after the first make

This `$(estar_objects):` rule has no prerequisites — not even the `.cpp` files. In make semantics, a target with a recipe and no prerequisites is considered up to date whenever the file exists, so after the first successful build, editing any estar source or header triggers no recompilation until a `make clean`. The egspp objects avoid this via their explicit per-object dependency lines (e.g. lines 99-100 for `egspp.o`).

Suggest per-object rules along the following lines — **This needs testing:**

```make
$(DSO1)estar.$(obje): $(ESTAR)estar.cpp $(ESTAR)estar.h $(ESTAR)estar_formulaCalcs.h $(ESTAR)estar_dataParser.h $(ESTAR)estar_dataTables.h $(config1h)
$(DSO1)estar_dataTables.$(obje): $(ESTAR)estar_dataTables.cpp $(ESTAR)estar_dataTables.h
$(DSO1)estar_dataParser.$(obje): $(ESTAR)estar_dataParser.cpp $(ESTAR)estar_dataParser.h $(ESTAR)estar_dataTables.h $(config1h)
$(DSO1)estar_formulaCalcs.$(obje): $(ESTAR)estar_formulaCalcs.cpp $(ESTAR)estar_formulaCalcs.h $(ESTAR)estar_dataParser.h $(ESTAR)estar_dataTables.h $(config1h)

$(estar_objects):
      $(CXX) $(INC1) $(ESTAR_INC) $(DEFS) $(opt) -c $(COUT)$@ $(ESTAR)$(notdir $(basename $@)).cpp
```

### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+        // section 2.3 of the report gives more detail about the range error.
+        if (y>=yq[0]) {
+            if (y-yq[lmax-1] <= 0) {
+                yl = log(y);
+                bp = fbspol(yl, yql, sf2.a, sf2.b, sf2.c, sf2.d);
+                if (solver == 1) {
+                    delta =  bp.density_corr;
+                }
+                else if (solver == 2) {
+                    xroot = bisec(q[bp.lb_index], q[bp.ub_index], tol, tau, f, eps, nmax);
```
ftessier, line 907: ### Out-of-bounds read on an exact top-of-grid hit

When `yl` falls exactly on the top grid point `yql[lmax-1]`, `fbspol` ends up reporting an upper index one past the grid. For an ascending grid the setup at lines 995-999 gives `idir = 0, mlb = 0, mub = n`, so the boundary test `s >= x[mub+idir-1]` (line 1006) is `s >= x[n-1]`, and an exact hit satisfies it. The extrapolation branch then computes `mu = mub + 2*idir - 1 = n - 1`, the unconditional `mu = mu + 1` at line 1034 advances it to `n`, and that is what comes back as `bp.ub_index`. The boundary test is not the only way there, either: for an exact hit the interior binary search converges to the same value, because equality goes right in `s < x[mav]`. Either way, `q[bp.ub_index]` on this line reads `q[lmax]` — one element past the end of the exactly-sized `vector<long double> q(lmax)` declared at line 543.

Worth noting the comment at line 903 says "y must be less than yq[lmax-1]" while the guard at 906 says `<=` — a mismatch inherited verbatim from the old estarMainCalc.cpp, so this boundary case seems to have never been decided. I'd argue the code has it right and the comment is what should change: `y == yq[lmax-1]` is exactly on the grid, where the spline is well defined — routing it to the `else` would abort a valid run with a misleading "too high and out of range" message. The bug is only that `fbspol` mishandles the boundary value it is legitimately given.

This is also not a regression: the old `bspol.cpp` has identical index arithmetic and only avoided a crash because `q` was a fixed `long double q[1200]` — the exact-hit case has always fed `bisec` an uninitialized bracket, silently. The exact-size vectors in this PR make the latent bug detectable, which is progress.

The trigger requires exact floating-point equality with the top grid point, so it is **rare** — but the fix is one clamp in `fbspol` right after `mu = mu + 1;` (line 1035), covering both the boundary branch and the binary-search equality path:

```c++
    // an exact hit on the top grid point lands past the last segment via
    // either branch above; clamp so ub_index stays a valid index
    if (mu > n - 1) {
        mu = n - 1;
    }
```

Suggest also updating the line-903 comment to `y must not exceed yq[lmax-1]`, and optionally making fbspol's `>=`/`<=` boundary tests strict so an exact on-grid value doesn't emit the misleading "extrapolating" warning; the clamp is needed regardless.

### AI attribution

[AIA HAb SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-HAb-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review; boundary-semantics analysis worked out jointly; comment drafted by AI, verified and approved by me.

HEN_HOUSE/egs++/estar/estar.cpp
```diff
@@ -0,0 +1,1173 @@
...
+    GetElements elemObject;
+    GetElements::GetElementsStruct GeElems;
+    int nepInt = *NEP;
+    int isCompInt = *ISCOMP;
+    int mediaNum = *mediaID; // this is the media id
+
+    // getElemArray now returns bool; propagate any parse failure back to the
+    // Fortran caller rather than crashing via assert.
+    if (!elemObject.getElemArray(formulaStr, nepInt, GeElems)) {
+        return 9;
```
ftessier: ### Parse failures return an error code the Fortran `CALL` never sees

The comment above states the intent — "propagate any parse failure back to the Fortran caller rather than crashing via assert" — but the Fortran side invokes these entry points as subroutines: `CALL estar(...)` (`pegs4_routines.mortran:1708`, `:1711`) and `CALL compoundstoelements(...)` (`get_media_inputs.mortran:1720`). A Fortran `CALL` discards the return value, so on that side there is no error path at all. In fact, a comment on the Mortran side notes that no checks are performed after the call to estar:

https://github.com/nrc-cnrc/EGSnrc/blob/fcbd83780e2e78a582e451d1f60250569b2fa5fc/HEN_HOUSE/src/pegs4_routines.mortran#L1718-L1719

1. **estar_ path.** When `getElemArray` fails (lines 97–103, e.g. a formula string with fewer components than `NEP`), `estar_` returns `9` and leaves `densityCorr`, `enGrid`, `IEV`, and `IPOTVAL` untouched. The caller has pre-set `IEV = 0` (`pegs4_routines.mortran:1706`), and the only place `IEV` is ever assigned is `*meanIval = fc.pot` (`estar.cpp:529`, via the 9th by-reference argument) — which this path never reaches. The run then proceeds with a zero I-value and zero density corrections: **silently wrong physics.**

2. **compoundstoelements_ path.** At line 350, `asym_tmp`, `rhoz_tmp`, `z_tmp`, and `nne_tmp` are never filled, yet the caller immediately does `nne(i)=nne_tmp;` (`get_media_inputs.mortran:1724`) using whatever that variable happened to hold.

**Suggested fix.** Every other error path in these files already uses `egsFatal`, so the minimal, consistent fix is to do the same here — most naturally inside `getElemArray` itself, replacing the `cout` block at lines 98–101 (which also bypasses the egs++ logger) with an `egsFatal` carrying the same expected-vs-found message. The two `return 9;` sites then become unreachable and can be removed, like the dead return already flagged at line 1056 (there's another after the `egsFatal` at line 510).


### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

ftessier (changes requested):
### Expected output changes, for the record

Documenting three places where this PR deliberately changes numerical output relative to current develop. None of these block the merge — ESTAR has never shipped in an official release, so this PR defines the first officially released ESTAR output — but they belong on the record for anyone comparing against develop:

1. **The material density now actually reaches the I-value selection.** The old code hardcoded `double rho = 1; // Material density not actually needed, so use a dummy` (old estarPreProcess.cpp:197); the new code passes the real density through `compoundstoelements_` and tests it against `const double rhocut = 0.1;` (estar_formulaCalcs.cpp:291). Compound mixtures with true density ≤ 0.1 g/cm³ now get gaseous instead of condensed I-values — a genuine bug fix, with correspondingly different I and δ for such media.

2. **The bisection now truly converges.** The old loop exited on `while (del > tolerance)` where `del` was the *signed* objective value — i.e. on the first midpoint that overshot the root, regardless of magnitude. The new loop sets `del = std::abs(fMid);` (estar.cpp:1170) and iterates to |f| ≤ tolerance. Expect last-digit shifts in δ at essentially every energy for estar-generated media.

3. **calciumcarbonate.density reverts six values at the 4th significant digit** (e.g. `4.00E-01,7.165E-02` becomes `7.164E-02`), back to their pre-#888 state — presumably connected to #706, worth one line saying so.

@rtownson, could you add these, or a summary of these, to the PR description? A before/after diff of generated density-correction files for one element, one compound, and one gaseous compound mixture would also pin down (1) and (2) empirically and close the testing question.

### AI attribution

[AIA PAI SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-PAI-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Findings surfaced by AI-assisted review and comment drafted by AI; analysis and text verified, edited, and approved by me.

HEN_HOUSE/makefiles/standard_makefile
```diff
@@ -87,8 +87,8 @@ failed_message = "______Operation failed______"
 $(EXECUTABLE): $(FORTRAN_FILE).$(FEXT) $(EGS_EXTRA_OBJECTS) $(EXE_DIR) compare
 	@echo $(empty)
 	@echo $(empty)
-	@echo "Fortran compiling $(FORTRAN_FILE).$(FEXT) using flags '$(EGS_EXTRA_FLAGS) $(FCFLAGS) $(OPTLEVEL_F)' and extra objects/libs '$(EGS_EXTRA_OBJECTS) $(ESTAR_DIR)estarPreProcess.cpp $(F77_CPP_LINKER) $(EGS_EXTRA_LIBS)'"
-	@$(F77) $(EGS_EXTRA_FLAGS) $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH) $(FOUT)$@ $(FORTRAN_FILE).$(FEXT) $(EGS_EXTRA_OBJECTS) $(ESTAR_DIR)estarPreProcess.cpp $(F77_CPP_LINKER) $(EGS_EXTRA_LIBS)
+	@echo "Fortran compiling $(FORTRAN_FILE).$(FEXT) using flags '-cpp $(EGS_EXTRA_FLAGS) $(FCFLAGS) $(OPTLEVEL_F)' and extra objects/libs '$(EGS_EXTRA_OBJECTS) $(EGS_EXTRA_LIBS) -legspp'"
+	@$(F77) -cpp $(EGS_EXTRA_FLAGS) $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH) $(FOUT)$@ $(FORTRAN_FILE).$(FEXT) $(EGS_EXTRA_OBJECTS) $(EGS_EXTRA_LIBS) -legspp
```
ftessier: ### Plain Mortran codes fail to link: -legspp has no library path

With this PR, every Fortran application now links against `libegspp` — but this line only tells the linker *that* it needs the library (`-legspp`), never *where to find it*: there is no `-L$(abs_dso)` here, and no library path added anywhere else on the Fortran side. The library lives in the egs++ dso directory, which is not somewhere `ld` looks on its own.

The reason this is easy to miss is that the familiar applications still build: anything whose `.make` pulls in `$(BEAMLIB_EXTRA_LIBS)` (the RZ codes, dosxyznrc, BEAM) gets the path by accident, because `IAEA_LIB = -L$(abs_dso) -Wl,-rpath,$(abs_dso) -liaea_phsp` happens to ride along on the same link line. The codes with no such luck are the simple ones: tutor1–7, `g`, `examin`, `ranlux_test`, `ranmar_test` (no `EGS_EXTRA_LIBS` at all), and `cavsphnrc`/`edknrc` (`EGS_EXTRA_LIBS = $(FLIBS)`, and `FLIBS` is empty in a standard gfortran configuration) — those get the bare `-legspp` and fail.

Reproduced on a fresh clone of this branch with a full `./configure` (gfortran/gcc/g++, Debian, in a clean container). The configuration itself completes — including "Building the EGSnrc C++ class library ... OK", so `libegspp.so` is built and present in `$(abs_dso)` — and then:

```
Fortran compiling tutor1_fcbd8378.F using flags '-cpp  -fPIC -O2 -mtune=native -mcmodel=medium -std=legacy' and extra objects/libs '  -legspp'
/usr/bin/ld: cannot find -legspp: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [.../HEN_HOUSE/makefiles/standard_makefile:91: .../egs_home/bin/fcbd8378/tutor1] Error 1
```

A fresh user choosing "Compile all" at the end of the installation hits this during the install itself. Adding the path and rpath fixes it — verified in a second, independent build of this branch, where tutor1 then links (RUNPATH embedded) and completes a full simulation:

```suggestion
	@$(F77) -cpp $(EGS_EXTRA_FLAGS) $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH) $(FOUT)$@ $(FORTRAN_FILE).$(FEXT) $(EGS_EXTRA_OBJECTS) $(EGS_EXTRA_LIBS) -L$(abs_dso) -Wl,-rpath,$(abs_dso) -legspp
```
**The echo on line 90 should mirror the change.** The same fix applies to `beam_makefile:98`. The cleaner long-term shape is probably an `EGSPP_LIB = -L$(abs_dso) -Wl,-rpath,$(abs_dso) -legspp` variable emitted into the main config next to `IAEA_LIB`: configure already generates exactly that `-L$(abs_dso) -Wl,-rpath,$(abs_dso)` string per platform for the C++ config.

The PR effectively bolted a C++ library onto the Fortran link without importing the C++ side's path machinery, and an `EGSPP_LIB` closes that gap where platform differences are already handled. Note the PR also removed the old `F77_CPP_LINKER` hook from `all_common.spec`, so there is currently no spec-level variable left to carry this.

### AI attribution

[AIA HAb SeCe Hin R Claude Fable 5](https://aiattribution.github.io/statements/AIA-HAb-SeCe-Hin-R-?model=Claude%20Fable%205-v1.0)

Issue surfaced and build-verified by AI-assisted review; independently reproduced by me on a fresh clone and configure; comment drafted by AI, verified, edited, and approved by me.

HEN_HOUSE/makefiles/beam_makefile
```diff
@@ -94,8 +94,8 @@ EXE_DIR = $(EGS_HOME)bin$(DSEP)$(my_machine)
 $(EXE_FILE): $(FORTRAN_FILE).$(FEXT) $(EXE_DIR)
 	@echo $(empty)
 	@echo $(empty)
-	@echo "Fortran compiling $(FORTRAN_FILE).$(FEXT) using flags '-cpp $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH)' and extra objects/libs '$(BEAM_EXTRA_OBJECTS) $(ESTAR_DIR)estarPreProcess.cpp -lstdc++ $(FLIBS)'"
-	@$(F77) -cpp $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH) $(FOUT)$@ $(FORTRAN_FILE).$(FEXT) $(BEAM_EXTRA_OBJECTS) $(ESTAR_DIR)estarPreProcess.cpp -lstdc++ $(FLIBS)
+	@echo "Fortran compiling $(FORTRAN_FILE).$(FEXT) using flags '-cpp $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH)' and extra objects/libs '$(BEAM_EXTRA_OBJECTS) $(FLIBS) -legspp'"
+	@$(F77) -cpp $(FCFLAGS) $(OPTLEVEL_F) $(COMPILE_TIME) $(GIT_HASH) $(FOUT)$@ $(FORTRAN_FILE).$(FEXT) $(BEAM_EXTRA_OBJECTS) $(FLIBS) -legspp
```
ftessier: See comment for `standard_makefile:91`.

ftessier (changes requested):

HEN_HOUSE/interface/egs_interface2.h
rtownson: Good catch! But including interface2.h is designed for egs applications, e.g. it requires array_sizes.h. So instead, I'm going to put the `extern __extc__ int egsEstar` directly into estar.cpp, and bypass the interface.

rtownson:
I've finished the revisions suggested by @ftessier. Since this changes the density corrections slightly due to a fix to the bisection algorithm, @mainegra could you re-run your test simulation? I'll also do some additional testing.

ftessier (approved):

PR #1400, opened 2026-04-24
Fix egs_view trying to load non-egs++ apps

rtownson:
Fix a segmentation fault in egs_view happening in the current `develop` branch, that only occurs when you have compiled non-egs++ apps as shared libraries (e.g. a BEAM accelerator). 

This bug never made it into `master`, so does not affect any release versions.


rtownson:
This commit is included in PR #1401, so closing it here.

PR #1401, opened 2026-04-24
Fix dynamic geometry critical bug in some apps

rtownson:
Fix a critical bug where dynamic geometries were visualized in egs_view, but the particles were transported in a static geometry. This was always the case in `egs_cbct`, `egs_chamber`, `egs_fac`, `egs_kerma`, and `cavity`. It was due to those applications overwriting the `simulateSingleShower` function, and thus not including the new changes to move geometries.

Any users with custom applications that overwrite (and don't call) the application class `simulateSingleShower` function should be updated to support motion (it's two lines to add).

Simulations in other applications were not affected.

Thanks to Patricia Oliver for reporting this!


ftessier:
@rtownson I think this includes all of #1400, can we close that one and just merge here?

mainegra (approved):

ftessier (approved):

PR #1407, opened 2026-05-06
Fix compiler error in egs_ng_geometry

rtownson:
Fix a compiler error that only occurred in stricter or older compilers in egs_nd_geometry.cpp. Was discovered on gcc 4.8.5.

Thanks to Jarkko for reporting this.


ftessier (changes requested):
- [x] Typo in commit title "ng" geometry instead of "nd".
- [x] Consider squashing the commits as `Fix string arguments passed to egsFatal/egsWarning`

mainegra (approved):

ftessier:
I am squashing on develop.

ftessier (approved):

PR #1425, opened 2026-06-21
Fix #934: expand environment variables in include file paths

### LINKED ISSUE #934: Expand environment variables in include files
Expanding environment variables in various egs++ applications has been added a while ago, but what about for `include file` keys in egsinp files? It does not seem to be supported (as far as I can tell...) and looking at the source in `egs_input.cpp`, I have no idea where the call to `egsExpandPath()` should be placed.

Can someone add support for this, please?

mchamberland:
Fixes #934.

Applies `egsExpandPath()` when opening `include file` paths in `addContentFromFile()` and inline include-file handling, matching behaviour of other egs++ file-path inputs (e.g. spectrum files, beam sources).

After this change, paths such as `$EGS_HOME/egs_brachy/lib/geometry/my.geom` or `%HEN_HOUSE%/pegs4/data/521icru.pegs4dat` work in `include file` statements. Only the first environment variable is expanded; it must be followed by `/` or `\`.

Documentation added to PIRS-898 common input syntax (`common.doxy`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/HEN_HOUSE/doc/src/pirs898-egs++/common.doxy
+++ b/HEN_HOUSE/doc/src/pirs898-egs++/common.doxy
@@ -59,12 +59,14 @@ and variance reduction techniques vary between applications.
 
 <em>Tip 2: All distances are in units of cm, and energies in MeV.</em>
 
-\subsection common_includefile Include files
+\subsection common_includefile Include files and file paths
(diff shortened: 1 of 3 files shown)


mchamberland:
Follow-up commit: pegsless `material data file` and `density correction file` now call existing Fortran `replace_env()` (same mechanism as dosxyznrc/beamnrc spectrum paths). PIRS-898 docs updated.

mchamberland:
Commit d5a17de accidentally copied `get_media_inputs.mortran` (and an older `common.doxy`) from the CLRP fork onto `develop`, reverting newer pegsless media documentation and other develop-only changes. That was a packaging mistake, not intentional.

**Fixed:** force-pushed `0b5c4866` — only **+2 lines** in `get_media_inputs.mortran` (`call replace_env` for material data file and density correction file) and a **+1 sentence** in PIRS-898. All develop media-definition docs are preserved.

rtownson (approved):

PR #1430, opened 2026-07-07
Fix #1422: Electric field work not applied to energy of charged particles

### LINKED ISSUE #1422: Combined bug report: Electric field work is not correctly applied and appears 100× too large when applied
**Describe the bug**
Dear EGS community,

There appear to be two related issues when using EGSnrc with egs_app and EEMF_macros.mortran for non-zero electric fields.

First, in **egsnrc.mortran**, the E-field work is computed and added to _de_, but the final particle energy update uses _edep_, which was saved before the electric-field work was added (see line 1588-1595):

            "Now done with multiple scattering,
            "update energy and see if below cut
            "below subtracts only energy deposited"
            peie  = peie - edep;
            "below subtracts energy deposited + work due to E field"
            "peie = peie - de;"   **<-- This line should not be commented.**
            eie   = peie;
            e(np) = peie;

In essence, this means that the E-field does no work on the particle.

Secondly, Once the _peie = peie - de;_ is included the energy change is applied, the magnitude appears to be 100x too large in the default **EEMF_macros.mortran**.

The **EEMF_macros.mortran** normalizes the E-field vector, stores its magnitude in _E_amp_, and then scales the magnitude, and unfortunately, also the normalized vector components (see line 289-291):

"Adjust the electric field to be in V/m instead of the V/cm"
  Ex = Ex * 100.;Ey = Ey * 100.;Ez = Ez * 100.;
  E_amp = E_amp*100.; 

Only the magnitude should be converted. 

**To Reproduce**
1. Build an egs application with EEMF_macros.mortran.
- Have the application print the kinetic energy of the particle in steps of your desire (e.g. when changing regions).
- My example is here: 

[egs_app.cpp](https://github.com/user-attachments/files/29136252/egs_app.cpp)

2. Use a simple geometry with region boundaries along Z in which electrons travels  along positive Z direction.
- My example is here (not the egsinp format has been changed to txt for upload here): 

[eemf_bug_test.txt](https://github.com/user-attachments/files/29136278/eemf_bug_test.txt)

A tip is to have the material travel in air with "set relative density = 1e-10" (to ensure minimal energy loss to collisions).
3. (Remember, that for E-fields to be included, the magnetic field must be defined as well, so set magnetic field = 0,0,0)
4. Set a constant E-field in the Z-direction.
5. First run with original **egsnrc.mortran**.
6. Next run wit modified **egsnrc.mortran**:
Line 1593: "peie = peie - de;" **-->** peie = peie - de;

--> 

7. Modify the EEMF_macros.mortran: 
Line 290: Ex = Ex*100.;Ey = Ey*100.;Ez = Ez*100.;  **-->** "Ex = Ex*100.;Ey = Ey*100.;Ez = Ez*100.;"
8. Run with modified **EEMF_macros.mortran**.

**Expected behavior**
For step 1-5 above, the expected behavior is the the kinetic energy (printed from the egs_application) should not change.

For step 6, the kinectic energy should now change according to ΔK = q E d. However, with a magnitude 100x larger than expected (due to EEMF_macros bug).
, where ΔK is the change in kinetic energy, q is the particle charge, E is the electric field strength, and d is the distance travelled.
ΔK = q E d = -1.60217663*10^(-19) coulomb * -10000 (V/cm) * 1 cm * 100 = 1 MeV

For step 8, the energy change should be as expected.
ΔK = q E d = -1.60217663*10^(-19) coulomb * -10000 (V/cm) * 1 cm  = 0.01 MeV

**Screenshots**
I've setup a box containing air (with relative density 1e-10) and 3 regions. The electron kinetic energy is 1 MeV.

Image

When running step 1-5, the energy change for two particles is unchanged. 
(note that regions (0,1,2) in the screenshot above corresponds to regions (2,3,4) in the output below)

Image

When running step 6, the output is:

Image

When running step 8, the output is:

Image

**Operating system**

- [X] Linux
- [ ] Windows
- [ ] macOS
- [ ] Other: 

**EGSnrc version**
v2025

**Additional context**
Thank you for developing and maintaining EGSnrc :)

rtownson:
Fix a bug where the electric field did not subtract energy from charged particles when work was done. 

Also, fix a mistake in the magnitude of the electric field that was noticeable now that the work was applied.

Thanks to @LouisLechNissen for finding this issue and the solution.

### DIFF
--- a/HEN_HOUSE/src/EEMF_macros.mortran
+++ b/HEN_HOUSE/src/EEMF_macros.mortran
@@ -287,7 +287,6 @@ REPLACE{$EMFIELD_INITIATE_SET_TUSTEP;}WITH{;
   E_amp = r0_em;
 
   "Adjust the electric field to be in V/m instead of the V/cm"
-  Ex = Ex*100.;Ey = Ey*100.;Ez = Ez*100.;
   E_amp = E_amp*100.;
 
   "activate the transport for the rest of the calculation"
--- a/HEN_HOUSE/src/egsnrc.mortran
+++ b/HEN_HOUSE/src/egsnrc.mortran
@@ -1588,9 +1588,9 @@ $start_new_particle;
             "Now done with multiple scattering,
             "update energy and see if below cut
             "below subtracts only energy deposited"
-            peie  = peie - edep;
+            "peie  = peie - edep;"
             "below subtracts energy deposited + work due to E field"
-            "peie = peie - de;"
+            peie = peie - de;
             eie   = peie;
             e(np) = peie;
 


LouisLechNissen:
Dear Reid and team,
 
I believe there is an error in the bug I've previously reported, and the solution.

  "below subtracts only energy deposited"
            peie  = peie - edep;
            "below subtracts energy deposited + work due to E field"
            "peie = peie - de;"
            peie = peie - de;
            eie   = peie;
            e(np) = peie;

should be  

  "below subtracts only energy deposited"
            **"peie  = peie - edep;"** <-- Should be commented while peie - de; is not.
            "below subtracts energy deposited + work due to E field"
            "peie = peie - de;"
            peie = peie - de;
            eie   = peie;
            e(np) = peie;


Or else the deposition loss is subtracted twice as far as I can understand. Should be double-checked.

Best regards,
Louis

rtownson:
Yes I agree that the `peie = peie - edep;` line needed to be commented out or `edep` gets subtracted twice. I've fixed this, thanks @LouisLechNissen.

HEN_HOUSE/src/EEMF_macros.mortran
```diff
@@ -287,7 +287,6 @@ REPLACE{$EMFIELD_INITIATE_SET_TUSTEP;}WITH{;
   E_amp = r0_em;
 
   "Adjust the electric field to be in V/m instead of the V/cm"
-  Ex = Ex*100.;Ey = Ey*100.;Ez = Ez*100.;
```
ftessier, line 290: @LouisLechNissen, is that consistent with the prescribed input units? I am unsure about this one, will check.

ftessier: ✅  Ok, so the input units are cleary `V/cm`, from docs and code, as they should be (cm default convention).

LouisLechNissen: Yes, the issue is that this is after normalization such that this these are not unit vectors anymore.

ftessier:
A bit of archeology, for the record:

- **Pre-2015 (EGS4-era emf_macros.mortran):** the original `$ADD-WORK-EM-FIELD` adds the potential-difference work to `DE` (emf_macros.mortran:306-314), expecting the core to subtract `de`. But at the 2015 GitHub import (9327c077), the active line was already `peie = peie - edep`. So E-field work in media has been discarded for the entire recorded history of the repo, old macros included. The old macros wired vacuum correctly themselves: `$VACUUM-ADD-WORK-EM-FIELD` does `PEIE=PEIE-DE` directly.

- **2015-11-25** (98c00821, "Insert hook for including work done by E field"): added the commented-out `"peie = peie - de;"` alternative plus the two explanatory comments, documenting the disconnect as an opt-in hook rather than fixing it. Nothing ever activated it.

- **2019-03-19** (cfbb245c, "Add enhanced electromagnetic transport macros"): the new EEMF `$ADD_WORK_EM_FIELD` tacks `de_EM` onto `de`. Its own comment says "This `de_EM` is tacked on to `de` in `ELECTR`". It was written trusting the previous hook, which was still commented out. The work term was computed and thrown away from day one. The same commit scaled the normalized field direction components by 100 along with the magnitude (EEMF_macros.mortran:290), breaking the unit-vector invariant assumed downstream.

- **2019-12-13** (7515d4ad): the BEAMnrc EEMF variant was added with the magnitude-only scaling (EEMF_macros_beamnrc.mortran:324), so the two macro files have disagreed since; this PR aligns them.

ftessier (approved):

ftessier:
@mainegra, let's get this one merged too!

mainegra (approved):

PR #1443, opened 2026-08-12
Improve egs_editor performance, fix egs_editor library names

rtownson:
Fixes performance issues for very large .egsinp files in egs_editor. This solves lag issues in egs_view for such files, by lazy-loading the input file, and only syntax checking on the current view window in the editor.

This also fixes an issue where all shared libraries inserted capitalized library names in egs_editor, which resulted in a crash when actually running the simulation. Now the capitalization of the library names matches the actual shared library files.

Finally, it also adds in egs_editor autocomplete support for some libraries that were not supported yet (anything that was new in the develop branch since egs_editor was added). This looks like changes across a lot of files, but the change is very minor and only impacts egs_view.

### DIFF
--- a/HEN_HOUSE/egs++/geometry/egs_vhp_geometry/egs_vhp_geometry.cpp
+++ b/HEN_HOUSE/egs++/geometry/egs_vhp_geometry/egs_vhp_geometry.cpp
@@ -67,6 +67,8 @@
 
 using namespace std;
 
+static bool EGS_VHP_LOCAL inputSet = false;
+
 #ifndef SKIP_DOXYGEN
 EGS_VoxelInfo *EGS_VoxelGeometry::v = 0;
 int            EGS_VoxelGeometry::nv = 0;
@@ -671,6 +673,48 @@ const static EGS_VHP_LOCAL char *vhp_error_msg1 =
     "createGeometry(VHP): wrong/missing %s input for micro_test geometry\n";
 
 extern "C" {
+
+    static void setInputs() {
+        inputSet = true;
+
+        setBaseGeometryInputs(false);
+
+        geomBlockInput->getSingleInput("library")->setValues({"egs_vhp_geometry"});
+
+        // Format: name, isRequired, description, vector string of allowed values
+        geomBlockInput->addSingleInput("phantom data", true, "The full filepath to the phantom data file, including the extension.");
+        geomBlockInput->addSingleInput("media data", true, "The full filepath to the media data file, including the extension.");
+        geomBlockInput->addSingleInput("slice range", false, "Two numbers, slice indices (indexed from 0). The first number must be smaller. Used to select a given slice range instead of using the entire phantom.");
+
+        // For bone spongiosa
+        geomBlockInput->addSingleInput("BSC thickness", false, "The thickness of the bone surface cells layer, in cm.");
+        geomBlockInput->addSingleInput("TB medium", false, "The trabecular bone medium name.");
+        geomBlockInput->addSingleInput("TB medium", false, "The bone marrow medium name.");
+        geomBlockInput->addSingleInput("micro matrix", false, "The binary micro matrix filename followed by organ indices (defined in the media data file).");
+    }
+
+    EGS_VHP_EXPORT string getExample(string type) {
+        string example;
+        example = {
+            R"(
+    # For file format definitions and more inputs, see the documentation.
+    #:start geometry:
+        library      = egs_vhp_geometry
+        name         = some_name
+        phantom data = phantom_data_file
+        media data   = media_data_file
+        slice range  = min_slice max_slice
+    :stop geometry:
+)"};
+        return example;
+    }
+
+    EGS_VHP_EXPORT shared_ptr<EGS_BlockInput> getInputs() {
+        if(!inputSet) {
+            setInputs();
+        }
+        return geomBlockInput;
+    }
 
     EGS_VHP_EXPORT EGS_BaseGeometry *createGeometry(EGS_Input *input) {
         string type;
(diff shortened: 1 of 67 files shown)


HEN_HOUSE/egs++/shapes/egs_dynamic_shape/egs_dynamic_shape.cpp
```diff
@@ -39,8 +39,54 @@
...
+static bool EGS_DYNAMIC_SHAPE_LOCAL inputSet = false;
+static shared_ptr<EGS_BlockInput> EGS_DYNAMIC_SHAPE_LOCAL shapeBlockInput = make_shared<EGS_BlockInput>("shape");
+
 extern "C" {
 
+    static void setInputs() {
+        inputSet = true;
+
+        setShapeInputs(shapeBlockInput);
+        shapeBlockInput->getSingleInput("library")->setValues({"egs_conical_shell"});
```
ftessier: This looks wrong, `egs_conical_shell` is probably a mispaste?

ftessier:
The string `egs_box` was lowercased, but not the string for the cd geometry. It probably should to match the pattern updated across the library:

```rg
egs_box/egs_box.cpp
49:static string EGS_BOX_LOCAL typeStr("egs_box");
84:        geomBlockInput->getSingleInput("library")->setValues(vector<string>(1, typeStr));

egs_cd_geometry/egs_cd_geometry.cpp
51:static string EGS_CDGEOMETRY_LOCAL typeStr("EGS_CDGeometry");
123:        geomBlockInput->getSingleInput("library")->setValues(vector<string>(1, typeStr));
```

Should the cd geometry not match the pattern, something like the following, modelled after the egs_nd_geometry case:

```c++
geomBlockInput->getSingleInput("library")->setValues({"egs_cdgeometry"});
```

HEN_HOUSE/egs++/geometry/egs_vhp_geometry/egs_vhp_geometry.cpp
```diff
@@ -672,6 +674,48 @@ const static EGS_VHP_LOCAL char *vhp_error_msg1 =
...
+
+        // Format: name, isRequired, description, vector string of allowed values
+        geomBlockInput->addSingleInput("phantom data", true, "The full filepath to the phantom data file, including the extension.");
+        geomBlockInput->addSingleInput("media data", true, "The full filepath to the media data file, including the extension.");
+        geomBlockInput->addSingleInput("slice range", false, "Two numbers, slice indices (indexed from 0). The first number must be smaller. Used to select a given slice range instead of using the entire phantom.");
+
+        // For bone spongiosa
+        geomBlockInput->addSingleInput("BSC thickness", false, "The thickness of the bone surface cells layer, in cm.");
+        geomBlockInput->addSingleInput("TB medium", false, "The trabecular bone medium name.");
+        geomBlockInput->addSingleInput("TB medium", false, "The bone marrow medium name.");
```
ftessier: I think this second input shoud be BM medium. See [line 297.](https://github.com/nrc-cnrc/EGSnrc/blob/9edee3ebfda3d81d0e8eb033a7e76bf9a70e41ef/HEN_HOUSE/egs%2B%2B/geometry/egs_vhp_geometry/egs_vhp_geometry.cpp#L297)

HEN_HOUSE/egs++/sources/egs_focal_spot_source/egs_focal_spot_source.cpp
```diff
@@ -193,6 +195,64 @@ void EGS_FocalSpot::setUp() {
...
+        srcBlockInput->addSingleInput("z position", true, "The z position of the source");
+        srcBlockInput->addSingleInput("spatial spread x", true, "The standard deviation of the source along x");
+        srcBlockInput->addSingleInput("spatial spread y", true, "The standard deviation of the source along y");
+        srcBlockInput->addSingleInput("spatial cutoff x", false, "Particles will not be generated outside [x0-cutoff, x0+cutoff]");
+        srcBlockInput->addSingleInput("spatial cutoff y", false, "Particles will not be generated outside [y0-cutoff, y0+cutoff]");
+        srcBlockInput->addSingleInput("angular spread x", false, "The standard deviation in degrees from the z-axis toward x");
+        srcBlockInput->addSingleInput("angular spread y", false, "The standard deviation in degrees from the z-axis toward y");
+        srcBlockInput->addSingleInput("x translation", false, "An offset from the origin along x");
+        srcBlockInput->addSingleInput("y translation", false, "An offset from the origin along y");
+        srcBlockInput->addSingleInput("z translation", false, "An offset from the origin along z");
```
ftessier: The input `z translation` doesn't exist; the parser reads `z of rotation` (line 109). The example text has it right, so the editor would red-underline a correct input.

ftessier:
Caught by Claude (Fable 5), not sure this is right, but worth a look and a test:

- **Bug:** `loadInput()` now does `editorLoaded = false; egsinpEdit->clear()`; (lines 1196-1197), and reloading only happens from the `currentChanged` tab signal. Then `saveEgsinp()` calls `reloadInput()` (line 1950) and `updateSimulationGeometry()` calls `loadInput(true, g)` (line 2294). If the user is already on the editor tab (which is exactly when you'd save), `currentChanged` never fires and the editor goes blank until they switch tabs away and back. Fix is a one-liner: after the reset, re-trigger `ensureEditorLoaded` when `tabWidget->currentWidget() == tab`.

- **Minor:** `ensureEditorLoaded()` from `saveEgsinp()` validates against a hidden viewport, and nothing re-validates on first show afterwards; syntaxHighlighter/setHighlighter added to EGS_Editor but never used; `egs_editor.h` now includes `egs_highlighter.h` where a forward declaration would do.

- **Gaps (non-blocking):** scale missing for `egs_mesh` and `egs_triangle_mesh` (plus normals, octree accelerate); type missing for `egs_lattice` (hexagonal).

- **Nit:** the four new `getExample(string type)` signatures are called through the no argument `getExampleFunction` typedef, a pre-existing pattern in 15 other libs, so not new, but new code could use `getExample()`.

ftessier (changes requested):

ftessier (approved):
Upon addressing the ealier comments, I approve the merge!

mainegra (approved):

PR #1449, opened 2026-09-11
Port the HEN_HOUSE GUIs to Qt 6 while keeping Qt 5 support

mchamberland:
## Summary

`egs_view` was ported to Qt 6 previously, but the three Qt applications under `HEN_HOUSE/gui` — `egs_gui`, `egs_inprz` and `egs_configuration` — were not, and currently do not build against Qt 6. This ports all three using the same approach as `egs_view`: conditional compilation only around APIs that actually changed, so one source tree builds against both Qt 5 and Qt 6.

42 files changed, +527 / −1738 (most of the deletions are `.ui` connection blocks, see below).

### Widget connections moved out of the `.ui` files (90 connections)

The `<connections>` blocks in `inputRZ.ui`, `executiondialog.ui`, `pegs_page.ui` and `pegs_runoutput.ui` referenced slots implemented on the subclass rather than on the generated form class. Qt 6's `uic` no longer produces working wiring for that arrangement, so all 90 connections were moved into the corresponding constructors.

This is the largest and highest-risk part of the change. I verified mechanically that every one of the 90 removed connections has a counterpart in C++ (each sender/slot pair appears in a `connect()` call in the same application), and confirmed at runtime that no connection fails to resolve.

### Signals removed in Qt 6

Twenty-eight connections used signals that no longer exist in Qt 6. These deserve attention because they use the string-based `SIGNAL()`/`SLOT()` syntax: the code still compiles, and `QObject::connect()` only fails at runtime, so the affected controls would silently stop working.

- `QComboBox::activated(QString)` / `highlighted(QString)` — where the receiving slot ignores the payload, these now use the surviving `int` overload (16 sites, no version guard needed, matching connects already present nearby). Where the text is needed, they use typed connections: `textActivated`/`textHighlighted` from Qt 5.14, with a `QOverload<const QString &>` fallback below that.
- `QComboBox::currentIndexChanged(const QString &)` → `currentTextChanged` for the four compiler combo boxes in `egs_configure`. These are never editable, so the emission points coincide.
- `QButtonGroup::buttonClicked(int)` → `idClicked(int)` from Qt 5.15.

### Enum and API moves covering the whole Qt 5 range

- `Qt::SkipEmptyParts` and `Qt::endl` exist only from Qt 5.14 on, while `QString::SkipEmptyParts` and the global `endl` they replaced are gone in Qt 6, so neither spelling covers the supported range. A local alias selects the right one at compile time (24 sites).
- `QRegExp` → `QRegularExpression`
- `QTime` → `QElapsedTimer` for elapsed timing
- `QPrinter::Letter` → `QPageSize::Letter` under Qt 6
- `Q3WhatsThis` → `QWhatsThis`
- `QLayout::setMargin(n)` → `setContentsMargins(n, n, n, n)`
- `Qt::WindowFlags f = 0` → `= {}`
- `foreach` → range-based `for`
- Obsolete `QMessageBox` signatures updated
- `QComboBox::setAutoCompletion` removed from the `.ui` files and restored for Qt 5 only in code
- `<weight>50</weight>` removed from `inputRZ.ui`, since Qt 6's `QFont::setWeight` takes `QFont::Weight` rather than `int`

### Build files

`greaterThan(QT_MAJOR_VERSION, 5): QT += core5compat` added to the three `.pro` files, Qt 6 only.

### macOS-specific Qt 6 startup crash in `egs_configuration`

`egs_configuration` aborted on startup under Qt 6 on macOS. This is not a porting bug — a bare twelve-line `QWizard` crashes identically, because Qt 6's `QWizardPrivate::findDefaultBackgroundPixmap()` resolves `com.apple.KeyboardSetupAssistant`, which no longer exists on current macOS, and passes the resulting nil URL to `NSBundle`. Qt 5 does not crash. There is no "no background pixmap" option, so supplying a 1×1 transparent pixmap skips the lookup; guarded to macOS + Qt 6. The wizard's `sizeHint` is unchanged from the Qt 5 baseline.

Also fixed along the way: `setFrameStyle(QFrame::StyledPanel || QFrame::Sunken)` used a logical-or where a bitwise-or was meant.

## Qt versions

- **Qt 6** — verified against 6.11.1.
- **Qt 5** — verified against 5.15.19. The pre-5.14 branches are compile-tested (by temporarily forcing them on), but I have no genuinely old Qt 5 available, so that path is compile-verified rather than run.

## Testing

- Clean builds of all three applications under Qt 5.15.19 and Qt 6.11.1 on macOS 26 (arm64): no errors, and no deprecation warnings on any of the changed connections.
- All three launched offscreen under both toolkits with zero `No such signal` / `No such slot` messages. The detection method was validated with a deliberately broken connect as a positive control.
- Mechanical verification that all 90 relocated connections exist in C++.
- Interactive testing of all three applications.

**Not built on Linux or Windows.** A check on those platforms would be welcome, particularly the static Linux build of `egs_configuration`, which is the one that populates `HEN_HOUSE/pieces/linux`.

## Notes for reviewers

The 16 combo connections switched to `activated(int)`/`highlighted(int)` and the four switched to `currentTextChanged` are deliberately **not** version-guarded, so they change behaviour on Qt 5 as well. The reasoning is that `activated(int)` and `activated(QString)` are emitted from the same place and the receiving slots ignore the argument, and that `currentTextChanged` coincides with `currentIndexChanged` for non-editable combo boxes. Worth a second opinion.


mchamberland:
@rtownson This took longer than I thought, but here it is. I've spot-checked both Qt6 and Qt5 builds (Qt5.15 only). A second pair of eyes and a test on Linux would be good.

